PR #6, opened 2022-09-21
Belief Propagation Example

JoeyT1994:
Request to merge some changes to main branch.

Changes:
1) New folder added in examples which contains two files. 
        i)  BeliefPropagationFunctions.jl - general purpose functions for implementing belief propagation
        ii) ExampleBeliefPropagation.jl - Code for running a small example of using belief propagation to calculate local expectation 
             values for a random ItensorNetwork
2) Examples/peps/utils.jl: changed inner to dagger one of the ITensorNetworks and calculate the correct contraction


examples/peps/utils.jl
```diff
@@ -116,15 +116,15 @@ function ITensors.apply(o⃗::Vector{ITensor}, ψ::ITensorNetwork; cutoff, maxdi
 end
 
 function flattened_inner_network(ϕ::ITensorNetwork, ψ::ITensorNetwork)
-  tn = inner(ϕ, ψ)
```
mtfishman, line 119: I think `inner` should already take the proper `dag` (it should already `dag` the first tensor network `ϕ`).

examples/BeliefPropagation/BeliefPropagationFunctions.jl
```diff
@@ -0,0 +1,170 @@
...
+using LinearAlgebra
+
+maybe_only(x) = x
+maybe_only(x::Tuple{T}) where {T} = only(x)
+
+#Function to construct the initial message tensors associated with a given ItensorNetwork (two for each edge in the graph, one forward one backward)
+function construct_initial_mts(psi::ITensorNetwork, g::NamedDimGraph, s::IndsNetwork)
+    #Make empty lists
+    forwardmts = ITensor[]
+    backwardmts = ITensor[]
```
mtfishman, line 17: What about storing the message tensors in a `Dictionary` that maps the edges the message tensors are on to the message tensor? Then it is easy to look up the message tensors associated with a vertex by searching through the incident edges of the vertex.

examples/BeliefPropagation/BeliefPropagationFunctions.jl
```diff
@@ -0,0 +1,170 @@
...
+    noprime!(psidagv; tags = tags(siteind))
+    M = psiv * psidagv
+    #Apply all the incoming message tensors to it
+    for m in mts
+        M = m*M
+    end
+
+    #Normalise it
+    Mdat = matrix(M)
+    M = M / sqrt(tr(Mdat*Mdat'))
```
mtfishman, line 55: You should be able to use `normalize!(M)` here.

mtfishman:
Glad to see how simple it is! We could probably just move `examples/BeliefPropagation/BeliefPropagationFunctions.jl` into the source file, say as `src/belief_propagation/belief_propagation.jl`. That would make it easier to use in other places.

examples/BeliefPropagation/ExampleBeliefPropagation.jl
```diff
@@ -0,0 +1,47 @@
...
+Random.seed!(1234)
+
+n =14
+z = 3
+chi = 2
+nupdates = 100
+
+println("One Dimensional Chain")
+g = NamedDimGraph(Graphs.SimpleGraphs.grid([n]), [i for i  = 1:n])
+s = siteinds("S=1/2", g)
```
mtfishman, line 23: I guess we should allow passing a `Graphs.SimpleGraph` to `siteinds` if that isn't supported right now. Also we could make a default constructor `NamedDimGraph(g::SimpleGraph) = NamedDimGraph(g, 1:nv(g))`.

examples/peps/utils.jl
mtfishman: If the `dag` isn't be handled correctly by `inner` than it should be fixed there.

Issue #15, opened 2022-11-01
Implement boundary MPS contraction for grid networks


mtfishman:

JoeyT1994:
My latest pull request has a function which does this for an ITensorNetwork defined over an Lx * Ly rectangular grid. It could probably be made more efficient but seems to give correct results.

Issue #10, opened 2022-10-25
Add network combining features


mtfishman:
Add tools for combining and uncombining all or some of the link/site indices of a tensor network.

An interface I could imagine is the following for combining the link indices of a network:
```julia
tnc, combiners = combine_linkinds(tn)
```
`tnc` would be a tensor network with the same graph structure as `tn` but with the link indices merged. `combiners` would be a `DataGraph` with the combiners as metadata on the edges. Then the graph could be uncombined like:
```julia
tn = uncombineinds(tn, combiners)
```

This could be a wrapper around or more general function:
```julia
tnc, combiners = combineinds(tn; links=:, sites=[])
```
i.e. where you can choose a subset of link or site indices to combine.

@JoeyT1994

PR #17, opened 2022-11-04
Belief Propagation

JoeyT1994:
Added functions for belief propagation 
Added functions for creating special initial states (partition function of classical models)
Added extra capability for flattened networks (combining and uncombining their indices)
Added functions for forming subgraphs
Added a few extra utility functions
Added Boundary MPS contractor
Added belief propagation example


src/utils.jl
```diff
@@ -11,3 +11,26 @@ maybe_real(x::Complex) = iszero(imag(x)) ? real(x) : x
...
+  for i = 1:length(edges)
+      if((src(edges[i]) == source && dst(edges[i]) == dest) || (dst(edges[i]) == source && src(edges[i]) == dest))
+          return i
+      end
+  end
+  return 0
+end
+
+#Get subset of edges involving a specific vertex
+function find_edges_involving_vertex(edges::Vector{NamedDimEdge{Tuple}}, vertex::Tuple)
```
mtfishman, line 27: If the input was a graph instead of an edge list, you could just use: `Ref(vertex) .=> neighbors(g, vertex)` for this. Is there a reason you would have an edge list but not the actual graph?

mtfishman:
Thanks @JoeyT1994. It looks like there is something funny going on with the formatting. Could you format your code by running:
```julia
julia> using JuliaFormatter

julia> format(".")
```
in the directory `~/.julia/dev/ITensorNetworks`?

src/subgraphs.jl
```diff
@@ -0,0 +1,110 @@
+#Given a graph g, form 'nsubgraphs' subgraphs. Try to keep all subgraphs the same size and minimise edges cut between them
+#Return the subgraphs (1... nsubgraphs) with their contained vertices. Also return a dictionary of the subgraphs connected to each sgraph
+#KaHyPar needs to be installed for this function
+function formsubgraphs(g::NamedDimGraph, nsubgraphs::Int64)
+
+    ps = partition(g, nsubgraphs, configuration = :edge_cut, imbalance = 0.0)
+
+    subgraphconns = Dict{Int, Vector{Int}}()
```
mtfishman, line 8: Can `subgraphconns` just be a `Graph`?

mtfishman: In fact this seems to call for using a `DataGraph`, where the vertex metadata is the `Vector{Tuple}` storing the vertices associated with a particular partition.

src/specialitensornetworks.jl
```diff
@@ -0,0 +1,92 @@
...
+          ψ[v] = noprime!(ψ[v])
+      end
+  
+    end
+  
+    return ψ
+end
+
+#Build an ITensor network on a graph specified by the inds network s. Bond_dim is given by link_space and entries are randomised (normal distribution, mean 0 std 1)
+function randomITensorNetwork(s; link_space)
```
mtfishman, line 83: I think this is already defined somewhere else in the package?

src/specialitensornetworks.jl
```diff
@@ -0,0 +1,92 @@
+#RETURN THE IDENTITY TENSOR NETWORK (I.E. NETWORK WITH IDENTITY TENSORS ON EACH SITE) FOR A GIVEN GRAPH
+function identityITensorNetwork(g::NamedDimGraph; link_space = 2)
```
mtfishman, line 2: Maybe there is a better name for this, like `delta_network`? I wonder if this is alreadt called something in the literature. Maybe `copy_tensor_network`.

mtfishman: Also maybe to avoid code duplication between this version and the one below, this could just make an `IndsNetwork` and call the one below (which I assume is supposed to accept an `IndsNetwork`?).

src/specialitensornetworks.jl
```diff
@@ -0,0 +1,92 @@
...
+      is = inds(ψ[v])
+      ψ[v] = delta(is)
+  end
+  return ψ
+end
+  
+  #BUILD Z OF CLASSICAL ISING MODEL ON A GIVEN GRAPH AT INVERSE TEMP BETA
+  #H = -\sum_{(v,v') \in edges}\sigma^{z}_{v}\sigma^{z}_{v'}
+  #TAKE AS AN OPTIONAL ARGUMENT A LIST OF VERTICES OVER WHICH TO APPLY A SZ. THE RESULTANT NETWORK CAN THEN BE CONTRACTED AND DIVIDED BY THE ACTUAL PARTITION FUNCTION TO GET THAT OBSERVABLE
+  function ising_partition_function_ITensorNetwork(g::NamedDimGraph, beta::Float64; szverts=nothing)
```
mtfishman, line 26: Maybe we can just call this `ising_network`?

src/specialitensornetworks.jl
mtfishman: Oh I see, you moved it here. I think that makes sense.

src/subgraphs.jl
JoeyT1994: That seems like a good idea and would make using the subgraphs in routines a bit more straightforward. I will try to re-write it in that way.

src/specialitensornetworks.jl
JoeyT1994: Good idea, I have renamed to delta_network and avoided the code duplication.

src/specialitensornetworks.jl
JoeyT1994: Good idea, I have renamed to ising_network and avoided code duplication by just building the indsnetwork and invoking the inds_network version of the function.

src/utils.jl
JoeyT1994: Ah I see yes this is much neater way to do that. The latest commit has this changed.

src/subgraphs.jl
JoeyT1994: I have implemented this change in the latest commit. This should make the code more streamlined.

src/abstractitensornetwork.jl
```diff
@@ -242,244 +242,318 @@ end
...
+tn = contract(tn, (2, v...) => (1, v...))
+end
+if isnothing(sequence)
+sequence = contraction_sequence(tn; contraction_sequence_kwargs...)
+end
+return contract(tn; sequence)[]
 end
 
 # TODO: rename `sqnorm` to match https://github.com/JuliaStats/Distances.jl?
 norm2(ψ::AbstractITensorNetwork; sequence) = contract_inner(ψ, ψ; sequence)
```
JoeyT1994, line 452: #10  The two functions below should be a starting point for this.

src/abstractitensornetwork.jl
```diff
@@ -242,244 +242,318 @@ end
...
+end
+end
+
+flatpsi = flattened_inner_network(psiin, psi)
+flatpsi = rename_vertices(flatpsi, Dictionary(vertices(flatpsi), vertices(psi)))
+
+return combine_link_indices(flatpsi; combiners = combiners)
+end
+
 #
```
JoeyT1994, line 527: https://github.com/mtfishman/ITensorNetworks.jl/issues/10 The two functions above should be a starting point for this.

src/boundarymps.jl
```diff
@@ -0,0 +1,34 @@
...
+        tn[(i+1,j+1)] = U*S
+        tn[(i+1, j)] = V
+      end
+    end
+  
+    Z = ITensors.contract(tn)
+  
+    return Z
+  
+  end
```
JoeyT1994: #15

src/abstractitensornetwork.jl
```diff
@@ -242,244 +242,318 @@ end
...
+O = ITensor(Op(op, v), s)
+psiin[v] = noprime!(O*psiin[v])
+end
+end
+
+flatpsi = flattened_inner_network(psiin, psi)
+flatpsi = rename_vertices(flatpsi, Dictionary(vertices(flatpsi), vertices(psi)))
+
+return combine_link_indices(flatpsi; combiners = combiners)
+end
```
JoeyT1994, line 525: #10 These functions deal with combining (uncombining) any duplicated (thickened) link indices in an itensornetwork.

src/beliefpropagation.jl
```diff
@@ -0,0 +1,244 @@
+#Construct the random initial Message Tensors for an ITensor Network, based on a partitioning into subgraphs specified ny 'sub graphs'
+#The ITensorNetwork needs to be flat (i.e. just sites and link indices, no site indices), and is assumed to only have 1 link indice between any two sites
+#id_init = 0 => Random, id_init = 1 => Identity Matrix initialisation (preferred)
+function construct_initial_mts(
+  g::NamedDimGraph, flatpsi::ITensorNetwork, dg_subgraphs::DataGraph; id_init=1
```
mtfishman, line 5: An alternative syntax for choosing the tensor initialization that would allow more flexibility would be the following:
```julia
construct_initial_mts(g, tn, subgraphs; init=ones) # Constant
construct_initial_mts(g, tn, subgraphs; init=randn) # Normal distribution around 0
construct_initial_mts(g, tn, subgraphs; init=rand) # Uniform distribution from 0 to 1
construct_initial_mts(g, tn, subgraphs; init=(d -> 2rand(d) .- 1) # Uniform distribution from -1 to 1
```
Then, internally you can just construct the tensors with:
```julia
itensor(init(dim(edge_inds)), edge_inds)
```

(This makes me think that we should have ITensor/ITensorNetwork constructors like this that can take an arbitrary function to allow flexibility in initializing tensors various distributions but that's a separate issue.)

src/boundarymps.jl
```diff
@@ -0,0 +1,35 @@
+#GIVEN AN ITENSOR NETWORK CORRESPONDING TO A Lx*Ly grid indexed as (i,j) then perform boundary sweeping using a max_dim of chi_max
+function boundary_PEPS_contractor(
```
mtfishman, line 2: Here is a simpler implementation in terms of MPO-MPS contraction:
```julia
function contract_boundary_mps(tn::ITensorNetwork; kwargs...)
  dims = maximum(vertices(tn))
  d1, d2 = dims
  vL = MPS([tn[i1, 1] for i1 in 1:d1])
  for i2 in 2:(d2 - 2)
    T = MPO([tn[i1, i2] for i1 in 1:d1])
    vL = contract(T, vL; kwargs...)
  end
  T = MPO([tn[i1, d2 - 1] for i1 in 1:d1])
  vR = MPS([tn[i1, d2] for i1 in 1:d1])
  return inner(dag(vL), T, vR)[]
end
```
Note here I am deducing the size of the grid from the vertices, which is a bit fickle but I think is ok. This version allows more flexibility for choosing different options for the MPO-MPS contraction backend.

There is an issue that for larger networks, in general the norm will either diverge or decay exponentially with the number of tensors in the network, so we will have to be careful about that. A solution is to normalize `vL` as the contraction proceeds and store the norms. A related issue is that in general we probably want to return the boundary MPS `vL` since they will be used to compute expectation values.

src/utils.jl
```diff
@@ -11,3 +11,16 @@ maybe_real(x::Complex) = iszero(imag(x)) ? real(x) : x
 
 maybe_only(x) = x
 maybe_only(x::Tuple{T}) where {T} = only(x)
+
+#Find the edge in the list of edges which involves v1 and v2
+function find_edge(edges::Vector{NamedDimEdge{Tuple}}, v1::Tuple, v2::Tuple)
```
mtfishman, line 16: I guess this isn't needed anymore?

src/beliefpropagation.jl
```diff
@@ -0,0 +1,244 @@
...
+  for m in mts
+    push!(Contract_list, m)
+  end
+
+  for v in subgraph
+    psiv = flatpsi[v]
+    push!(Contract_list, psiv)
+  end
+
+  M = ITensors.contract(Contract_list)
```
mtfishman, line 52: I guess here we want to use some contraction sequence optimization.

src/beliefpropagation.jl
```diff
@@ -0,0 +1,244 @@
...
+
+  numerator = ITensors.contract(num_tensors_to_contract)[1]
+  denominator = ITensors.contract(denom_tensors_to_contract)[1]
+  out = numerator / denominator
+
+  return out
+end
+
+#Starting with initial guess for messagetensors, monitor the convergence of an observable on a single site v (which is emedded in psiflatO)
+function iterate_single_site_expec(
```
mtfishman, line 215: This could be a use case for [`Observers.jl`](https://github.com/GTorlai/Observers.jl), which allows users to pass functions that will perform measurements at each iteration.

src/beliefpropagation.jl
```diff
@@ -0,0 +1,244 @@
...
+  normalize!(M)
+
+  return M
+end
+
+#Do an update of all message tensors for a given flat ITensornetwork and its partition into sub graphs
+function update_all_mts(
+  g::NamedDimGraph,
+  flatpsi::ITensorNetwork,
+  mts::Dict{Pair,ITensor},
```
mtfishman, line 63: A slight code simplification could be to store the message tensors as the edge metadata of the `dg_subgraphs` object. I think it makes sense to put them together in one object, since then all of the metadata/caching required for the generalized BP algorithm is in one place.

However, an issue I see is that the standard `DataGraph` is an undirected graph, and by default it enforces that the metadata for an edge in both directions is the same. That was a subtle design choice I went back and forth about, which makes sense for `IndsNetwork` but for this case I see it is too restrictive. It could be that we will need to use a `DataDiGraph` (though that doesn't exist right now I think).

Also, is there a reason to pass around `g`? Doesn't it contain the same graph information as `flatpsi`?

src/beliefpropagation.jl
```diff
@@ -0,0 +1,244 @@
...
+        for e in NamedDimEdge.(Ref(vertex) .=> neighbors(g, vertex))
+          if (find_subgraph(dst(e), dg_subgraphs) == j)
+            edge_ind = commoninds(flatpsi, e)[1]
+            push!(edge_inds, edge_ind)
+          end
+        end
+      end
+      if (id_init == 1)
+        A = Array(delta(edge_inds), edge_inds)
+        X1 = ITensor(A, edge_inds)
```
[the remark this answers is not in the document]
JoeyT1994: Good spot, I have implemented that now.

src/boundarymps.jl
JoeyT1994: Okay nice. Yes I agree for now we should just assume the itensornetwork has been constructed via grid((lx,ly)) but could always add in a more rigorous check later on.

src/utils.jl
JoeyT1994: Yes I believe this is now redundant so I have taken it out.

src/boundarymps.jl
mtfishman: Yes, we could just add a function `is_grid` which checks the vertices and then a function `grid_size` to get the size of the grid.

src/beliefpropagation.jl
JoeyT1994: Is ITensors not automatically using some optimization (https://docs.juliahub.com/ITensors/P3pqL/0.2.3/ContractionSequenceOptimization.html#ITensors.NDTensors.contract)? Or do I have to enable this with ITensors.enable_contraction_sequence_optimization() ?

src/beliefpropagation.jl
JoeyT1994: Okay cool, yes that would be nice to weave in observers for some of the belief propagation stuff.

src/beliefpropagation.jl
JoeyT1994: Yeah that would be neat and makes a lot of sense to have dg as a date_di_graph. Would it be easy to define such a thing in DataGraphs.jl? 

Good point about g. All the relevant info should be in psi so I have removed the need for g in any of the belief propagation.jl

src/beliefpropagation.jl
JoeyT1994: Ah yes this is nice. I did change this but the only difficulty is I can't find a shorthand for the identity tensor? I.e. there seems to be no Julianic way of doing eye(dim(edge_inds)) and getting an identity tensor? Hence I had to add in a separate if statement so if someone puts init="I" it will build an identity tensor. Do you know of a way around this?

mtfishman: Hmm good question, I think I was confused about what type of tensor you were trying to make (I thought you were trying to make a tensor filled with all ones but I see you want to make a diagonal tensor with diagonal values of one).

A more flexible interface would be the following:
```julia
construct_initial_mts(g, tn, subgraphs; init=(I...) -> allequal(I) ? 1 : 0) # Diagonal tensor
construct_initial_mts(g, tn, subgraphs; init=(I...) -> 1) # Tensor filled uniformly with 1
construct_initial_mts(g, tn, subgraphs; init=(I...) -> randn()) # Normal distribution around 0
construct_initial_mts(g, tn, subgraphs; init=(I...) -> rand()) # Uniform distribution from 0 to 1
construct_initial_mts(g, tn, subgraphs; init=(I...) -> 2rand() - 1) # Uniform distribution from -1 to 1
```
so it would take a function that maps an index of the tensor to a value. Then the tensor could be constructed internally like:
```julia
itensor([init(Tuple(I)...) for I in CartesianIndices(tuple(dim.(edge_inds)...))], edge_inds)
```

src/beliefpropagation.jl
mtfishman: I would design it where the user passes a keyword argument with a function that computes the contraction sequence, and then the user can decide on the particular algorithm.

For example:
```julia
function updatemt(
  g::NamedDimGraph, flatpsi::ITensorNetwork, subgraph::Vector{Tuple}, mts::Vector{ITensor}; contraction_sequence::Function=tn -> ITensorNetworks.contraction_sequence(tn; alg="optimal")
)
  # [...]
  M = contract(Contract_list; sequence=contraction_sequence(Contract_list))
end
```

src/beliefpropagation.jl
mtfishman: It wouldn't be hard to define `DataDiGraph` but let's leave that for a future PR.

src/beliefpropagation.jl
JoeyT1994: Okay great, I have designed it that way now.

src/beliefpropagation.jl
JoeyT1994: Good idea, I have put this in

mtfishman:
@JoeyT1994 I'll merge this and then fix issues caused by #22.

After I do that, it would be good to add some tests for these functions in a follow up PR.

PR #18, opened 2022-11-04
Additions needed for sweeping algorithms on trees

leburgel:
Work in progress, this adds some functionalities for tree tensor networks which are needed to be able to run sweeping algorithms on them. It's quite a lot, so in addition to raw code review I'll try to summarize the main additions below, as well as start a list of things I thought might be good to discuss while coding (some of which are not very relevant and should just be filed as issues...).

Summary of additions:
- `ITensorNetwork`:
    - Typeonversion
    - `ITensors.replaceinds`:
        - modified to check for equality of underlying graphs, and to take into account unassigned vertices or edges
    - Priming and tagging:
        - added `linkinds` and `siteinds` filters similar to the `ITensors.MPS` case
    - Truncation
    - `linkdims`
    - Common site index checking
    - Site index combiners
- `AbstractTreeTensorNetwork`
    - Some additional constructors
        - Constructors starting from dense tensors
        - Random and product state constructors
    - Orthogonalization, contraction and truncation
    - Norm, lognorm and normalization
    - Some scalar arithmetic
    - Addition (direct sum algorithm)
- `TreeTensorNetworkOperator` (new type)
    - Automatic generation from `ITensors.OpSum`
    - Expectation value and variance (`inner`)
- `ProjTTNO`, `ProjTTNOApply` and `ProjTTNOSum` (new types)


src/models.jl
```diff
@@ -1,10 +1,78 @@
...
+    for e in edges(g)
+      ℋ += J1 / 2, "S+", maybe_only(src(e)), "S-", maybe_only(dst(e))
+      ℋ += J1 / 2, "S-", maybe_only(src(e)), "S+", maybe_only(dst(e))
+      ℋ += J1, "Sz", maybe_only(src(e)), "Sz", maybe_only(dst(e))
+    end
   end
-  for v in vertices(g)
-    ℋ += h, "X", maybe_only(v)
+  if !iszero(J2)
+    # TODO, more clever way of looping over next to nearest neighbors?
```
mtfishman, line 20: This would be nice functionality to have, and is worth splitting off into a separate function (`next_nearest_neighbors`?).

It looks like we could make use of [Graphs.neighborhood_dists](https://juliagraphs.org/Graphs.jl/dev/algorithms/connectivity/#Graphs.neighborhood_dists-Union{Tuple{T},%20Tuple{U},%20Tuple{AbstractGraph{T},%20Integer,%20Any},%20Tuple{AbstractGraph{T},%20Integer,%20Any,%20AbstractMatrix{U}}}%20where%20{U%3C:Real,%20T%3C:Integer}) to create a more general function for outputting the vertices that are the k-nearest neighbors of a specified vertex, for example:
```julia
function k_nearest_neighbors(g::AbstractGraph, v, d)
  nds = neighborhood_dists(g, v, d)
  # Filter `nds` for vertices with distances equal to `d`
end
```
Then `next_nearest_neighbors(g, v) = k_nearest_neighbors(g, v, 2)`.

src/itensornetwork.jl
```diff
@@ -86,15 +102,33 @@ function insert_links(ψ::ITensorNetwork, edges::Vector=edges(ψ); cutoff=1e-15)
...
 end
 
-function ITensorNetwork(is::IndsNetwork, initstate::Function)
+function ITensorNetwork(
+  ::Type{ElT}, is::IndsNetwork, states_map::Dictionary
+) where {ElT<:Number}
   ψ = ITensorNetwork(is)
   for v in vertices(ψ)
-    ψ[v] = state(initstate(v), only(is[v]))
+    ψ[v] = ITensors.convert_leaf_eltype(ElT, state(only(is[v]), states_map[v]))
```
mtfishman, line 110: I guess `ITensors.state` should optionally take an element type, could you raise an issue about that in ITensors.jl?

In the meantime, you should be able to just use `convert_eltype` here since it is just an ITensor (`ITensors.convert_leaf_eltype` is more for nested data structures like `MPS`, `Vector{ITensor}`, `ITensorNetwork`, etc.).

src/abstractitensornetwork.jl
```diff
@@ -186,7 +243,7 @@ const map_inds_label_functions = [
   :setprime,
   :noprime,
   :replaceprime,
-  :swapprime,
+  :swapprime, # TODO: fix this one (broken)
```
emstoudenmire, line 246: Would it be possible to fix this case if it's broken? If it's something subtle, could you add a `@test_broken` style unit test (https://docs.julialang.org/en/v1/stdlib/Test/#Broken-Tests) to flag this as needing fixing?

src/abstractitensornetwork.jl
```diff
@@ -483,3 +606,83 @@ function visualize(
...
+  return linkdim(tn, edgetype(tn)(edge))
+end
+
+function linkdim(tn::AbstractITensorNetwork, edge::AbstractEdge)
+  ls = linkinds(tn, edge)
+  return prod([isnothing(l) ? 1 : dim(l) for l in ls])
+end
+
+function linkdims(tn::AbstractITensorNetwork)
+  return Dictionary(edges(tn), map(e -> linkdim(tn, e), edges(tn)))
```
mtfishman, line 632: I'm going back and forth about whether or not this should output a `DataGraph` with the dimensions as metadata on the edges.

I think an advantage is that in principle it could more easily be used to then construct a new `ITensorNetwork` (kind of like how `IndsNetwork` is currently used, but more like an `IndexSpaceNetwork` or `SpaceNetwork`). Also, it should then automatically handle making sure that inputs of edges in either direction still output the correct dimension (i.e. `linkdims(tn)[(1, 1) => (2, 1)]` and `linkdims(tn)[(2, 1) => (1, 1)]` should both work and output the same dimension).

src/ITensorNetworks.jl
```diff
@@ -4,17 +4,20 @@ using DataGraphs
...
 using DocStringExtensions
 using Graphs
+using IsApprox
 using ITensors
 using ITensors.ContractionSequenceOptimization
 using ITensors.ITensorVisualizationCore
 using MultiDimDictionaries
 using NamedGraphs
 using Requires
+using SparseArrayKit
```
mtfishman, line 14: Is this used anywhere?

emstoudenmire:
@mtfishman @leburgel :  More of a minor comment, but should the `TreeTensorNetworkOperator` and `TreeTensorNetworkState` types just be named `TTNO` and `TTNS` ? I see that these are provided as aliases for the above, but then I think most people, including developers, would prefer to always use the abbreviations. Let me know if there's a more technical reason for needing the longer names.

mtfishman:
> @mtfishman @leburgel : More of a minor comment, but should the `TreeTensorNetworkOperator` and `TreeTensorNetworkState` types just be named `TTNO` and `TTNS` ? I see that these are provided as aliases for the above, but then I think most people, including developers, would prefer to always use the abbreviations. Let me know if there's a more technical reason for needing the longer names.

That choice is mostly due to my bias towards being more explicit and avoiding abbreviations as much as possible. I don't see the disadvantage of providing the aliases which people can use if they want (and we can use internally if we decide that style is preferred).

src/models.jl
leburgel: Sounds good. I assume it would make sense to add this in NamedGraphs.jl? I'll try to find some time to add a PR soon.

src/itensornetwork.jl
leburgel: Issue raised, and I switched to `convert_eltype`.

src/abstractitensornetwork.jl
leburgel: I haven't really thought about what is going wrong for this specific case, I just noticed it in passing and added a reminder. It might not be a straightforward fix with the current implementation, but I would have to actually look at it. There's no tests for the priming and tagging part yet anyway, I can sort this out when I get to that.

src/abstractitensornetwork.jl
leburgel: I hadn't thought of indexing with edges in either direction (or with pairs of vertices for that matter), I'll change it to return a `DataGraph`. To be specific, I guess it would be a `NamedDimDataGraph` to be consistent with with the `data_graph` of an `ITensorNetwork`?

src/ITensorNetworks.jl
leburgel: Only in an intermediate internal step, namely the conversion of an `OpSum` to a `DataGraph` of `SparseArray`s with operators as entries used to represent a sparse symbolic tree tensor network Hamiltonian. This format was just the easiest for visualization and debugging for me personally, but I'm definitely open to suggestions for a different format (if we even want to keep this intermediate representation around at all in the end).

src/abstractitensornetwork.jl
leburgel: Done, but just printing `linkdims` of a network is a bit weird now since it shows both edge orientations separately for each edge when printing a `DataGraph`. But that's a potential issue for DataGraphs.jl

leburgel:
> > @mtfishman @leburgel : More of a minor comment, but should the `TreeTensorNetworkOperator` and `TreeTensorNetworkState` types just be named `TTNO` and `TTNS` ? I see that these are provided as aliases for the above, but then I think most people, including developers, would prefer to always use the abbreviations. Let me know if there's a more technical reason for needing the longer names.
> 
> That choice is mostly due to my bias towards being more explicit and avoiding abbreviations as much as possible. I don't see the disadvantage of providing the aliases which people can use if they want (and we can use internally if we decide that style is preferred).

I stuck to explicit type names and short aliases since the choice had already been made for `TreeTensorNetworkState`, but I haven't really paid attention to consistently use either after that. I think I mostly used the shorthands later on in the code, but this was not really purposeful and can be changed if desirable.

src/treetensornetwork/opsum_to_ttno.jl
```diff
@@ -0,0 +1,594 @@
+# convert ITensors.OpSum to TreeTensorNetworkOperator
+
+# TODO: fix symbolic SVD compression for certain combinations of long-range interactions!
+
+# 
+# Utility methods
+# 
+
+# number of neighbors of vertex in graph
+function num_neighbors(graph::AbstractGraph, vertex...)
```
mtfishman, line 10: Looks like this is the same as `degree`: https://juliagraphs.org/Graphs.jl/dev/core_functions/core/#Graphs.degree. Hopefully `Graphs.degree` already works for `AbstractNamedGraph` and `AbstractDataGraph`.

src/treetensornetwork/ttno.jl
```diff
@@ -0,0 +1,148 @@
...
+
+A finite size tree tensor network operator type. 
+Keeps track of the orthogonality center.
+
+# Fields
+
+- itensor_network::ITensorNetwork
+- ortho_lims::Vector{Tuple}: A vector of vertices defining the orthogonality limits.
+"""
+mutable struct TreeTensorNetworkOperator <: AbstractTreeTensorNetwork
```
mtfishman, line 12: I wonder if we can get away with just having a `TreeTensorNetwork` type that is treated as an operator or a state depending on the context.

If we really need to distinguish them for the sake of dispatch, we could have a wrapper type like `Map(ttn)` or `Operator(ttn)`, which could be like an `ITensorNetworkMap`: https://github.com/ITensor/ITensors.jl/blob/v0.3.20/src/ITensorNetworkMaps/ITensorNetworkMaps.jl

src/treetensornetwork/abstracttreetensornetwork.jl
```diff
@@ -0,0 +1,399 @@
...
+function logdot(ψ1::TTNT, ψ2::TTNT; kwargs...) where {TTNT<:AbstractTreeTensorNetwork}
+  return _log_or_not_dot(ψ1, ψ2, true; kwargs...)
+end
+
+function loginner(ψ1::TTNT, ψ2::TTNT; kwargs...) where {TTNT<:AbstractTreeTensorNetwork}
+  return logdot(ψ1, ψ2; kwargs...)
+end
+
+# TODO: stick with this traversal or find optimal contraction sequence?
+function _log_or_not_dot(
```
mtfishman, line 255: For code simplicity, it may be better to just implement `logdot`/`loginner` directly, and then just use `inner(x, y) = exp(loginner(x, y))`.

I think I was worried that `exp(loginner(x, y))` might lose some precision but I don't think that's the case in practice.

src/treetensornetwork/projttno.jl
```diff
@@ -0,0 +1,52 @@
+"""
+ProjTTNO
+"""
+mutable struct ProjTTNO <: AbstractProjTTNO
```
mtfishman, line 4: I think it is better to not make this `mutable`, in general it is easier to reason about non-mutable structs.

If we need to update `pos` in-place, we can wrap it in a `Ref`.

src/treetensornetwork/projttno.jl
```diff
@@ -0,0 +1,52 @@
...
+  pos::Union{Vector{<:Tuple},NamedDimEdge{Tuple}} # TODO: cleanest way to specify effective Hamiltonian position?
+  H::TTNO
+  environments::Dictionary{NamedDimEdge{Tuple},ITensor}
```
mtfishman, line 7: It may be nice to make this a directed `DataGraph`.

src/treetensornetwork/projttnosum.jl
```diff
@@ -0,0 +1,61 @@
+"""
+ProjTTNOSum
+"""
+mutable struct ProjTTNOSum
```
mtfishman, line 4: Again, I think this should not be `mutable`.

I was also thinking about using the `Applied` system in `ITensors.LazyApply` to avoid having to actually define types like this. That would be a more general way to represent lazy combinations of projected TTNs (i.e. help with representing more general combinations, like sums of projected TTNS and TTNO, sums of TTNOs applied to TTNS, etc. without having to hardcode specific types).

Maybe we could chat about how that might work, this would be a good opportunity to rethink some of the type designs here (some of these designs are holdovers from porting the C++ version).

src/models.jl
mtfishman: Yeah, it would make sense to add to `NamedGraphs.jl`.

src/abstractitensornetwork.jl
mtfishman: Yeah, ideally we would comment it out and then add `@test_broken` as a reminder for us to investigate it.

src/abstractitensornetwork.jl
```diff
@@ -150,6 +189,14 @@ function siteinds(tn::AbstractITensorNetwork, vertex...)
   return uniqueinds(tn, vertex...)
 end
 
+function siteinds(::typeof(all), tn::AbstractITensorNetwork, vertex...)
+  return siteinds(tn, vertex...)
+end
+
+function siteinds(::typeof(only), tn::AbstractITensorNetwork, vertex...)
```
mtfishman, line 196: I eventually regretted adding functions like this. I think it is easy enough to write `only(siteinds(tn, v))`.

src/ITensorNetworks.jl
mtfishman: Ok, just wanted to double check, that sounds reasonable for now.

src/abstractitensornetwork.jl
mtfishman: This should all be easier now with the latest updates to `DataGraphs.jl`, you should be able to just use a `DataGraph`.

src/abstractitensornetwork.jl
mtfishman: Also, I've decide in the latest versions of `NamedGraphs.jl` and `DataGraphs.jl` to not allow splatted vertex indexing in general (except in the case of `data_graph[1, 1]` for getting vertex data), so this and related functions should just be:
```julia
function siteinds(tn::AbstractITensorNetwork, vertex)
  [...]
end
```

src/abstractitensornetwork.jl
leburgel: Done.

src/abstractitensornetwork.jl
leburgel: Done.

src/treetensornetwork/opsum_to_ttno.jl
leburgel: Good catch, completely missed this. `Graphs.degree` works fine now thanks to mtfishman/NamedGraphs.jl#17 and mtfishman/DataGraphs.jl#7.

src/treetensornetwork/abstracttreetensornetwork.jl
leburgel: I removed `_log_or_not_dot` in favor of just defining `logdot` but I haven't replaced `inner(...)` with `exp(loginner(...))` everywhere yet. I'll do this later (I still have to define a three-argument logdot for expectation values and sort out some minor things).

src/abstractitensornetwork.jl
leburgel: I removed the extra filters for `siteinds` and `linkinds`, as well as all splatted vertex indexing. I don't even remember why I added the filters in the first place, I must have been overloading some other `ITensors.jl` function that used this interface and added them without thinking.

src/models.jl
mtfishman: Note that [Graphs.neighborhood_dists](https://juliagraphs.org/Graphs.jl/dev/algorithms/connectivity/#Graphs.neighborhood_dists-Union%7BTuple%7BT%7D,%20Tuple%7BU%7D,%20Tuple%7BAbstractGraph%7BT%7D,%20Integer,%20Any%7D,%20Tuple%7BAbstractGraph%7BT%7D,%20Integer,%20Any,%20AbstractMatrix%7BU%7D%7D%7D%20where%20%7BU%3C:Real,%20T%3C:Integer%7D) should now work as of [NamedGraphs](https://github.com/mtfishman/NamedGraphs.jl/pull/17).

src/abstractitensornetwork.jl
mtfishman: No problem. There is going to be some tension between the MPO/MPS types in ITensors.jl and AbstractITensorNetwork types in ITensorNetworks.jl in terms of the interface, since I think ITensorNetworks.jl is an opportunity to rethink some of the MPO/MPS interface using a more general network perspective.

src/treetensornetwork/abstracttreetensornetwork.jl
```diff
@@ -0,0 +1,366 @@
...
+    @warn "log(norm²) is $lognorm2_T, which is not real up to a relative tolerance of $rtol and an absolute tolerance of $atol. Taking the real part, which may not be accurate."
+  end
+  return 0.5 * real(lognorm2_ψ)
+end
+
+function loginner(ψ1::TTNT, ψ2::TTNT; kwargs...) where {TTNT<:AbstractTreeTensorNetwork}
+  return logdot(ψ1, ψ2; kwargs...)
+end
+
+# TODO: stick with this traversal or find optimal contraction sequence?
```
mtfishman, line 236: I think ultimately this should just be designed based on calling a more general `contract` function with a contraction sequence appropriate for a tree network, and we can use a tree traversal as a way to quickly derive a reasonable sequence, and users can pass their own contraction sequence or function for computing a contraction sequence. But that can be left for a future PR.

mtfishman: Tracking this issue in #25.

src/treetensornetwork/ttno.jl
mtfishman: Tracking in #26.

src/treetensornetwork/abstracttreetensornetwork.jl
mtfishman: Sounds good. Also I think it would be better to have the implementation function called `loginner` instead of `logdot` since that is the version people use the most, and it will make it easier to find the implementation.

src/models.jl
leburgel: Tracking this issue in mtfishman/NamedGraphs.jl#22.

src/treetensornetwork/abstracttreetensornetwork.jl
leburgel: Done.

src/treetensornetwork/projttno.jl
leburgel: Could be a good idea, adding/invalidating environments by adding/removing directed edges sounds nice. I would have to think about what to do with the inside the operator position for a 2-site effective operator (remove them altogether?) and whether I could get rid of some other things by using more `DataGraph`-like methods instead.

mtfishman:
@leburgel you bring up a lot of good questions and points in your first post but it is tough to try to discuss and resolve them from this PR. Could you split up questions that haven't been resolved yet into separate issues?

leburgel:
> @leburgel you bring up a lot of good questions and points in your first post but it is tough to try to discuss and resolve them from this PR. Could you split up questions that haven't been resolved yet into separate issues?

Yes, I'll drop the things that are no longer relevant and move the rest to new issues.

src/treetensornetwork/projttno.jl
leburgel: I updated things so that all tree tensor network types are immutable. I can get away with not updating the orthogonality center (for `TTNS` and `TTNO`) or position (for `ProjTTNO`, `ProjTTNOSum` as well as relatives in `ITensorTDVP.jl`) in-place, as long as things such as truncation, orthogonalization, local updates in TDVP, ..., are also out of place. We can maybe discuss whether this is fine or whether we do want to be able to update in-place via a `Ref`.

src/treetensornetwork/projttnosum.jl
leburgel: This definitely sounds like a nice idea.

src/treetensornetwork/projttno.jl
mtfishman: In general, I have found it is best to avoid in-place operations unless there is a performance reason for using it, since it makes the code easier to reason about (and also makes it easier for automatic differentiation). So I would say try to avoid in-place operations in general, and then we can profile things and see if anything is slow because of that. Because most of our code is going to be dominated by tensor contractions and decompositions, we can generally get away with some extraneous shallow copying of high level data structures (like graphs, dictionaries, etc.).

src/treetensornetwork/abstractprojttno.jl
```diff
@@ -0,0 +1,145 @@
...
+function incident_edges(P::AbstractProjTTNO{V})::Vector{NamedEdge{V}} where {V}
+  on_edge(P) && return [pos(P), reverse(pos(P))]
+  edges = [
+    [edgetype(P)(n => v) for n in setdiff(neighbors(underlying_graph(P), v), sites(P))] for
+    v in sites(P)
+  ]
+  return collect(Base.Iterators.flatten(edges))
+end
+
+function internal_edges(P::AbstractProjTTNO{V})::Vector{NamedEdge{V}} where {V}
```
mtfishman, line 36: This is a good idea for a generic graph function. It is the complement of the edge boundary of a subgraph (https://en.wikipedia.org/wiki/Boundary_(graph_theory) and `boundary_edges` added in https://github.com/mtfishman/NamedGraphs.jl/pull/20).

As we discussed, a simple implementation would be `internal_edges(g::AbstractGraph, vertices) = edges(subgraph(g, vertices))`. Then this function could be defined as `internal_edges(P::AbstractProjTTNO) = internal_edges(underlying_graph(P), sites(P))`.

PR #43, opened 2023-01-09
Move `ITensorTDVP.jl` MPS solver code into `ITensorNetworks.jl`

mtfishman:
This moves the code in the package [ITensorTDVP.jl](https://github.com/ITensor/ITensorTDVP.jl) into this package, including tests and examples.

This is in preparation for @leburgel's PR https://github.com/ITensor/ITensorTDVP.jl/pull/43, which extends those MPS solver functions to general tree tensor networks which are defined in this package. We agreed it makes the most sense to just move that PR here to keep `ITensorTDVP.jl` dependent on only `ITensors.MPS`/`ITensors.MPO`.

It also makes `KrylovKit.eigsolve` an alias for `ITensorNetworks.dmrg` (previously `ITensorTDVP.dmrg`).

@emstoudenmire we had discussed merging https://github.com/ITensor/ITensorTDVP.jl/pull/43 into `ITensorTDVP.jl` but I spoke to Lander and we both agreed this makes more sense this way. I think also we can be more aggressive about deprecating old types in this package, like `ITensors.AbstractObserver`, `ITensors.Sweeps`, `ITensors.AbstractMPS` and subtypes, etc.


emstoudenmire:
Sounds good – if you and Lander find this to be the easiest then I'm all for it. I just didn't want Lander to have to re-create his PR here if that ended up being more work; but if it's less work then it is the way to go.

emstoudenmire:
Also fine to deprecate those older types or leave them out completely. We can always re-wrap them into the existing system later if we want to. Or else provide an "upgrade path" from using functions taking ITensors.MPS to ITensorNetworks.MPS.

PR #44, opened 2023-01-10
Sweeping algorithms for tree tensor networks

leburgel:
Moved here from ITensor/ITensorTDVP.jl#43

A first working attempt at generalizing MPS sweeping algorithms to tensor networks on arbitrary tree geometries. All algorithms should work for both `ITensors.MPS` and `ITensorNetworks.TreeTensorNetworksState` (except some of the newer additions which I haven't generalized yet), with the same syntax currently used in [ITensorTDVP.jl](https://github.com/ITensor/ITensorTDVP.jl).

Aside from some patching to work with `MPS` and trees using a common interface, the main internal changes are how the sequence of update steps within an individual sweep is generated and how individual local updates are performed. The idea is that a sequence of `SweepStep`s is generated for the given geometry depending on the kind of sweeping being used (e.g. 1- or 2-site, reverse steps or not), where each `SweepStep` contains all the information needed for a single local update. This allows to split off the geometry-dependent part and adds flexibility in the sweeping sequences used. In addition, by treating each local update step in the same way we can use a single update routine for all local updates (whether it's a bond- or site-update, forwards or backwards evolution...) irrespective of what happens before or after in the sequence.


mtfishman:
Thanks @leburgel, this is great. It is a big step forward for ITensor, and will become our standard DMRG/TDVP code!

Issue #46, opened 2023-01-12
Support MPO-(MPS with dangling site indices) contraction


shinaoka:
I made a fresh start after #44 was merged into the main branch.

For QTT, we may want to contract an MPO (with two site indices at each vertex) and an MPS with dangling site indices.
I want to sout out issues step by step here.

A minimum example for `naive` algoritm works:

```julia
nbit = 3
sites = siteinds("Qubit", nbit)

H = randomMPO(sites) + randomMPO(sites)
H = replaceprime(H, 1=>2, 0=>1)

psi = randomMPO(sites) + randomMPO(sites)

# MPO-MPO
Hpsi_ref = contract(H, psi; alg="naive")

# MPO-MPS
Hpsi_ref2 = contract(H, MPS([A for A in psi]); alg="naive")

@show H
@show psi
@show Hpsi_ref
@show Hpsi_ref2
@assert MPS([A for A in Hpsi_ref]) ≈ Hpsi_ref2
```

```
H = MPO
[1] ((dim=2|id=58|"Qubit,Site,n=1")'', (dim=2|id=58|"Qubit,Site,n=1")', (dim=2|id=68|"Link,l=1")')
[2] ((dim=2|id=656|"Qubit,Site,n=2")'', (dim=2|id=656|"Qubit,Site,n=2")', (dim=2|id=121|"Link,l=2")', (dim=2|id=68|"Link,l=1")')
[3] ((dim=2|id=270|"Qubit,Site,n=3")'', (dim=2|id=270|"Qubit,Site,n=3")', (dim=2|id=121|"Link,l=2")')

psi = MPO
[1] ((dim=2|id=58|"Qubit,Site,n=1")', (dim=2|id=58|"Qubit,Site,n=1"), (dim=2|id=787|"Link,l=1"))
[2] ((dim=2|id=656|"Qubit,Site,n=2")', (dim=2|id=656|"Qubit,Site,n=2"), (dim=2|id=46|"Link,l=2"), (dim=2|id=787|"Link,l=1"))
[3] ((dim=2|id=270|"Qubit,Site,n=3")', (dim=2|id=270|"Qubit,Site,n=3"), (dim=2|id=46|"Link,l=2"))

Hpsi_ref = MPO
[1] ((dim=2|id=58|"Qubit,Site,n=1")'', (dim=2|id=58|"Qubit,Site,n=1"), (dim=4|id=716|"CMB,Link"))
[...]

PR #47, opened 2023-01-12
Remove `TTNS` and `TTNO` types in favor of `TTN`

mtfishman:
This remove `TTNS` and `TTNO` types in favor of `TTN`.

It seemed to essentially require just finding and replacing `TTNS`/`TTNO` for `TTN`, giving a first indication that having separate types wasn't really necessary.

Some constructors may become ambiguous, but I think those can be fixed on a case-by-case basis (for example making a product state vs. product operator, we could pass two sets of indices to indicate it is an operator, or wrap the inputs in `Op` or `State`).


mtfishman:
Besides code simplicity, I think a big advantage of this change is psychological and philosophical: ITensor (and tensor networks) were designed to move beyond thinking too strictly in terms of linear algebra, so removing the distinction between MPO/MPS or TTNO/TTNS will force us to write more general (and I think in the end simpler) tensor network code. Often I've found that thinking too much about something being a "state" or "operator" when writing MPO/MPS code backs me into a corner and therefore I don't anticipate generalizations, such as MPS-like objects that have dangling indices which are then not quite states any more (see for example #46).

I think there will be use cases for having operator-like tensor networks, for example in DMRG. In that case, I think the correct concept is thinking about tensor network maps, i.e. tensor networks that take a tensor network as an input and then maps the result back to the same space.

For that, I think we could have a generic `ITensorNetworkMap` object which wraps an `AbstractITensorNetwork`, which defines how the indices get mapped to the original ones after contraction (which by default could be `mapprime(tn, 1 => 0)`, for example). I've already experimented with that in ITensors.jl: https://github.com/ITensor/ITensors.jl/blob/v0.3.23/src/ITensorNetworkMaps/ITensorNetworkMaps.jl. I'm hoping that can mostly be avoided, at least in user-facing code, but when it is needed it should be composed on top of the `TreeTensorNetwork` type design in a more general way.

PR #52, opened 2023-01-23
Fix `contract` tests

leburgel:
Fixes tests for `contract(::AbstractTTN, ::AbstractTTN)` by fixing a bug in `inner(::AbstractTTN, ::AbstractTTN)` and changing the initial states used in the tests. Aside from the bug making it it seem like the algorithm failed when it actually didn't, the updated behavior for the initial state always starts from a product state by default. So the `nsite=2` case needs more sweeps to be able to grow the bond dimension enough, and the `nsite=1` case couldn't work without explicitly supplying an initial state with a nontrivial bond dimension.

### DIFF
--- a/src/treetensornetworks/abstracttreetensornetwork.jl
+++ b/src/treetensornetworks/abstracttreetensornetwork.jl
@@ -154,7 +154,7 @@ function inner(
   ψ::AbstractTTN;
   root_vertex=default_root_vertex(ϕ, ψ),
 )
-  ϕᴴ = sim(dag(ψ); sites=[])
+  ϕᴴ = sim(dag(ϕ); sites=[])
   ψ = sim(ψ; sites=[])
   ϕψ = ϕᴴ ⊗ ψ
   # TODO: find the largest tensor and use it as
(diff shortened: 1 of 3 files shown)


src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -154,7 +154,7 @@ function inner(
   ψ::AbstractTTN;
   root_vertex=default_root_vertex(ϕ, ψ),
 )
-  ϕᴴ = sim(dag(ψ); sites=[])
+  ϕᴴ = sim(dag(ϕ); sites=[])
```
mtfishman, line 157: Yikes that was a bad bug!

mtfishman:
Great, thanks @leburgel! Glad it was a simple fix and not even an issue with `contract`.

That is an issue with the current default for `init`, fundamentally it is a difficult issue for any of these algorithms (DMRG, TDVP, etc.) to choose general initial guesses and that's way we make people input initial guesses in DMRG. So we'll have to think carefully about that going forward, and maybe just force people to input an initial guess but give tools like forming a random TTN of a certain bond dimension to help with it and also provide tools for extracting the external indices in a form that works with this function, based on the input TTNs being contracted.

PR #53, opened 2023-01-24
Contraction tree to graph

JoeyT1994:
This PR adds functionality for converting a contraction sequence (outputted by `contraction_sequence(tn)`) to a graph.
Testing is also included.

It is possible the output (which is just a `NamedGraph`) should be defined as it's own `type` so that methods like `leaf_node` can be assigned explicitly to it.


src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
+
+"""The main object here is `g' a NamedGraph which represents a graphical version of a contraction sequence.
+It's vertices describe a partition between the leaves of the sequence (will be labelled with an n = 1 or n = 3 element tuple, where each element of the tuple describes the leaves in one of those partition)
+n = 1 implies the vertex is actually a leaf.
+Edges connect vertices which are child/ parent and also define a bi-partition"""
+
+
+"""Function to take a sequence (returned by ITensorNetworks.contraction_sequence) and construct a graph g which represents it (see above)"""
+function create_contraction_tree(contract_sequence)
```
mtfishman, line 9: I would call this `contraction_sequence_to_graph` since we use the terminology `contraction_sequence` elsewhere.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
+
+"""The main object here is `g' a NamedGraph which represents a graphical version of a contraction sequence.
```
mtfishman, line 2: I would put this into the doctstring of `create_contraction_tree`/`contraction_sequence_to_graph`.

src/ITensorNetworks.jl
```diff
@@ -82,6 +82,7 @@ include("specialitensornetworks.jl")
 include("renameitensornetwork.jl")
 include("boundarymps.jl")
 include("beliefpropagation.jl")
+include("ContractionTreeToGraph.jl")
```
mtfishman, line 85: I think we should name this file `contraction_tree_to_graph.jl`. File names are generally lower case unless the file defines a module.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+
+
+"""Function to take a sequence (returned by ITensorNetworks.contraction_sequence) and construct a graph g which represents it (see above)"""
+function create_contraction_tree(contract_sequence)
+    #Collect all the leaves
+    leaves = collect(Leaves(contract_sequence))
+    #Build empty graph
+    g = NamedGraph()
+    #Construct the left branch
+    spawn_child_branch(contract_sequence[1], g, leaves)
```
mtfishman, line 15: I would call this:
```julia
fill_contraction_sequence_graph_vertices!(g, contract_sequence)
```
and then have it organized like:
```julia
function fill_contraction_sequence_graph_vertices!(g, contract_sequence)
  leaves = collect(Leaves(contract_sequence))
  fill_contraction_sequence_graph_vertices!(g, contract_sequence, leaves)
  return g
end

function fill_contraction_sequence_graph_vertices!(g, contract_sequence, leaves)
  fill_contraction_sequence_graph_vertices!(g, contract_sequence[1], leaves)
  fill_contraction_sequence_graph_vertices!(g, contract_sequence[2], leaves)
  return g
end
```
or something like that. But the point would be to hide more details away in layers of functions and out of the top level `create_contraction_tree`/`contraction_sequence_to_graph` function.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+"""Utility functions for the graphical representation of a contraction sequence
+Perhaps it should be a specific object or type of tree?!"""
+
+"""Determine if a node is a leaf"""
+function leaf_node(g::AbstractGraph, v)
+    return length(neighbors(g, v)) == 1
+end
+
+"""Determine if an edge involves a leaf (at src or dst)"""
+function leaf_edge(g::AbstractGraph, e)
```
mtfishman, line 65: A better name would be `is_leaf_edge`.

I've been putting functions like this that are generic to all graph types in the `src/Graphs/abstractgraph.jl` file of `NamedGraphs.jl`. Maybe for now we could move these to a `src/Graphs/abstractgraph.jl` of `ITensorNetworks.jl` and then move them to `NamedGraphs.jl` later.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+    return leaf_node(g, src(e)) || leaf_node(g, dst(e))
+end 
+
+"""Determine if a node has no neighbours which are leaves"""
+function no_leaf_neighbours(g::AbstractGraph, v)
```
mtfishman, line 70: I would reverse the logic of this and name it `has_leaf_neighbor`.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+        if(leaf_node(g, vn))
+            return false
+        end
+    end
+    return true
+end
+
+"""Get all edges which do not involve a leaf
+(why does edges(g)[contraction_edge(g, edges(g)) .== true] not work?)"""
+function non_leaf_edges(g::AbstractGraph)
```
mtfishman, line 81: Maybe `internal_edges`? That would follow the terminology of "internal vertices" and "external vertices": https://en.wikipedia.org/wiki/Tree_(graph_theory)#Tree

Also what about: `filter(e -> !is_leaf_edge(g, e), edges(g))`.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+    return edges(g)[findall(==(1), [!leaf_edge(g,e) for e in edges(g)])]
+end
+
+"""Get all nodes which aren't leaves and don't have leaf neighbours (these represent internal points in the contraction sequence)"""
+function internal_nodes(g::AbstractGraph)
+    return vertices(g)[findall(==(1), [no_leaf_neighbours(g, v) && !leaf_node(g, v) for v in vertices(g)])]
```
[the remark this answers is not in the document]
mtfishman: Oh I see, this doesn't just exclude the internal vertices, but also excludes the vertices that are neighbors with the internal vertices. Maybe there is a better graph terminology for this, based on height or depth from the base of the tree? I.e. https://en.wikipedia.org/wiki/Tree_(data_structure)#Terminology

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+end
+
+"""Get all nodes which have leaf neighbours but aren't leaves themselves (these represent start/ end points of a contraction sequence)"""
+function external_nodes(g::AbstractGraph)
+    return vertices(g)[findall(==(1), [!no_leaf_neighbours(g, v) && !leaf_node(g,v)  for v in vertices(g)])]
+end
+
+"""Get the vertex bi-partition that a given edge between non-leaf nodes represents"""
+function edge_bipartition(g::AbstractGraph, e)
```
mtfishman, line 96: So this function is a bit tricky to name since it isn't really for generic tree graph, since it assumes information about the leaves are stored within the vertices.

To clarify that, I would call this `contraction_tree_leaf_bipartition` (where in principle there could be a `leaf_bipartition` function that just analyzed the graph itself but did the same thing, which would work for any tree graph).

It hits a bit on a question about terminology. So far we've used `contraction_sequence` as the terminology for the nested vector data structure, so it would be good to have a consistent terminology for the graph representation. We could just go with `contraction_tree` for that, which implies it is a tree graph.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+    return left_bipartition, right_bipartition
+end
+
+"""Given a contraction node, get the keys living on all its neighbouring leaves"""
+function external_node_keys(g::AbstractGraph, v)
+    return [Base.Iterators.flatten(v[findall(==(1), [length(vi) == 1 for vi in v])])...]
+end
+
+"""Given a contraction node, get all keys which are not living on a neighbouring leaf"""
+function external_contraction_node_ext_keys(g::AbstractGraph, v)
```
mtfishman, line 118: This is a bit confusing, it would help to get a more extended explanation for what this is used for.

test/test_contractiontreetograph.jl
```diff
@@ -0,0 +1,70 @@
...
+    ψ = randomITensorNetwork(s; link_space=2)
+    ψψ = flatten_networks(ψ,ψ)
+
+    seq = contraction_sequence(ψψ);
+
+    g_seq = create_contraction_tree(seq)
+
+    #Get all leaf nodes (should match number of tensors in original network)
+    #why does vertices(g_seq)[leaf_node(g_seq, vertices(g_seq))) .== true] not work?"""
+    g_seq_leaves = vertices(g_seq)[findall(==(1), [leaf_node(g_seq,v) for v in vertices(g_seq)])]
```
mtfishman, line 22: Isn't this already a function? It seems better to avoid complicated code like this in a test.

test/test_contractiontreetograph.jl
```diff
@@ -0,0 +1,70 @@
...
+
+    @test length(g_seq_leaves) == n*n
+
+    g_seq_internal_edges = non_leaf_edges(g_seq)
+
+    internal_edge_test = 0
+    #Check all internal edges define a correct bi-partition
+    for eb in g_seq_internal_edges
+      es = edge_bipartition(g_seq, eb)
+      if(length(es) != 2)
```
mtfishman, line 32: Easier to just drop the `internal_edge_test` and write this as `@test length(es) != 2`.

test/test_contractiontreetograph.jl
```diff
@@ -0,0 +1,70 @@
...
+        internal_edge_test += 1
+      end
+
+      es =  vcat(es[1],es[2])
+
+      if(Set([e.I for e in es]) != Set(vertices(ψψ)))
```
mtfishman, line 38: Again, easier to just drop the `internal_edge_test` and write this as `@test issetequal([e.I for e in es], vertices(ψψ))`.

Also, I found it confusing that the collection is called `es` but they are vertices.

test/test_contractiontreetograph.jl
```diff
@@ -0,0 +1,70 @@
...
+        end
+        vs =  vcat(v[1], v[2])
+        if(Set([vsi.I for vsi in vs]) != Set(vertices(ψψ)))
+          vertex_test += 1
+        end
+      end
+
+    end
+
+    @test vertex_test == 0
```
mtfishman, line 68: Same suggestions as above, no need for the `vertex_test` variable, just use `@test` within the loop.

mtfishman:
Thanks @JoeyT1994! It's a very good question you bring up about what the type should be.

I would prefer to not introduce a new type, and I'm hoping with good generic code and good function names we can make it clear when a function should be used on a tree graph (i.e. a `NamedGraph` that happens to have a tree structure) and additionally a tree graph representing a contraction sequence, with internal vertex names that store information about the leaves of the subgraphs. For example, for functions that only make sense for contraction tree graphs or make use of that vertex naming convention, we could use function names of the form `contraction_tree_*`.

mtfishman:
Finally, even though we have that vertex naming convention, it may be better to try to write code that is written for generic graphs and/or tree graphs that don't have any special vertex naming convention, even if the code is a bit slower in principle. That way we build out useful tools for working with general tree graphs, and we can always make specialized functions that make use of the vertex naming convention for cases where it actually makes a performance difference.

src/ContractionTreeToGraph.jl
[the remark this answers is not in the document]
JoeyT1994: Good point. I have now written an `abstractgraph.jl` function which determines the path distance to the closest leaf for a given vertex. I then used this define a function which gets a subset of vertices within a certain distance of a leaf. This functionality should serve the purpose of the `external_node` and `internal_node` functions and go beyond it.

src/ContractionTreeToGraph.jl
```diff
@@ -0,0 +1,120 @@
...
+    left_bipartition = Set(c1) ∈ vdst_set ? c1 : Set(c2) ∈ vdst_set ? c2 : c3
+
+    c1, c2, c3 = [dst(e)[1]..., dst(e)[2]...], [dst(e)[2]..., dst(e)[3]...], [dst(e)[1]..., dst(e)[3]...]
+    right_bipartition = Set(c1) ∈ vsrc_set ? c1 : Set(c2) ∈ vsrc_set ? c2 : c3
+
+    return left_bipartition, right_bipartition
+end
+
+"""Given a contraction node, get the keys living on all its neighbouring leaves"""
+function external_node_keys(g::AbstractGraph, v)
```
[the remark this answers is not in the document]
JoeyT1994: It is supposed to get all the keys directly associated with the neighbouring leaves of that vertex. But I agree it is not clear/ hard to see whether it should really be a function at all.

I wrote it because when you want to figure out the tri-partition a non_leaf vertex represents one has to distinguish between keys which are coming from neighbouring leaves, and those which are not. The keys from neighbouring sources are going to be directly used in the contraction, whilst those from non-neighbouring sources just form 'environments' as opposed to being directly used.

Should I just remove for now?

src/ContractionTreeToGraph.jl
JoeyT1994: This is the opposite of the `external_node_keys` function (get all keys which aren't associated to neighbouring leaves).

src/contraction_tree_to_graph.jl
```diff
@@ -0,0 +1,83 @@
...
+    end
+end
+
+"""Utility functions for the graphical representation of a contraction sequence"""
+
+"""Get the vertex bi-partition that a given edge between non-leaf nodes represents"""
+function contraction_tree_leaf_bipartition(g::AbstractGraph, e)
+
+    if(is_leaf_edge(g, e))
+        println("ERROR: EITHER THE SOURCE OR THE VERTEX IS A LEAF SO EDGE DOESN'T REALLY REPRESENT A BI-PARTITION")
```
mtfishman, line 62: Isn't it still well defined as a bi-partition, where one partition is just the leaf vertex?

src/ContractionTreeToGraph.jl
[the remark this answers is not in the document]
mtfishman: I think I understand. I think this is a good reason to first think about implementing that kind of functionality through generic `AbstractGraph` functions that work for general graphs or tree graphs, and hopefully through functions like that we can implement most functionality like this.

For example, a general graph/tree function for that might just be grabbing a subgraph in each direction away from the specified vertex and then collecting the leaves of each subgraph, which seems like a simpler function to write and easier logic to follow.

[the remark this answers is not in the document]
mtfishman: So for now I'd say get rid of this function and we can try to come up with general `AbstractGraph` functions which help with various situations that arise in tensor network calculations when using contraction trees.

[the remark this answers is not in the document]
JoeyT1994: Okay I will get rid of it and we will go that route.

src/contraction_tree_to_graph.jl
JoeyT1994: Yes good point! I have made this function work for any edge now.

mtfishman:
Looks good, thanks again!

PR #57, opened 2023-01-31
Belief propagation v2

JoeyT1994:
This PR introduces significant improvements to the Belief Propagation code within `ITensorNetworks.jl`

1) `beliefpropagation.jl` is now completely network agnostic and will simply form message tensors, update them etc... independent of the actual network geometry. All partitioning is handled by the recently added `partition.jl` interface.
2) `beliefpropagation.jl` has had the overly specific expectation value functions removed. Instead these have been replaced by `get_environment` and `calculate_contraction` -> which are more general and the workhorses of those functions anyway.
3) The testing and belief propagation example have been updated to reflect these changes.
4) A catch has been added to `subgraph_vertices` in `partition.jl`, which checks if the number of partitions required is 1 (both KaHyPar and Metis error in this case --- although the issue has been raised and is being fixed).
5) A function `unwrap_graph_vertices` has been added to `partition.jl` which recursively goes through a datagraph and fetches all the vertex data within each vertex.
6) A function `split_index` has been added to `abstractitensornetwork.jl` which splits a series of indexes in a given `itensornetwork` and applies the users choice of mapping to them.


mtfishman:
Still to do:

- Rename `unwrap_graph_vertices` to `nested_graph_leaf_vertices`, implement in terms of `hasmethod(vertices, Tuple{...})` to determine the leaf graphs.
- Improve `partition` interface to allow for keyword arguments `nvertices_per_partition` or `vertex_groups`.
- Remove `assert_correct_environment`, plan to just use set functions proposed in #58.

Looking at it more closely, I think a function like `calculate_contraction` is useful. I would think about tweaking the interface a bit but let's keep as-is for now.

src/partition.jl
```diff
@@ -281,10 +281,19 @@ graph from the results of `partition`.
 A graph partitioning backend such as Metis or KaHyPar needs to be installed for this function to work
 if the subgraph vertices aren't specified explicitly.
 """
-function partition(
-  g::AbstractGraph; npartitions=nothing, nvertices_per_partition=nothing, kwargs...
-)
-  return partition(g, subgraph_vertices(g; npartitions, nvertices_per_partition, kwargs...))
+function partition(g::AbstractGraph; npartitions = nothing, nvertices_per_partition = nothing, sgraph_vertices = nothing, kwargs...)
```
mtfishman, line 284: Better to call it `subgraph_vertices`. If it conflicts with the function call `subgraph_vertices(...)` inside the function, you can use `ITensorNetworks.subgraph_vertices(...)` to call the function.

src/partition.jl
```diff
@@ -281,10 +281,19 @@ graph from the results of `partition`.
...
+
+  if(count([npartitions != nothing, nvertices_per_partition != nothing, sgraph_vertices != nothing ]) >= 2)
```
mtfishman, line 286: Could use:
```julia
if count(isnothing, (npartitions, nvertices_per_partition, , subgraph_vertices))
  [...]
end
```
Also I use the style that I don't wrap the check in parenthesis, i.e. please use `if expr` instead of `if(expr)`.

mtfishman: Also, should the check be:
```julia
count(isnothing, (npartitions, nvertices_per_partition, subgraph_vertices)) != 1
```
i.e. you have to specify exactly one of them? What happens if you don't specify any of them?

src/partition.jl
```diff
@@ -281,10 +281,19 @@ graph from the results of `partition`.
...
+    throw(error("ERROR: CANNOT SPECIFY MULTIPLY PARTITIONING OPTIONS, PLEASE GIVE ONLY 1"))
```
mtfishman, line 287: In general we should not use all upper case, just use casing of normal writing (including in comments).

mtfishman: Also, `error` already throws so you shouldn't need `throw`.

JoeyT1994: Sure, I will change that.

src/partition.jl
JoeyT1994: If you don't specify any then it will throw an error too from the call `partition(g, subgraph_vertices(g; npartitions, nvertices_per_partition, kwargs...))` which has its own error message built in. Good though I think to do the check 
`count(isnothing, (npartitions, nvertices_per_partition, subgraph_vertices)) != 1` in this function as it makes the most sense.

src/partition.jl
JoeyT1994: Ah I see, yes that is why I didn't call it that but that is a better way round!

src/partition.jl
```diff
@@ -281,17 +281,17 @@ graph from the results of `partition`.
...
 
 
-  if(sgraph_vertices == nothing)
-    return partition(g, subgraph_vertices(g; npartitions, nvertices_per_partition, kwargs...))
+  if subgraph_vertices == nothing
+    return partition(g, ITensorNetworks.subgraph_vertices(g; npartitions, nvertices_per_partition, kwargs...))
   else
-    return partition(g, sgraph_vertices)
+    return partition(g, subgraph_vertices)
   end
```
mtfishman, line 295: Here I would do:
```julia
if isnothing(subgraph_vertices)
  subgraph_vertices = ITensorNetworks.subgraph_vertices(g; npartitions, nvertices_per_partition, kwargs...)
end

return partition(g, subgraph_vertices)
```
i.e. organize it as processing the inputs/arguments to the function and then calling the function.

JoeyT1994:
Note I also ran this branch through `JuliaFormatter` hence why lots of files appear changed (should just be formatting)

examples/belief_propagation/ApproximateNormsV2.jl
```diff
@@ -0,0 +1,140 @@
...
+  group_partition_vertices,
+  contract_inner,
+  split_index,
+  ising_network,
+  get_environment,
+  find_subgraph,
+  update_all_mts
+
+using OMEinsumContractionOrders
+include("ContractionTreeFunctions.jl")
```
mtfishman, line 24: I don't see this file included in the PR.

mtfishman: Maybe this whole file should be left for a future PR?

JoeyT1994: Yes sorry it wasn't supposed to be included. I accidentally added it and the code-workspace file. Will take them out with .gitignore

mtfishman: You shouldn't add a normal file like that to `.gitignore`, that's only for files you know you will always want to ignore (like system files, ones generated by text editors, etc.). You should be able to just remove this one with `git rm`, and just not include it with `git add` in the first place if you're not ready to commit it.

JoeyT1994:
> @JoeyT1994 I formatted the package in #61. Could you merge those changes? Hopefully then this PR will only show your own changes and not the formatting changes as well.

Great, yes I have done that merge.

src/beliefpropagation.jl
```diff
@@ -99,237 +99,54 @@ function update_all_mts(
...
-  num_tensors_to_contract = ITensor[]
-  denom_tensors_to_contract = ITensor[]
-
-  if (subgraph1 == subgraph2)
-    connected_subgraphs = neighbors(subgraphs, subgraph1)
-    for k in connected_subgraphs
-      push!(num_tensors_to_contract, subgraphs[k => subgraph1])
-      push!(denom_tensors_to_contract, subgraphs[k => subgraph1])
-    end
+#CAN SWITCH DIR TO GET OUTGOING ENVIRONMENT BUT THIS WON'T BE QUITE RIGHT IF WE ARE DOING GENERAL BELIEF PROPAGATION
```
mtfishman, line 105: This should get moved above the docstring (also please make it normal-cased, not upper case).

PR #63, opened 2023-02-08
Full Update

JoeyT1994:
This PR adds functionality to `ITensorNetworks.jl` for applying a full update, i.e. applying a two-site gate to an `ITensorNetwork `whilst accounting for the environment around that gate. 

Changes:

1) The `apply()` function now checks whether to do a QR decomposition before applying the gate. This is based on memory (i.e. will it take up less memory to do the QR decomposition first).
2) A new `apply_fullupdate.jl` file which contains the workhorse
``` 
apply_fullupdate(o::ITensor,ψ::AbstractITensorNetwork, envs::Vector{ITensor}; nsweeps=10,maxdim=nothing,
normalize=false,print_fidelity_loss=false,isposdef=true) 
```
for doing a full update based on an ITensornetwork `ψ`, a gate `o` and a Vector of ITensors which represents the environment. If the environment vector is empty then `apply()` is called instead.
Note that the environments are expected to have been formed over `ψ ⊗ prime(dag(ψ); sites=[])` and this is left to the user to be dealt with beforehand (for now).

3) A corresponding test has been added.


src/apply.jl
```diff
@@ -25,16 +25,43 @@ function ITensors.apply(
...
     end
-    oψᵥ = apply(o, ψ[v⃗[1]] * ψ[v⃗[2]])
-    ψᵥ₁, ψᵥ₂ = factorize(
-      oψᵥ, inds(ψ[v⃗[1]]); cutoff, maxdim, tags=ITensorNetworks.edge_tag(e)
-    )
+
+    #Check whether to do a memory efficient QR first (do both sites combined for now, although we could split if we wish)
+    #if no maxdim is provided just do it based on inds counting, if it is provided do it based on whichever takes less memory
+    d1, d2 = prod(dim.(inds(ψ[v⃗[1]]; tags="Site"))),
+    prod(dim.(inds(ψ[v⃗[2]]; tags="Site")))
```
mtfishman, line 32: We should always avoid using tags for this kind of thing (at this point I only use tags for printing, defining operators, and for infinite tensor networks).

Probably you can use some index set logic and graph analysis.

test/test_fullupdate.jl
```diff
@@ -0,0 +1,76 @@
...
+  x = (rand(N, N) + rand(N, N) * im) / sqrt(2)
+  f = qr(x)
+  diagR = sign.(real(diag(f.R)))
+  diagR[diagR .== 0] .= 1
+  u = f.Q * diagm(diagR)
+
+  return u
+end
+
+function ITensors.op(::OpName"RandomTwoSiteU", ::SiteType"S=1/2")
```
mtfishman, line 27: There is already a `"RandomUnitary"` gate which works for any number of sites: https://github.com/ITensor/ITensors.jl/blob/v0.3.25/src/physics/site_types/generic_sites.jl

src/apply.jl
JoeyT1994: Okay, I have written it now so that it does it based on the indices unique to `ψ[v⃗[1]]` compared to its neighbours.

I.e:

```
v1_siteinds = setdiff(inds(ψ[v⃗[1]]), flatten(unique(inds.([ψ[v] for v in neighbors(ψ, v⃗[1])]))))
v2_siteinds = setdiff(inds(ψ[v⃗[2]]), flatten(unique(inds.([ψ[v] for v in neighbors(ψ, v⃗[2])]))))
d1, d2 = prod(dim.(v1_siteinds)), prod(dim.(v2_siteinds))
```

Was that along the lines you were thinking?

mtfishman: I thought the overall logic could be a bit simpler, like:
```julia
dim_v1 = dim(uniqueinds(ψ[v⃗[1]], ψ[v⃗[2]])
dim_v2 = dim(uniqueinds(ψ[v⃗[2]], ψ[v⃗[1]])
dim_shared = dim(commoninds(ψ[v⃗[1]], ψ[v⃗[2]]))
```
and then you can choose whether or not to QR based on those three.

So it doesn't seem like you should need the site/external indices in particular. However if you did (or need it for something else), there is the function `uniqueinds(tn::AbstractITensorNetwork, vertex)`: https://github.com/mtfishman/ITensorNetworks.jl/blob/7bdcc89b672510cdcab6e4bb7618f58c074db586/src/abstractitensornetwork.jl#L230-L232

src/apply.jl
```diff
@@ -25,16 +25,49 @@ function ITensors.apply(
...
+    prod(dim.(uniqueinds(uniqueinds(ψ[v⃗[2]], ψ[v⃗[1]]), inds(ψ[v⃗[2]]; tags="Site"))))
+    if (maxdim == nothing && length(inds(ψ[v⃗[1]])) + length(inds(ψ[v⃗[2]])) <= 6) ||
+      (maxdim != nothing && maxdim * maxdim * d1 * d2 >= χv1 * χv2)
+      oψᵥ = apply(o, ψ[v⃗[1]] * ψ[v⃗[2]])
+      ψᵥ₁, ψᵥ₂ = factorize(
+        oψᵥ, inds(ψ[v⃗[1]]); cutoff, maxdim, tags=ITensorNetworks.edge_tag(e)
+      )
+    else
+      Qv1, Rv1 = factorize(
+        ψ[v⃗[1]], uniqueinds(uniqueinds(ψ[v⃗[1]], ψ[v⃗[2]]), inds(ψ[v⃗[1]]; tags="Site"))
```
mtfishman, line 51: Again, I would hope we could avoid using `tags` to filter indices. I thought `uniqueinds(ψ[v⃗[1]], ψ[v⃗[2]])` would be sufficient here?

src/apply_fullupdate.jl
```diff
@@ -0,0 +1,231 @@
...
+  )
+  term2_tns = vcat(
+    [
+      p_cur,
+      q_cur,
+      noprime(prime(dag(p_cur)); tags="Site"),
+      noprime(prime(dag(q_cur)); tags="Site"),
+    ],
+    envs,
+  )
```
mtfishman, line 80: This kind of code using too much priming and tag filtering raises red flags for me. Is there a way to do this by analyzing the graph/environment and using graph and index set operations?

src/apply.jl
```diff
@@ -25,16 +25,49 @@ function ITensors.apply(
...
+      Qv2, Rv2 = factorize(
+        ψ[v⃗[2]], uniqueinds(uniqueinds(ψ[v⃗[2]], ψ[v⃗[1]]), inds(ψ[v⃗[2]]; tags="Site"))
+      )
+      Rv1_new, Rv2_new = factorize(
+        noprime(Rv1 * o * Rv2), inds(Rv1); cutoff, maxdim, tags=ITensorNetworks.edge_tag(e)
+      )
+      ψᵥ₁ = Qv1 * Rv1_new
+      ψᵥ₂ = Qv2 * Rv2_new
+    end
+
```
mtfishman, line 62: To clarify the code structure, this is all still a "simple update" (i.e. no environment) but uses QR factorizations to improve performance?

It seems better to split this part into a separate function which you call inside this function. Also, shouldn't this same code and logic apply to the "full update"?

It seems like this is all pointing towards just having a single `apply` code that skips some steps in the "simple update" case, for example has a special case where the environment is empty so doesn't do an optimization and just a factorization.

mtfishman:
Thanks @JoeyT1994, I think the code could use some cleanup and simpifying, I've pointed out a few places in the comments. Also we should probably chat about this, it seems awkward to have two different versions of `apply`, it would be much easier for development going forward to have just one version which has a limit of "simple update" that skips the optimization part. Otherwise it seems like we are doubling effort, like how both versions need the logic of whether or not to do a QR based on the dimensions (but that code looks very different in the two versions from what I see above).

JoeyT1994:
Yes I agree that combining these two functions into one makes a lot of sense. I didn't want to mess with the `apply()` function but will have a go at combining them if you think that makes sense from here on. I will also try to simplify certain sections based on your suggestions.
Happy to discuss this next week.

src/apply.jl
JoeyT1994: Agreed this should make it simpler. But isn't the condition for whether to do a QR: `dim_v1*dim_v2 < d*d*dim_shared*dim_shared`? So there is the need to work out `d`, the dimension of the physical index?

src/apply.jl
JoeyT1994: But you don't want the "Site" index to go onto `Qv1` no? Which is what would happen if you just did `uniqueinds(ψ[v⃗[1]], ψ[v⃗[2]])`

JoeyT1994: I.e. you need `uniqueinds(uniqueinds(ψ[v⃗[1]], ψ[v⃗[2]]), uniqueinds(ψ, v⃗[1]))` (if you want to avoid tag specification - which makes sense).

src/apply.jl
mtfishman: Ah right. So in this case the "site indices" are really defined as the ones that are shared with the gate, so we should use the gate in the logic, i.e.:
```julia
outer_dim_v1 = dim(uniqueinds(ψ[v⃗[1]], o, ψ[v⃗[2]]))
dim_v1 = dim(commoninds(ψ[v⃗[1]], o))
shared_dim = dim(commoninds(ψ[v⃗[1]], ψ[v⃗[2]]))
```
and the comparison is `outer_dim_v1 > shared_dim * dim_v1`. Maybe I got the details wrong but hopefully you get the point that we should use the graph connectivity and set index logic as much as possible.

src/apply.jl
mtfishman: Right I forgot you want the site indices to go onto `Rv1`.

src/apply.jl
mtfishman: Maybe I have it wrong about using `o` to determine the site indices vs. `uniqueinds(ψ, v⃗[1])`, those can be different if the gate/operator is only applied to a subset of the site indices of a vertex, like applying a gate to a density matrix.

mtfishman:
Yeah I think that's the way to go, if it seems reasonable to take on in this PR.

src/apply.jl
JoeyT1994: I think what you have there makes the most sense? If the gate is only being applied to a subset of the site indices of a vertex then we should treat the remaining site indices as ones we want to put onto Q and not R in the decomposition? Hence the lines you wrote there would work?

src/apply_fullupdate.jl
JoeyT1994: I will clean it up and try to do it more naturally.

JoeyT1994:
Yeah I don't think it will be too much work.

JoeyT1994:
I have now combined everything into `apply.jl` and the `apply()` function will do either a simple update or full update depends on the arguments.

src/apply.jl
```diff
@@ -268,7 +268,7 @@ function optimise_p_q(
 
   fend = print_fidelity_loss ? fidelity(envs, p_cur, q_cur, p, q, o) : 0
 
-  if (print_fidelity_loss && real(fend - fstart) <= 0.0 && nsweeps >= 1)
+  if (print_fidelity_loss && real(fend - fstart) < -1e-15 && nfullupdatesweeps >= 1)
```
mtfishman, line 271: You could use:
```julia
diff = real(fend - fstart)
if print_fidelity_loss && diff < -eps(diff) && nfullupdatesweeps >= 1
  # [...]
end
```
which would adjust the threshold for the element type instead of hardcoding `-1e-15`.

src/apply.jl
```diff
@@ -25,16 +29,52 @@ function ITensors.apply(
...
-    ψᵥ₁, ψᵥ₂ = factorize(
-      oψᵥ, inds(ψ[v⃗[1]]); cutoff, maxdim, tags=ITensorNetworks.edge_tag(e)
-    )
+
+    outer_dim_v1, outer_dim_v2 = dim(uniqueinds(ψ[v⃗[1]], o, ψ[v⃗[2]])),
+    dim(uniqueinds(ψ[v⃗[2]], o, ψ[v⃗[1]]))
+    dim_shared = dim(commoninds(ψ[v⃗[1]], ψ[v⃗[2]]))
+    d1, d2 = dim(commoninds(ψ[v⃗[1]], o)), dim(commoninds(ψ[v⃗[2]], o))
+    if outer_dim_v1 * outer_dim_v2 <= dim_shared * dim_shared * d1 * d2
+      Qᵥ₁, Rᵥ₁ = ITensor(1.0), copy(ψ[v⃗[1]])
```
mtfishman, line 38: If we use `ITensor(true)` here it will make it work better with arbitrary element types (`ITensor(1.0)` is implicitly `Float64`, which will then force everything to be double precision).

mtfishman:
Looks good, thanks!

PR #64, opened 2023-02-09
[binary_tree_partition] [2/2]: Add `binary_tree_partition`

LinjianMa:
Add 
```julia
function binary_tree_partition(tn::ITensorNetwork, inds_btree::Vector)
```
Given an input `tn` and a rooted binary tree of indices, return a partition of `tn` with the same binary tree structure as `inds_btree`.


mtfishman:
Looks good, thanks!

PR #42, opened 2023-01-08
[binary_tree_partition] [1/2]: Add mincut helper functions and introduce `_binary_tree_partition_inds`

LinjianMa:
1. Overload `GraphsFlows.mincut` and `mincut_partitions`
2. Add `inds_binary_tree` and `inds_mps_order`. For a given network, the first one outputs a binary tree of uncontract inds, and the second one outputs an inds order that can be used to approximate the network by an MPS.

PR #66, opened 2023-02-14
Implement `approx_itensornetwork`

LinjianMa:
**Interfaces:**

```julia
"""
Approximate a `binary_tree_partition` into an output ITensorNetwork
with the same binary tree structure. `root` is the root vertex of the
pre-order depth-first-search traversal used to perform the truncations.
"""
function approx_itensornetwork(
  ::Algorithm"density_matrix",
  binary_tree_partition::DataGraph;
  root,
  cutoff=1e-15,
  maxdim=10000,
  contraction_sequence_alg,
  contraction_sequence_kwargs,
)
"""
Approximate a given ITensorNetwork `tn` into an output ITensorNetwork with `output_structure`.
`output_structure` outputs a directed binary tree DataGraph defining the desired graph structure.
"""
function approx_itensornetwork(
  ::Algorithm"density_matrix",
  tn::ITensorNetwork,
  output_structure::Function=path_graph_structure;
  cutoff,
  maxdim,
  contraction_sequence_alg,
  contraction_sequence_kwargs,
)
"""
Approximate a given ITensorNetwork `tn` into an output ITensorNetwork
with a binary tree structure. The binary tree structure is defined based
on `inds_btree`, which is a directed binary tree DataGraph of indices.
"""
function approx_itensornetwork(
  ::Algorithm"density_matrix",
  tn::ITensorNetwork,
  inds_btree::DataGraph;
  cutoff=1e-15,
  maxdim=10000,
  contraction_sequence_alg="optimal",
  contraction_sequence_kwargs=(;),
)
```

Note: The current code is designed so that its performance is perfect (as I’m aware of, this should already be the state-of-the-art density matrix algorithm implementation, in terms of the caching implementations)

**Some internally-used definitions:**

density matrix:
<details>
  <summary>Example</summary>

```julia
"""
Example:
  Consider a tensor network below,
    1
    /\
   9  2
  /   /\
     3  6
    /|  /\
   4 5 7  8
  /  | |   \
  The density matrix for the edge `NamedEdge(2, 3)` squares the subgraph with vertices 3, 4, 5
     |
     3
    /|
   4 5
   | |
   4 5
   |/
   3
   |
  The density matrix for the edge `NamedEdge(5, 3)` squares the subgraph
    with vertices 1, 2, 3, 4, 6, 7, 8, 9
      1
      /\
     /  2
    /   /\
   /   3  6
  9   /|  /\
  |  4   7  8
  |  |   |  |
  |  4   7  8
  |  |/  | /
  |  3   6
  |  |  /
  |  | /
  |  2
  9 /
  |/
  1
  The density matrix for the edge `NamedEdge(4, 3)` squares the subgraph
    with vertices 1, 2, 3, 5, 6, 7, 8, 9
      1
      /\
     /  2
    /   /\
   /   3  6
  9   /|  /\
  |    5 7  8
  |    | |  |
  |    5 7  8
  |  |/  | /
  |  3   6
  |  |  /
  |  | /
  |  2
  9 /
  |/
  1
"""
```
</details>

partial density matrix:

<details>
  <summary>Example</summary>

```julia
"""
Example:
  Consider a tensor network below,
    1
    /\
   9  2
  /   /\
     3  6
    /|  /\
   4 5 7  8
  /  | |   \
  The partial density matrix for the Edge set `Set([NamedEdge(2, 3), NamedEdge(5, 3)])`
    squares the subgraph with vertices 4, and contract with the tensor 3
    |
    3
   /
  4 - 4 -
  The partial density matrix for the Edge set `Set([NamedEdge(4, 3), NamedEdge(5, 3)])`
    squares the subgraph with vertices 1, 2, 6, 7, 8, 9, and contract with the tensor 3
      1
      /\
     /  2
    /   /\
   /   3  6
  9   /|  /\
  |      7  8
  |      |  |
  |      7  8
  |      | /
  |      6
  |     /
  |  | /
  |  2
  9 /
  |/
  1
  The density matrix for the Edge set `Set([NamedEdge(4, 3), NamedEdge(2, 3)])`
    squares the subgraph with vertices 5. and contract with the tensor 3
    |
    3
   /
  5 - 5 -
"""
```
</details>

**Calculate `_DensityMatrix` using `_PartialDensityMatrix`:**

`_PartialDensityMatrix` is introduced for the efficiency purpose. For a density matrix `dm` with root `v` and children `c1,c2`, we can calculate `dm` via two `_PartialDensityMatrix` `pdm1`, `pdm2`, where `pdm1` is defined on `v` with child `c1` and `pdm2` defined on `v` with child `c2`. The contraction can be done by property sim the inds of `pdm2`, then contract with `pdm1`. In this way, everytime when we want to use a specific `dm` we don’t need to re-compute it, since `pdm1` or `pdm2` can be cached and we can reuse them.

**The density matrix algorithm core part is below:**
```julia
 alg_graph = _DensityMartrixAlgGraph(partition, out_tree, root)
  output_tn = ITensorNetwork()
  for v in post_order_dfs_vertices(out_tree, root)[1:(end - 1)]
    U = _rem_vertex!(
      alg_graph, v; cutoff, maxdim, contraction_sequence_alg, contraction_sequence_kwargs
    )
    add_vertex!(output_tn, v)
    output_tn[v] = U
  end
``` 
each ` _rem_vertex!` will remove `v` in the partition, and output its projector `U`. This `U` is then used to update the output tree tensor network. `_rem_vertex!` also update and use all the caches.

**Discussions:**
Note that it’s very possible that the `_DensityMatrixAlgCaches` has similar concepts as the `ITensorNetworkCache` @mtfishman plans to work on. If so I would be happy to discuss and we can think about how to design it in a better way.


src/itensors.jl
```diff
@@ -41,3 +41,20 @@ trivial_space(x::Index) = _trivial_space(x)
...
 trivial_space(x::Tuple{Vararg{Index}}) = trivial_space(first(x))
+
+"""
+Given an input tensor and a Dict (ind_to_newind), replace inds of tensor that are also
+keys of ind_to_newind to the value of ind_to_newind.
+Note that it is the same as
+ITensors.replaceinds(tensor, collect(keys(ind_to_newind)) => collect(values(ind_to_newind))).
+Based on benchmark, this implementation is more efficient when the size of ind_to_newind is large.
+TODO: we can remove this function once the original replaceinds performance is improved.
+"""
```
LinjianMa, line 52: I realized another problem with the current `replaceinds` implementation is that `collect(keys(ind_to_newind))` cannot be too large: https://github.com/ITensor/ITensors.jl/blob/8535b7fd45915bcd628b5544b84f3ff6189011f9/NDTensors/src/tupletools.jl#L4

LinjianMa:
@mtfishman Not sure what happened with the documentation CI, any ideas?

src/itensors.jl
mtfishman: Thanks, yeah we should fix that in ITensors.jl, I've seen this pop up other places as well. Using `intersect`/`commoninds` like you did here would be a simple fix.

mtfishman:
Not sure, I hadn't even realized we had doctests. I don't think that doctest even makes sense since it relies on https://github.com/GiggleLiu/ITensorContractionOrders.jl which was moved into this package anyway, so we could just remove that doctest (https://github.com/mtfishman/ITensorNetworks.jl/blob/64f17f1d7a567d2d6b7c56e551f84f2ff34a1f25/src/requires/omeinsumcontractionorders_itensorcontractiontree.jl#L12-L20).

mtfishman:
This makes me think about a slightly different interface and code structure. Here is a summary:
```julia
function approx_itensornetwork(tn::ITensorNetwork, output_structure::IndsNetwork; alg="density_matrix", cutoff, maxdim)
  partitioned_tn = partition(tn, output_structure) # Partition the network based on the `output_structure`, outputs a `DataGraph`
  return approx_itensornetwork(partitioned_tn; alg, cutoff, maxdim)
end

function approx_itensornetwork(partitioned_tn::DataGraph; alg="density_matrix", cutoff, maxdim)
  return approx_itensornetwork(Algorithm(alg), partitioned_tn; cutoff, maxdim)
end

function approx_itensornetwork(alg::Algorithm"density_matrix", partitioned_tn::DataGraph; cutoff, maxdim)
  @assert is_tree(partitioned_tn) # For now restrict the desired tensor network structure to be a tree
  # Implementation of the density matrix algorithm on the partitioned tensor network
end

function approx_itensornetwork(alg::Algorithm"orthogonalize", partitioned_tn::DataGraph; cutoff, maxdim)
  @assert is_tree(partitioned_tn) # For now restrict the desired tensor network structure to be a tree
  # Implementation of the orthogonalization algorithm on the partitioned tensor network.
  # Probably has overlap with the density matrix algorithm so code should be shared through generic functions.
end

function approx_itensornetwork(tn::ITensorNetwork, output_structure::Function=path_graph_structure; alg, cutoff, maxdim)
  output_structure_indsnetwork = output_structure(tn) # Outputs an `IndsNetwork`
  return approx_itensornetwork(tn, output_structure_indsnetwork; alg, cutoff, maxdim)
end

function path_graph_structure(tn::ITensorNetwork)
  # Outputs a maximimally unbalanced binary tree IndsNetwork defining the desired graph structure
end

function binary_tree_structure(tn::ITensorNetwork)
  # Outputs a binary tree IndsNetwork defining the desired graph structure
end
```
Besides some restructuring/renaming, I think the main difference would be using an `IndsNetwork` to define the desired structure of the output tensor network instead of a nested tree structure like you did in #42 and #64. The reasons would be:
1. I think an actual graph structure like `IndsNetwork` is easier to work with and easier for users to construct and understand.
2. It would anticipate generalizing the function to the case where the desired output network isn't a tree, but some more general tensor network (such as approximating the contraction of a 2D-like network as a PEPS/2D tensor network).

Let me know if that makes sense. Some of the proposal may clash with realities about the flow/logic of the algorithm so that would be helpful to know as well.

src/approx_binarytree_itensornetwork.jl
```diff
@@ -0,0 +1,512 @@
...
+Contract of a vector of tensors, `network`, with a contraction sequence generated via sa_bipartite
+"""
+function _optcontract(network::Vector)
+  @timeit_debug ITensors.timer "[approx_binary_tree_itensornetwork]: _optcontract" begin
+    if length(network) == 0
+      return ITensor(1.0)
+    end
+    @assert network isa Vector{ITensor}
+    @timeit_debug ITensors.timer "[approx_binary_tree_itensornetwork]: contraction_sequence" begin
+      seq = contraction_sequence(network; alg="sa_bipartite")
```
LinjianMa, line 209: @mtfishman The reason I import `OMEinsumContractionOrders` is that I use a subroutine that automatically contract a vector of tensors with reasonable contraction sequence, and the one I'm using here is `sa_bipartite`. I'm not sure if we can have this without directly `using OMEinsumContractionOrders`

mtfishman: A user would have to load `OMEinsumContractionOrders` before loading `ITensorNetworks` to use functionality that requires `OMEinsumContractionOrders` (such as your function `approx_binary_tree_itensornetwork`), i.e.:
```julia
using OMEinsumContractionOrders
using ITensorNetworks

tn = approx_binary_tree_itensornetwork(...)
```
so that we don't have to depend directly on `OMEinsumContractionOrders` in the library. It's a way to make it so that not everyone depends on certain libraries if they don't need the functionality that is enabled by that library.

I would consider depending directly on `OMEinsumContractionOrders` since it's not a heavy dependency but let's hold off on that decision for future development.

LinjianMa:
@mtfishman The suggestions above as to the interfaces make sense to me, here are some details I would like to bring up:
1. In addition to the tree structure defined by `indsnetwork`, the density matrix algorithm and also the orthogonalize algorithm needs another input `root` to indicate the root vertex for the bfs traversal. We can add this in the function argument, but I'm not sure if this is always useful (for non-tree approximation we may not need this).
2. The current density matrix algorithm assumes that the bfs tree of `indsnetwork` with specific `root` is a binary tree. It's possible to generalize it to arbitrary trees, but it takes time for an efficient implementation (in particular, the caching logic needs to be designed carefully). So for now, I suggest we restrict the density matrix algorithm to only work with binary trees, and we could add a function like `is_binary_tree` to make sure the structure is allowed
3. As to `partition(tn, output_structure)`: for now with the density matrix algorithm the partition function is called `binary_tree_partition`. We can make it a specialized `partition` function, for example
```julia
partition(alg::Algorithm"mincut_recursive_bisection", tn::ITensorNetwork, output_structure::IndsNetwork, root)
```

src/approx_binarytree_itensornetwork.jl
LinjianMa: that makes sense!

mtfishman:
That's helpful, thanks. Maybe what we could do is use a directed `DataGraph` instead of an `IndsNetwork`, which would then have to be an [arborescence](https://en.wikipedia.org/wiki/Arborescence_(graph_theory))/directed rooted tree in order for it to work with the current algorithm.

Definitely fine to keep it working only for binary trees, good idea to add a check for that. Also sounds like a good plan to specialize on `partition`.

src/approx_binarytree_itensornetwork.jl
mtfishman: We could also pass through a `contraction_sequence_alg` keyword argument from the top level function `approx_itensornetwork` so users could choose a backend (also `contraction_sequence_kwargs=(;)` which would be a `NamedTuple` storing keyword arguments to pass to `contraction_sequence` to customize even more).

LinjianMa: SG, done

src/binary_tree_partition.jl
```diff
@@ -41,10 +41,11 @@ Return the root vertex of a directed tree data graph
 end
 
 """
-Check if a data graph is a directed binary tree
+Check if a named graph is a directed binary tree
+  TODO: we may want to move this to `NamedGraphs.jl`
 """
-@traitfn function _is_directed_binary_tree(graph::AbstractDataGraph::IsDirected)
-  if !is_tree(undirected_graph(underlying_graph(graph)))
+@traitfn function _is_directed_binary_tree(graph::AbstractNamedGraph::IsDirected)
```
mtfishman, line 47: Is this more specifically checking if it is a [rooted directed tree](https://en.wikipedia.org/wiki/Arborescence_(graph_theory)) (also known as an arborescence)? Since of course there are directed trees that are not rooted, i.e. [polytrees](https://en.wikipedia.org/wiki/Polytree). All of this terminology is a bit crazy but I think it would would be best to be as specific as possible so we can understand it when we read it later, so I would name this `_is_rooted_directed_binary_tree` if it also implicitly checks if it is rooted.

src/binary_tree_partition.jl
```diff
@@ -143,9 +152,8 @@ function _remove_deltas(tn::ITensorNetwork, deltas::Vector{ITensor})
       _root_union!(ds, find_root!(ds, i2), find_root!(ds, i1))
     end
   end
-  tn = map_data(
-    t -> replaceinds(t, deltainds => [find_root!(ds, i) for i in deltainds]), tn; edges=[]
-  )
+  sim_deltainds = [find_root!(ds, i) for i in deltainds]
```
mtfishman, line 155: I got confused by the the function name `_remove_deltas`, maybe we could call it `_contract_deltas`?

Also, it makes more sense to me to just pass a single tensor network, i.e.:
```julia
function _contract_deltas(tn::ITensorNetwork)
  # [...]
end
```
and then analyze which tensors are deltas. It looks like we don't have a function for that yet in NDTensors/ITensors but we could define:
```julia
is_delta(it::ITensor) = is_delta(tensor(it))
is_delta(t::Tensor) = false
is_delta(t::UniformDiagTensor) = isone(getdiagindex(t, 1))
```
(for now in ITensorNetwork.jl and then we can move it to NDTensors.jl/ITensors.jl). Note that `UniformDiagTensor` is defined in `NDTensors` and may not be exported. Also, it would be fairly easy to extend to the case where the diagonal elements aren't one (i.e. a scaled delta tensor) by scaling the network by the uniform diagonal values but we could leave that for future work since that isn't a common situation anyway.

src/binary_tree_partition.jl
```diff
@@ -219,7 +227,12 @@ function partition(
     end
   end
   tn_deltas = ITensorNetwork(vcat(output_deltas_vector...))
-  return partition(ITensorNetwork{Any}(disjoint_union(out_tn, tn_deltas)), subgraph_vs)
+  par = partition(ITensorNetwork{Any}(disjoint_union(out_tn, tn_deltas)), subgraph_vs)
```
mtfishman, line 230: Was there an issue using `ITensorNetwork` instead of `ITensorNetwork{Any}`, i.e. let it try to infer the vertex type?

src/itensors.jl
```diff
@@ -41,3 +41,20 @@ trivial_space(x::Index) = _trivial_space(x)
...
+function ITensors.replaceinds(tensor::ITensor, ind_to_newind::Dict{<:Index,<:Index})
```
mtfishman, line 53: Could you raise an issue in `ITensors.jl` reminding us to fix this over there?

LinjianMa: SG, will do

src/binary_tree_partition.jl
LinjianMa: Yes renaming it to `_contract_deltas` makes sense, `_remove_deltas` is a bit confusing.

I'm actually also thinking about the name `_simplify_deltas`, so the function would be 
```julia
function _simplify_deltas(tn::ITensorNetwork)
  # [...]
  return new_tn, simplified_deltas
end
```
For example, when the input `tn` contains tensors `randomITensor(i), randomITensor(j), delta(k,l), delta(l,m)`, the output `new_tn` will contain `randomITensor(i), randomITensor(j)` and `splited_deltas` contains `delta(k,m)`.

The reason we want to split `simplified_deltas` and `new_tn` in the function output is that there can be new tensors in 
`simplified_deltas`, and to add them into `new_tn` we need to give each of them a vertex name. It's not easy to have a general naming convention to me so I think it's easier to split it out from `new_tn`.

LinjianMa: actually please ignore my last message, there are simple ways to fix my concern. The AP
```julia
function _contract_deltas(tn::ITensorNetwork)
  # [...]
  return new_tn
end
```
should work well.

src/itensors.jl
LinjianMa: ITensors.jl issue: https://github.com/ITensor/ITensors.jl/issues/1078

src/binary_tree_partition.jl
LinjianMa: @mtfishman Refactoring `_contract_deltas` involves multiple changes, so I opened a new PR for that: #73 
Once that is merged I will then update this PR.

src/binary_tree_partition.jl
LinjianMa: @mtfishman thanks, yes these make sense. I would like to write it in a better way so refactored these set of functions in #74 , I will update this PR once that is merged

src/binary_tree_partition.jl
LinjianMa: The reason I force it to be `ITensorNetwork{Any}` is that in `approx_itensornetwork` I update each partition inplace during the density matrix algorithm, and that update sometimes will assign a new itensornetwork type to each partition.

But it's right that forcing the type here doesn't make sense. I made a modification to update the datatype of partition inside `approx_itensornetwork` right before the update of density matrix algorithm, so the problem should be fixed now.

LinjianMa:
@mtfishman could you retake a look at this?

mtfishman:
@LinjianMa yes I was trying to make some time for this, I should have time in the next few days.

There is a lot of code in `approx_itensornetwork.jl`, can you write up a summary of the code logic in the first post, and include any subtle design choices that I should focus on?

Also, I have a more general question about how this fits into the over all algorithm. My current understanding is that the algorithm can be summarized as follows.

**Algorithm**
Contract the subgraphs `Algorithm Input 2` of a tensor network `Algorithm Input 1` according to the subgraph contraction sequence `Algorithm Input 3`, by traversing the contractions sequence and performing pairwise contractions of subgraphs using the `Subalgorithm` below.

**Algorithm Inputs**

- `Algorithm Input 1`: A tensor network we want to approximately contract.
- `Algorithm Input 2`: A partitioning of `Algorithm Input 1` into subgraphs.
- `Algorithm Input 3`: A subgraph contraction sequence for the subgraphs `Algorithm Input 2`.

**Subalgorithm**
For each pairwise contraction of two subgraphs of `Algorithm Input 2` according to the contraction sequence `Algorithm Input 3`, approximately contract them using the density matrix algorithm, outputting a binary tree that approximates the contraction of the two subgraphs.

**Subalgorithm Inputs**

- `Subalgorithm Input 1`: Two subgraphs from `Algorithm Input 2`, determined by traversing the contraction sequence `Algorithm Input 3`.
- `Subalgorithm Input 2`: A desired binary tree output structure (determined for example from the function `binary_tree_structure` introduced in #69, which takes as an input the two subgraphs `Subalgorithm Input 1`).

**Questions**

1. Is the function `approx_itensornetwork` introduce in this PR an implementation of the `Subalgorithm`? If so, am I correct to assume that `Algorithm` will be implemented in a future PR, as a simple function that traverses the contraction sequence tree `Algorithm Input 3` and performs pairwise contractions of subgraphs from `Algorithm Input 2` using the `Subalgorithm` (`approx_itensornetwork`)?
2. From previous meetings, my understanding was that the desired binary tree output structure (`Subalgorithm Input 2`) relied on information of the overall graph connectivity, i.e. made use of the overall graph structure of whatever part of original tensor network  (`Algorithm Input 1`) remains uncontracted in order to determine the binary tree which helps minimize the number of swaps needed for future subgraph contractions through the `Subalgorithm`. That doesn't seem to be the case in the current code design, could you clarify that for me?

LinjianMa:
@mtfishman Sure, I will summarize `approx_itensornetwork.jl`

For your questions:
1. Yes, this is correct.
2. Note that we have an interface
```julia
function approx_itensornetwork(
  tn::ITensorNetwork,
  inds_btree::DataGraph;
  alg::String,
  cutoff=1e-15,
  maxdim=10000,
  contraction_sequence_alg="optimal",
  contraction_sequence_kwargs=(;),
)
```
to let `approx_itensornetwork` be used in  `Algorithm Input 1`, we will let `inds_btree` be a directed tree of indices decided by some other functions, and `inds_btree` will be chosen based on the original tensor network.

So the interface below 
```julia
function approx_itensornetwork(
  tn::ITensorNetwork,
  output_structure::Function;
  alg::String,
  cutoff=1e-15,
  maxdim=10000,
  contraction_sequence_alg="optimal",
  contraction_sequence_kwargs=(;),
)
```
will not be used there since `output_structure` only knows the structure of `tn`. But this can still be useful so we keep this interface.

mtfishman:
Ok, so to summarize: `approx_itensornetwork` can accept any binary tree structure, which _could_ be chosen based on the rest of the tensor network but that logic isn't in this PR. So that will be some extra logic that will be implemented in a future PR, and `approx_itensornetwork` and `binary_tree_structure` (introduced in #69) are written generally enough so that they can be used once that logic is implemented. Is that correct? If so, that helps me understand the scope and context of this PR.

The part I'm still confused about is that it seems like in order to implement the logic of determining the output structure from the rest of the graph it would affect the function `binary_tree_structure` introduced in #69, which from what I can tell doesn't implement that logic.

LinjianMa:
@mtfishman yes that's correct.

Note that the current `binary_tree_structure` and `path_graph_structure` functions are also general enough. The general API is
```julia
function binary_tree_structure(tn::ITensorNetwork, outinds::Vector{<:Index})
```
where `outinds` is part of `noncommoninds` of `tn`, and this can be used to get the tree ordering of part of the noncommoninds. This function and also some other logic can help us decide the swap-efficient binary tree structures.

LinjianMa:
@mtfishman the first post is updated

mtfishman:
Thanks for the summary @LinjianMa, that's helpful. I'll review this PR by the end of the week.

LinjianMa:
@mtfishman The PR has been updated, please let me know if you have any other questions!

mtfishman:
Looks great, thanks for all of the documentation and thanks for iterating on the code! This will be really useful. 
I'll go ahead and merge.

Issue #60, opened 2023-02-02
Improvements to `alternating_update`


mtfishman:
Here's an issue to track improvements to make to the general tensor network solver code, `alternating_update`, following up on #59.

- [ ] Remove the time argument `t` from `alternating_update`, since it only makes sense for solvers that implement time evolution like `tdvp`.
- [ ] Improve keyword argument code patterns based on: https://itensor.github.io/ITensors.jl/dev/DeveloperGuide.html#Keyword-Argument-Best-Practices
- [ ] Generalize `ProjTTN` types to a more general `ITensorNetworkCache` type (or specifying custom caching types which are relevant for different contraction backends), based on a contraction sequence tree.
- [ ] Replace `ProjTTNSum` with a more general lazy sum, based on the `Applied` type in `ITensors.LazyApply`.
- [ ] Generalize to accept `AbstractITensorNetwork`, and ensure the interface requirements are general enough to work for more general networks.
- [ ] Think more carefully about the interface and implementation for solving various linear equations, like `Ax ≈ λx`, `Ax ≈ b`, `x ≈ y`, etc. which shares caching structures, works with general tensor networks, contraction backends, etc.

emstoudenmire:
- [ ] Switch default tdvp solver to `exponentiate`

mtfishman:
Thanks, I added it.

emstoudenmire:
Proposed redesign of internals:
  1. Eliminate outermost for loop over substep (see the function update_step in update_step.jl)
  2. Pass a data structure which says how to do a single (full) sweep over the system. This defines what a sweep means, as in nsweeps=3 does three of these.
  3. This data structure can carry an optional "payload" (named tuple?). Using tdvp as an example, the payload would store the explicit, signed and weighted time steps, already preprocessed. In the case of a PEPS it could store the row and column number.
  4. There is an optional, second type of callback function (not a solver) which can post process the payload before feeding it into the solver. (May not be necessary to add this #4 callback feature if the payloads are sufficiently preprocessed.)
  

Basically Lander's data structures become the definition of what a sweep is and are generalized to not just store the forward/reverse sign, but any data relevant to that specific point in the sweep.

PR #67, opened 2023-02-14
Remove t argument from alternating_update and improve keyword argument handling

emstoudenmire:
Improvements Made in this PR
-----------------------------------------
* Removes `t` argument from `alternating_update` in favor of using the `tdvp` interface to process time evolution related inputs. (`alternating_update` is now only given `nsweeps`, basically, plus some other time related keywords, some of which are slated for removal, others of which are just passed down to the ultimate solver).
* This PR makes significant improvements to keyword argument handling. 
* The `_compute_nsweeps` function has been moved into `tdvp.jl` and is only used by the `tdvp` interface
* The solver no longer takes a `t` argument. The only solver which uses time concepts now is the `tdvp` solver, which gets passed a `time_step` keyword arg (this value includes all of the information about sub time steps, the forward or reverse sign, etc.)

Planned Improvements in Future PRs (Can discuss moving them into this PR if it's a good idea)
------------------------------------------
* No longer handling the `current_time` computation and printing directly in the `alternating_update`, `update_step` etc. code directly (would let us remove most of the remaining time related keyword args, like `time_start`, or needing to pass `current_time` up and down the function stack)
* Finishing factoring out the logic around the `TDVPOrder` type so that it only goes in `tdvp.jl` and is no longer involved in algorithms like `dmrg`, `linsolve`, etc.
* Figuring out if we want to factor out the forward/reverse step logic into tdvp.jl also, or whether it's ok for some of that to persist inside `alternating_update`
* Further improvements to keyword arg handling
* Finish simplying solvers to only be a single function, rather than a function that returns another function (unless neccessary)


src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -27,39 +24,41 @@ function _extend_sweeps_param(param, nsweeps)
...
+  solver,
+  PH,
+  psi0::AbstractTTN;
+  checkdone=nothing,
+  order=2,
+  outputlevel=0,
+  time::Number=Inf,
+  time_start=0.0,
+  time_step=time,
+  nsweeps=nothing,
```
mtfishman, line 52: Since `time`, `time_step`, `time_start`, etc. are specific to the `tdvp` backend, maybe we should only define them in that function? We could just default to `nsweeps=1`, or not set a default for `nsweeps` and expect users/particular solvers to set it.

mtfishman: That would move the `_compute_nsweeps` logic into `tdvp`, which to me makes sense since it only seems to be needed by solver backends that actually perform some sort of time evolution.

mtfishman: I guess that leaves a question about how to properly track the current time of the evolution, but maybe that can be handled by an `Observer` or by the `solver` that gets passed to `alternating_update` by the `tdvp` wrapper.

emstoudenmire: This could be a nice simplication of the code! I'll look into it & we definitely want to do some version of this redesign.

emstoudenmire:
What do you think of the keyword name? `time` versus other ones above.

mtfishman:
Also, what do you think about using the keyword argument `nsteps` instead of `nsweeps` in the `tdvp` interface? Then `nsteps`, `t`, and `time_step` could get processed into `nsweeps` which gets input into `alternating_update`.

emstoudenmire:
I think making the interface take `nsteps` is probably the right way to go. I'll look into it more!

It will be interesting how much we can "orthogonalize" the alternating_update code away from TDVP-specific concepts. Right now, there is still a lot of the plumbing that e.g. relies on a `TDVPOrder` object or the concept of sub time steps. But I think I can further refactor the code to have things like perhaps two versions of `update_step` (which is a short function anyway), one which deals with `TDVPOrder` concepts and another which just subs in defaults for more normal dmrg-like algorithms. Either that, or we could elevate the `TDVPOrder` object to a more general pattern, like `UpdatePattern` or some other concept.

Luckily I don't think `local_update` or `_insert_tensor` depend on TDVP concept in as much of an essential way.

mtfishman:
Yes, I think we should push the `TDVPOrder` to the top level `tdvp` wrapper, and then process that in a more general functionality for determining the `nweeps` and time steps for a given `t`, `TDVPOrder`, `time_step`, and `nsteps`.

emstoudenmire:
The time-dependent Hamiltonian tests aren't passing. I'll work on it –

emstoudenmire:
Looks like tests are passing now.

This PR overall is just a stepping stone and some initial cleanup to get the code to where a refactor is easier to do, and to check some boxes on the associated issue #60. Definitely a lot more work to do on this code! I think the rest of the #60 boxes will be easier to check when I put in the design we discussed about passing an array of sweeping steps and "payload" data for the solver to consume.

emstoudenmire:
I'm having trouble getting the tests to pass, and it seems to unfortunately be something not obviously related to this PR. I see an error about part of a graph related code trying to access an empty tuple at [1] but  I'm not otherwise sure what is really behind this.

I thought maybe it was a bug introduced by another recent PR, possibly, but when I run some of the ITensorNetworks tests on my own machine on the main branch they pass.

This PR is ready to be merged in from my end, if you agree, and of course once tests pass. Meanwhile I am working on some of the other improvements to this code on a new branch starting from this one, to be sent in another PR.

src/treetensornetworks/solvers/alternating_update.jl
mtfishman: I think at this point we have a good plan for this but it is more involved so should be handled in future PRs.

mtfishman: Tracking plans for this in #60.

emstoudenmire: Agreed, yes these definitely need to go but I think the way to do it is the future refactor of the code (next PR after this one) to pass the data structure which carries a named tuple of time step information, and which creates the extra observer function for printing the current time.

mtfishman:
Looks like tests are passing, I'll merge. Thanks, nice to see some of this code getting cleaned up!

mtfishman:
I think the failures are related to https://github.com/mtfishman/DataGraphs.jl/pull/17, I'll look into it.

emstoudenmire:
Thanks! The code is already simplifying a lot just from these changes and I think it can be simplified even more.

PR #68, opened 2023-02-17
Let randomITensorNetwork support general distributions

LinjianMa:
The motivation is that I would like to test on random tensor networks with uniform distribution and variable lower and upper bound, and I can imagine one may also want to test random tensor networks with other distributions.

New API:
```julia
function randomITensorNetwork(
  s::IndsNetwork, distribution::Distribution; link_space=nothing
)
```
The distribution can be `Normal()`, `Uniform(-1.0, 1.0)` or others.

One caveat is that `Distributions` seem not support general `eltype`. So the previous implementation that support normal distribution with general `eltype` is unchanged.


mtfishman:
Nice, thanks! Too bad `Distributions` doesn't have better support for more general element types.

I would vote for putting the `Distributions` input as the first argument, since the general convention in ITensor is that information specifying the data goes first in the argument list.

LinjianMa:
> Nice, thanks! Too bad `Distributions` doesn't have better support for more general element types.
> 
> I would vote for putting the `Distributions` input as the first argument, since the general convention in ITensor is that information specifying the data goes first in the argument list.

SG, done!

PR #69, opened 2023-02-19
Introduce `binary_tree_structure`, `path_graph_structure`, and rename `binary_tree_partition`

LinjianMa:
Introduce APIs below as is discussed in https://github.com/mtfishman/ITensorNetworks.jl/pull/66:
```julia
"""
Outputs a maximimally unbalanced directed binary tree DataGraph defining the desired graph structure
"""
function path_graph_structure(tn::ITensorNetwork)
end

"""
Outputs a directed binary tree DataGraph defining the desired graph structure
"""
function binary_tree_structure(tn::ITensorNetwork)
end

function partition(
  ::Algorithm"mincut_recursive_bisection", tn::ITensorNetwork, inds_btree::DataGraph
)

function partition(tn::ITensorNetwork, inds_btree::DataGraph; alg::String)
  return partition(Algorithm(alg), tn, inds_btree)
end
```

### DIFF
--- a/src/mincut.jl
+++ b/src/mincut.jl
@@ -2,6 +2,36 @@
 MAX_WEIGHT = 1e32
 
 """
+Outputs a maximimally unbalanced directed binary tree DataGraph defining the desired graph structure
+"""
+function path_graph_structure(tn::ITensorNetwork)
+  return path_graph_structure(tn, noncommoninds(Vector{ITensor}(tn)...))
+end
+
+"""
+Given a `tn` and `outinds` (a subset of noncommoninds of `tn`), outputs a maximimally unbalanced
+directed binary tree DataGraph of `outinds` defining the desired graph structure
+"""
+function path_graph_structure(tn::ITensorNetwork, outinds::Vector{<:Index})
+  return _binary_tree_structure(tn, outinds; maximally_unbalanced=true)
+end
+
+"""
+Outputs a directed binary tree DataGraph defining the desired graph structure
+"""
+function binary_tree_structure(tn::ITensorNetwork)
+  return binary_tree_structure(tn, noncommoninds(Vector{ITensor}(tn)...))
+end
+
+"""
+Given a `tn` and `outinds` (a subset of noncommoninds of `tn`), outputs a
+directed binary tree DataGraph of `outinds` defining the desired graph structure
+"""
+function binary_tree_structure(tn::ITensorNetwork, outinds::Vector{<:Index})
+  return _binary_tree_structure(tn, outinds; maximally_unbalanced=false)
+end
+
+"""
 Calculate the mincut between two subsets of the uncontracted inds
 (source_inds and terminal_inds) of the input tn.
 Mincut of two inds list is defined as the mincut of two newly added vertices,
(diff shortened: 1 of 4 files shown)


src/mincut.jl
```diff
@@ -126,6 +151,30 @@ function __binary_tree_partition_inds(
   end
 end
 
+function _nested_vector_to_directed_tree(inds_tree_vector::Vector)
```
mtfishman, line 154: It seems like at least conceptually there is overlap with the functionality added in #53, is there anything that could be shared or are the data structures/types involved too different?

mtfishman:
Over all this looks good, glad to see that changing the data structure didn't change much and I like the `path_graph_structure` and `binary_tree_structure` interface.

src/mincut.jl
LinjianMa: Thanks for letting me know! Yes I think the functionality should be very similar for both. I do have some questions for the PR you referred: 
1. If I understand correctly, #53 changes a contraction tree into a `NamedGraph`. I think it's actually more of a `NamedDiGraph`, as is defined in this PR?
2. In #53, each non-leaf vertex in the contraction tree `NamedGraph` includes the data of tensors: https://github.com/mtfishman/ITensorNetworks.jl/blob/main/src/contraction_tree_to_graph.jl#L43 I'm not sure if that is necessary. In my implementation, the graph is a `DataGraph(NamedDiGraph(), Index)` or `DataGraph(NamedDiGraph(), ITensor)`, and there will be data only at the leaves of the graph. For intermediate vertices, we can infer its data by simply gathering the data of its leaves. And since the data graph is defined on a directed graph, these properties can be retrieved pretty easily. 

Please let me know what you think: if what I mentioned above makes sense, I think there is room to improve the implementation of the current `contraction_path_to_graph`

mtfishman: 1. The reason for using an undirected graph in that case is because we wanted to keep it agnostic about the root node, for example since if you want to take gradients you might want to change the root node like in https://arxiv.org/abs/1310.8023. But maybe there should be two functions, one that preserves the root and outputs a directed graph and another that outputs the undirected version.
2. It doesn't include any actual "data", mostly that is just the vertex naming convention. The naming convention is that each internal vertex is named by 3 sets of vertices (for binary trees of degree 3, for example), where each set contains the leaf vertices of the subtree of each incident edge of the vertex. The reasoning is that it is an informative and unique name for the internal vertices (as opposed to just picking arbitrary or random names). If you have suggestions for a better naming convention I'd be open to suggestions.

LinjianMa: OK, I think I can see the difference now, thanks Matt. So to summarize the differences:
1. In my case the input is a nested vector of indices, while in the `contraction_sequence_to_graph` the input is just a sequence without actual data. We can transfer one to the other of course, but that also involves some interfaces.
2. For my case, I want a directed graph with a fixed root, rather than an undirected graph. As you have mentioned, keeping the graph undirected have some use cases for the contraction sequence.

I think the naming convention for the contraction sequence tree makes sense, previously I misunderstand it as the real data. 

So maybe it's good to keep both implementations for now? The use cases are similar but they do have multiple differences

mtfishman: Yes, let's just keep them separate for now since it's not "obvious" how to combine them. Partially I just wanted to make you aware of that code and functionality. We could more abstractly think about functionality for converting from AbstractTrees to Graphs (and back), which I think would be useful, but will requiring various applications to think about the right abstraction for that, how to deal with metadata, different graph types, etc.

PR #71, opened 2023-02-22
Add interface to `SweepContractor` in examples

LinjianMa:
This will be useful later when we compare our algorithms with SweepContractor


mtfishman:
Nice, thanks!

PR #72, opened 2023-02-23
Symmetric gauge

JoeyT1994:
This PR introduces the function `symmetrise_itensornetwork(ψ::ITensorNetwork)` which will gauge an ITensorNetwork `ψ` and make it symmetric. As a result the contract of any local tensor `ψ[v]` with its adjoint whilst keeping one virtual index uncontracted will be a diagonal matrix. 

A test is introduced for this PR:
i) It checks `symmetrise_itensornetwork(ψ)` is just a gauge transform and has an overlap of unity with `ψ`
ii) It checks that when one does simple belief propagation on `symmetrise_itensornetwork(ψ)` the resulting message tensors are diagonal (a property guaranteed by the symmetrise function).

@mtfishman : I haven't implemented a check for setting any small negative diagonal elements of `D` to 0. This is because we also have to take the inverse of `D` and so such a reset would create infinities in `inv_D`. Shall I just throw an error/ warning if any diagonal elements of `D` are below a small, positive tolerance?


src/gauging.jl
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  return out
+end
+
+"""Put an ITensorNetwork into the symmetric gauge and also return the message tensors (which are the diagonal bond matrices)"""
+function symmetrise_itensornetwork(ψ::ITensorNetwork)
+  ψψ = ψ ⊗ prime(dag(ψ); sites=[])
+  vertex_groups = nested_graph_leaf_vertices(partition(ψψ, group(v -> v[1], vertices(ψψ))))
+  mts = compute_message_tensors(ψψ; vertex_groups=vertex_groups)
```
mtfishman, line 16: You can just use `vertex_groups` as a shorthand for `vertex_groups=vertex_groups`.

src/gauging.jl
```diff
@@ -0,0 +1,58 @@
...
+  is = [[j for i in 1:length(inds(A))] for j in 1:mindim(out)]
+  for i in is
+    out[i...] = f(out[i...])
+  end
+
+  return out
+end
+
+"""Put an ITensorNetwork into the symmetric gauge and also return the message tensors (which are the diagonal bond matrices)"""
+function symmetrise_itensornetwork(ψ::ITensorNetwork)
```
mtfishman, line 13: What about just `symmetric_gauge(ψ::ITensorNetwork)`? `symmetrise` is quite a general term (i.e. it could mean forcing the tensors to be symmetric about permutations of the indices, which is not what this is doing in general, though for a network that is actually rotationally invariant maybe it helps).

src/gauging.jl
```diff
@@ -0,0 +1,58 @@
...
+    Ce = rootX * replaceinds(rootY, edge_ind, edge_ind_sim)
+
+    U, S, V = svd(Ce, edge_ind)
+    rootS = modify_diagonal_els_itensor(S, x -> x^0.5)
+
+    ψsymm[vsrc] = replaceinds(ψsymm[vsrc] * U * rootS, commoninds(S, V), edge_ind)
+    ψsymm[vdst] = replaceinds(ψsymm[vdst], edge_ind, edge_ind_sim)
+    ψsymm[vdst] = replaceinds(ψsymm[vdst] * dag(rootS * V), commoninds(U, S), edge_ind)
+
+    replaceinds!(S, inds(S), inds(mts[s1 => s2]))
```
mtfishman, line 52: Let's use:
```julia
S = replaceinds(S, inds(S) => inds(mts[s1 => s2]))
```
(I wish we didn't have the in-place version, there is essentially no performance advantage and it can lead to confusing bugs if you accidentally modify the indices of a reference to the same ITensor).

Though this kind of code is a bit suspicious to me, I we certain we always replace the indices in the correct way (i.e. with the correct ordering?). I guess because `S` is diagonal it is relatively safe but the code above using `replaceinds` definitely raises alarm bells for me.

src/gauging.jl
```diff
@@ -0,0 +1,58 @@
...
+
+  for e in edges(ψ)
+    vsrc, vdst = src(e), dst(e)
+
+    s1, s2 = find_subgraph((vsrc, 1), mts), find_subgraph((vdst, 1), mts)
+    edge_ind = commoninds(mts[s1 => s2], ψsymm[vsrc])
+    edge_ind_sim = sim(edge_ind)
+
+    X_D, X_U = eigen(mts[s1 => s2]; ishermitian=true)
+    Y_D, Y_U = eigen(mts[s2 => s1]; ishermitian=true)
```
mtfishman, line 29: Maybe to handle some of the potential zeros, we could add a cutoff here. For example, we could default to `cutoff=10eps(real(scalartype(ψ)))` and control it with a kwarg in the top level function, for example:
```julia
function symmetric_gauge(
  ψ::ITensorNetwork;
  eigen_message_tensor_cutoff=10eps(real(scalartype(ψ))),
)
  # ...
  X_D, X_U = eigen(mts[s1 => s2]; ishermitian=true, cutoff=eigen_message_tensor_cutoff)
  # ...
end
```

mtfishman:
`modify_diagonal_els_itensor` is a nice idea. I would use the name `map_diag` (in analogy to Julia's `Base.map`).

Here's an implementation using some internal/hidden functions from NDTensors:
```julia
using ITensors
using ITensors.NDTensors
using ITensors.NDTensors: getdiagindex, setdiagindex!

function map_diag!(f::Function, it_destination::ITensor, it_source::ITensor)
  return itensor(map_diag!(f, tensor(it_destination), tensor(it_source)))
end
map_diag(f::Function, it::ITensor) = map_diag!(f, copy(it), it)

function map_diag!(f::Function, t_destination::Tensor, t_source::Tensor)
  for i in 1:diaglength(t_desination)
    setdiagindex!(t_destination, getdiagindex(t_source, i), i)
  end
  return t_destination
end
map_diag(f::Function, t::Tensor) = map_diag!(f, copy(t), t)

# Convenience functions
sqrt_diag(it::ITensor) = map_diag(sqrt, it)
inv_diag(it::ITensor) = map_diag(inv, it)
invsqrt_diag(it::ITensor) = map_diag(inv ∘ sqrt, it)
pinv_diag(it::ITensor) = map_diag(pinv, it)
pinvsqrt_diag(it::ITensor) = map_diag(pinv ∘ sqrt, it)
```

Passing a cutoff to `eigen` like I comment above or using a pseudoinverse (`LinearAlgebra.pinv`) would be approaches to handle zero or small eigenvalues. However, to deal with negative eigenvalues, I think the most general approach is adding a small positive shift (a regularization), for example `X_D = map_diag(x -> x + regularization, X_D)` where we can have `regularization` as a keyword argument passed to `symmetric_gauge`. Maybe we could default to `regularization=10eps(real(scalartype(ψ))`, or automatically set it to something slightly larger than the most negative eigenvalue of `X_D` (though that could be uncontrolled if it is not close enough to being positive semidefinite, for example if the message tensors aren't properly converged).

src/gauging.jl
JoeyT1994: Yeah that sounds like a good name. Happy with that.

src/gauging.jl
JoeyT1994: Yes because `S` is diagonal I didn't think much about the ordering of the indices but maybe that is a bit sloppy. It is not clear to me which index of `S` should be the `edge_ind` and which should be `prime(edge_ind)` though? Perhaps we could just be a bit more precise and use `replaceinds(S, [commonind(U, S), commonind(V,S)] => [edge_ind, prime(edge_ind)]`?

src/gauging.jl
JoeyT1994: Okay yeah this sounds like a good idea. Will do that.

JoeyT1994:
Thanks Matt! Let me implement those changes. So we will have one keyword argument for the cutoff of `eigen` and one for regularising the values of `X_D` and avoiding anything small and negative. 

I guess ideally those functions for mapping the diagonal of an ITensor should go in `ITensors.jl` but perhaps there is not enough of a use case for it to make a PR for them?

src/gauging.jl
mtfishman: Yes, that would be a good solution. At least then it is explicit when someone reads the code, even if there are no practical issues because it is diagonal.

mtfishman:
I think those are good candidates for `ITensors.jl`/`NDTensors.jl` but we can just keep them in `ITensorNetworks.jl` for now (maybe in new folders `src/NDTensorsExt` and `src/ITensorsExt`) and test them out a bit more here.

test/test_symmetriseitensornetwork.jl
```diff
@@ -0,0 +1,36 @@
...
+
+  ψψ_symm = ψ_symm ⊗ prime(dag(ψ_symm); sites=[])
+  vertex_groups = nested_graph_leaf_vertices(
+    partition(ψψ_symm, group(v -> v[1], vertices(ψψ_symm)))
+  )
+  ψ_symm_mts_V2 = compute_message_tensors(ψψ_symm; vertex_groups=vertex_groups)
+
+  for e in edges(ψ_symm_mts_V2)
+    #Test all message tensors are approximately diagonal (note tolerance is a bit dependent on how many iters used to compute mts)
+    @test isapprox(norm(diag(ψ_symm_mts_V2[e])), norm(ψ_symm_mts_V2[e]); atol=1e-4)
```
mtfishman, line 34: We can use:
```julia
m_e = ψ_symm_mts_V2[e]
@test diagITensor(vector(diag(m_e)), inds(m_e)) ≈ m_e atol=1e-4
```
to compare elementwise, which is better than just comparing the norms.

JoeyT1994:
Added a Keyword argument to specify number of iterations for calculating message tensors in `symmetric_gauge()`. Increasing this this helps improve the convergence to diagonal message tensors a lot so we can increase to `atol=1e-8` in the check.

mtfishman:
Looks good, thanks!

PR #73, opened 2023-02-28
Introduce `_contract_deltas`

LinjianMa:
```julia
"""
Given an input tensor network `tn`, remove redundent delta tensors
in `tn` and change inds accordingly to make the output `tn` represent
the same tensor network but with less delta tensors.
"""
function _contract_deltas(tn::ITensorNetwork)
  ...
end

"""
Given an input `partition`, contract redundent delta tensors in `partition`
without changing the tensor network value. `root` is the root of the
dfs_tree that defines the leaves.
Note: for each vertex `v` of `partition`, the number of non-delta tensors
  in `partition[v]` will not be changed.
Note: only delta tensors of non-leaf vertices will be contracted.
"""
function _contract_deltas(partition::DataGraph; root=1)
 ...
end
```


src/Graphs/abstractgraph.jl
```diff
@@ -18,11 +18,6 @@ function internal_edges(g::AbstractGraph)
   return filter(e -> !is_leaf_edge(g, e), edges(g))
 end
 
-"""Get all vertices which are leaves of a graph"""
```
LinjianMa, line 21: This function is already in `NamedGraphs` so we should remove it.

src/contract_deltas.jl
```diff
@@ -0,0 +1,160 @@
...
+
+"""
+Given an input `partition`, contract redundent delta tensors in `partition`
+without changing the tensor network value. `root` is the root of the
+dfs_tree that defines the leaves.
+Note: for each vertex `v` of `partition`, the number of non-delta tensors
+  in `partition[v]` will not be changed.
+Note: only delta tensors of non-leaf vertices will be contracted.
+"""
+function _contract_deltas(partition::DataGraph; root=1)
```
mtfishman, line 127: So the idea here is that it does the same thing as `_contract_deltas(tn::ITensorNetwork)` but only within the partitions, i.e. it doesn't remove deltas that cross partitions?

If so, couldn't that be implemented by looping through the subgraphs and calling `_contract_deltas(tn_subgraph::ITensorNetwork)`?

Also, should this be `root=first(vertices(partition))` to handle more general vertex partition names?

LinjianMa: so the idea is that it will also contract delta cross partitions; except the partition defined on leaf vertices. The constraint on leaf vertices is to make sure that the noncommininds of the partition will still be adjacent to the leaves after removing deltas (otherwise the leaf partition could be empty and the noncommoninds then will be adjacent to inner partitions, which is not a good case to deal with, especially for my density matrix algorithm implementation)

And great idea for the root interface, will do that

LinjianMa: therefore, calling `_contract_deltas(partition[v])` for each `v` doesn't work since since it will not contract delta cross partitions

mtfishman: I see, thanks. That seems a bit specialized for using the general function name `_contract_deltas`. What if we named this version `_contract_deltas_ignore_leaf_partitions` to make that clearer?

Also, it seems better to implement this by calling out to `_contract_deltas(tn::ITensorNetwork)`, for example by making an ITensorNetwork built by unioning the tensor networks of the partitions living on the nonleaf vertices `nonleaf_vertices`, calling `_contract_deltas(tn::ITensorNetwork)`, and then partitioning the resulting network.

LinjianMa: that sounds good, let me think about how to implement it

LinjianMa: Done, please let me know if you have any questions

mtfishman: Thanks for the quick change, looks reasonable!

PR #76, opened 2023-03-21
Square root ising

JoeyT1994:
Added Functions to `specialitensornetworks.jl` for creating the Tensor Network State corresponding to the square root of the classical ising partition function.

Also:

i) Removed a repeated docstring.
ii) Specified the link_space argument in `ising_network` so it works for both an index network with siteinds and without.

### DIFF
--- a/src/specialitensornetworks.jl
+++ b/src/specialitensornetworks.jl
@@ -76,21 +76,21 @@ end
 
 """Build the wavefunction whose norm is equal to Z of the classical ising model
 s needs to have site indices in this case!"""
-function square_root_ising_network(
+function ising_network_state(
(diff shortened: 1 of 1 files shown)


mtfishman:
@JoeyT1994 looks like this broke a BP test.

JoeyT1994:
Okay I see. I fixed it now with the latest commit.

mtfishman:
Maybe `ising_network_state` would be a better name? `square_root` doesn't seem quite right for this (it's more that it is a sate such that the norm of the state gives the Ising model).

mtfishman:
Thanks Joey, this will be very nice to have!

PR #77, opened 2023-03-28
Refactor alternating_update to use sweep_regions

emstoudenmire:
# Major Changes

- `update_sweep` now accepts a keyword argument `sweep_regions` which is an array of
  "Frankentuples" i.e. a tuple of the form `(region, (; step_kwargs...))` that
  tells it how to move through the tree and what arguments to pass to the solver 
  and observer at each step

- Made generic `half_sweep` function which handles tree iteration logic and 
  uses a `region_function` callback to actually compute the regions that go
  into `sweep_regions`. (I originally tried to compute the TDVP "in between" 
  regions from an array that didn't have them but this caused numerous problems.)

- Simplified what the TDVPOrder type does. It is now basically a custom Val.

As a result of the above changes, the `alternating_update` and `update_step` 
codes now basically have any TDVP-specific logic factored out of them, with
the one exception of handling printing in a more custom way. The custom 
printing idea, based on appending printing functions to the set of 
Observer functions, will be in a future PR.


src/treetensornetworks/solvers/tree_sweeping.jl
```diff
@@ -1,76 +1,49 @@
...
-  return steps
-end
-
-function one_site_sweep(direction::Base.ReverseOrdering, args...; kwargs...)
-  return reverse(reverse.(one_site_sweep(Base.Forward, args...; kwargs...)))
-end
+put_kwargs(v::Vector) = (v, (;))
+put_kwargs(j::Integer) = ([j], (;))
+put_kwargs(t::Tuple{<:Integer}) = (t, (;))
+put_kwargs(t::Tuple{<:Integer,<:Integer}) = (t, (;))
```
mtfishman, line 20: Here's a more elegant and general approach:
```julia
function append_missing_namedtuple(t::Tuple)
  return reverse(prepend_missing_namedtuple(reverse(t)))
end
prepend_missing_namedtuple(t::Tuple) = ((;), t...)
prepend_missing_namedtuple(t::Tuple{<:NamedTuple,Vararg}) = t
```
to circumvent the fact that you can only use a `Vararg` in the last position of an argument list/Tuple.

This gives:
```julia
julia> append_missing_namedtuple((1, 2, 3))
(1, 2, 3, NamedTuple())

julia> append_missing_namedtuple((1, 2, 3, (; x=2)))
(1, 2, 3, (x = 2,))
```

I think this should only accept `Tuple`s of the form `(arg1, arg2, ...)` or `(arg1, arg2, ..., (; kwarg1, kwarg2, ...))`, in the sense of a [FrankenTuple](https://ararslan.github.io/FrankenTuples.jl/latest/), to keep it simpler.

If we need to convert a non-tuple to a tuple, we could have a different pre-processing function `to_tuple` that converts to a tuple if it isn't already:
```julia
to_tuple(x) = (x,)
to_tuple(x::Tuple) = x
```
so then we would use:
```julia
steps = append_missing_namedtuple.(to_tuple.(steps))
```
to canonicalize the steps.

src/treetensornetworks/solvers/tree_sweeping.jl
```diff
@@ -1,76 +1,71 @@
-# 
-# Sweep step
-# 
+direction(step_number) = isodd(step_number) ? Base.Forward : Base.Reverse
```
emstoudenmire, line 2: `make_region` below is now a bit complicated, but it is the unification of the four previous functions one_site_region, two_site_region, tdvp_one_site_region, tdvp_two_site_region.

emstoudenmire:
The changes in this PR are so large it may be better to do the review in person, when you have time. I could give you a tour of the new code. It's very different from the previous code so the diffs don't really help, though maybe you're better at reading them than I am.

test/test_treetensornetworks/test_solvers/test_dmrg.jl
```diff
@@ -41,6 +37,38 @@ using Test
     H, psi; nsweeps, maxdim, cutoff, nsite, solver_krylovdim=3, solver_maxiter=1
   )
   @test inner(psi', H, psi) ≈ inner(psi2', H_mpo, psi2)
+
```
emstoudenmire, line 40: Note the new ability below to pass in a custom sweeping pattern from outside the algorithm, represented as a simple array of arrays (or array of tuples is ok).

mtfishman:
@emstoudenmire I think this looks good and I'm happy to merge this and then build on top of it in future PRs.

Let's start some issues with reminders of things we plan on improving, I've already forgotten from our last conversation. I've started with #90.

emstoudenmire:
Good idea to make a tracking issue. Glad you agree we can use this PR as a base. That's definitely how I envision it i.e. the tree_sweeping.jl code definitely could be improved and the design iterated, but this PR does accomplish the goal of totally isolating it into a closed, distinct subsystem. Also there is the observer-based printing idea but that is coming up in a follow-up PR based on this one.

Could you then please merge and I'll work on finalizing the observer printing PR, and might stick a few other improvements into that.

PR #80, opened 2023-04-05
[Easy] let `contract_deltas` handle the case with no deltas

### LINKED ISSUE #79: Approx_itensornetwork() on a rank two tensor yields an error
If I try to call `approx_itensornetwork()` on a tensor network `T` consisting of a single rank two tensor `M` I get an error. 

I.e. the following code:

```
i = Index(2, "i")
M =  randomITensor(i, i')
M_itn = ITensorNetwork([M])

@show approx_itensornetwork(M_itn, binary_tree_structure(M_itn); alg = "density_matrix")
```
yields:

image

But will work if the tensor `M` has more than two indices.

LinjianMa:
This also fixes #79


mtfishman:
Great, thanks!

PR #81, opened 2023-04-06
Let `approx_itensornetwork` support complex tensor network

LinjianMa:


mtfishman:
Thanks, good to see it only requires some small changes.

PR #83, opened 2023-04-06
Let binary tree partition support the disconnected tensor network

LinjianMa:
This PR makes binary tree partition work when the input tn is disconnected. Example:
```julia
inds = [Index(2, "$i") for i in 1:5]
tn = ITensorNetwork([randomITensor(i) for i in inds])
par = partition(tn, binary_tree_structure(tn); alg="mincut_recursive_bisection")
```
For this case, we still want `par` to be connected so that it has a binary tree structure. To achieve that, we add multiple scalar tensors with value 1.0 to maintain connectivity, without changing the output value.

### DIFF
--- a/src/binary_tree_partition.jl
+++ b/src/binary_tree_partition.jl
@@ -5,20 +5,40 @@ function _binary_partition(tn::ITensorNetwork, source_inds::Vector{<:Index})
   tn = map_data(t -> replaceinds(t, external_inds => external_sim_ind), tn; edges=[])
   tn_wo_deltas = rename_vertices(v -> v[1], subgraph(v -> v[2] == 1, tn))
   deltas = Vector{ITensor}(subgraph(v -> v[2] == 2, tn))
+  scalars = rename_vertices(v -> v[1], subgraph(v -> v[2] == 3, tn))
   new_deltas = [
     delta(external_inds[i], external_sim_ind[i]) for i in 1:length(external_inds)
   ]
   deltas = [deltas..., new_deltas...]
-  tn = disjoint_union(tn_wo_deltas, ITensorNetwork(deltas))
+  tn = disjoint_union(tn_wo_deltas, ITensorNetwork(deltas), scalars)
   p1, p2 = _mincut_partition_maxweightoutinds(
     tn, source_inds, setdiff(external_inds, source_inds)
   )
   source_tn = _contract_deltas(subgraph(tn, p1))
   remain_tn = _contract_deltas(subgraph(tn, p2))
+  outinds_source = noncommoninds(Vector{ITensor}(source_tn)...)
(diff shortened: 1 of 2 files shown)


mtfishman:
Seems like a good solution.

I think this strategy of adding trivial link indices to add connectivity to an `ITensorNetwork` is similar to these two functions:
https://github.com/mtfishman/ITensorNetworks.jl/blob/818b6e4493075b164ac7b1ce84cf762fe03c39ba/src/itensornetwork.jl#L245-L254
https://github.com/mtfishman/ITensorNetworks.jl/blob/3df4c849016e9a790727029fc2d8cdda915782fb/src/abstractitensornetwork.jl#L806-L825

Can we use one of those directly here? It looks like `insert_missing_internal_inds` is basically the same as what you're using. If that isn't useful, no worries, the code you're adding is pretty simple and partially I'm pointing it out as a way to let you know about the existence of those functions. Maybe we can think about nice abstractions for this concept of inserting dummy link/internal indices into an `ITensorNetwork`, which I think is quite useful and will arise in multiple places.

LinjianMa:
> Seems like a good solution.
> 
> I think this strategy of adding trivial link indices to add connectivity to an `ITensorNetwork` is similar to these two functions:
> 
> https://github.com/mtfishman/ITensorNetworks.jl/blob/818b6e4493075b164ac7b1ce84cf762fe03c39ba/src/itensornetwork.jl#L245-L254
> 
> 
> https://github.com/mtfishman/ITensorNetworks.jl/blob/3df4c849016e9a790727029fc2d8cdda915782fb/src/abstractitensornetwork.jl#L806-L825
> 
> Can we use one of those directly here? It looks like `insert_missing_internal_inds` is basically the same as what you're using. If that isn't useful, no worries, the code you're adding is pretty simple and partially I'm pointing it out as a way to let you know about the existence of those functions. Maybe we can think about nice abstractions for this concept of inserting dummy link/internal indices into an `ITensorNetwork`, which I think is quite useful and will arise in multiple places.

@mtfishman Ah I see, thanks for pointing out! It's indeed doing the same thing. I thought about using `insert_missing_internal_inds` or part of the function, but after thinking about this in detail I realize the `_binary_partition` function has several special places, so summarize that here for discussion:
1. minor issue: here we want to insert two scalars to two partitions, respectively, so some modification is needed to use `insert_missing_internal_inds` 
2. Major issue: in `insert_missing_internal_inds`, each scalar is contracted with one tensor in the `tn`. This can be a potential issue here: it's possible that all tensors in a partition are delta tensors used to maintain the tree structure, and we will remove these delta tensors later for the efficiency purpose. However, contracting a delta with this scalar makes the output not a scalar tensor, which can be a potential issue when manipulating the partitions later.

mtfishman:
No problem. Let's just keep in mind if there are some small functions for inserting these "virtual link indices" into a network that we can share across these different use cases.

PR #85, opened 2023-04-11
Belief Propagation Version 3

JoeyT1994:
This PR introduces a more generalized version of belief propagation where the message tensors are now ITensorNetworks whose structure can be specified by the user.
Message tensor updates therefore can utilize the `approx_itensornetwork()` function and thus there is a new degree of freedom one can utilise for belief propagation. Default behaviour (not specifying contraction arguments for `update_mt()`) is to have the message tensors take the form of an `ITensorNetwork` containing a single `ITensor`

Corresponding tests and examples have been updated to reflect these changes.

Within `beliefpropagation.jl`, the `calculate_contraction` function has been replaced with `calculate_contraction_network` and it is then up to the user to decide how to contract the resulting network to obtain, e.g., an expectation value.

I have also defined `ITensorNetwork(t::ITensor)` to default to `ITensorNetwork(ITensor[t])`


src/beliefpropagation.jl
```diff
@@ -7,35 +7,60 @@ function construct_initial_mts(
 end
 
 function construct_initial_mts(
-  tn::ITensorNetwork, subgraphs::DataGraph; init=(I...) -> @compat allequal(I) ? 1 : 0
+  tn::ITensorNetwork,
+  subgraphs::DataGraph;
```
mtfishman, line 11: Do we need to pass both the tensor network and the subgraphs, or are the subgraphs sufficient?

mtfishman: Also I would rename this function something like `message_tensors`. For two reasons:
1. Using "initial" implies we are going to be using the message tensors that we construct to initialize a calculation, but we could be using them for another purpose.
2. Spelling out the name fully instead of using the shorthand "mt" for "message tensor".

src/beliefpropagation.jl
```diff
@@ -47,53 +72,56 @@ DO a single update of a message tensor using the current subgraph and the incomi
...
 
 """
 Do an update of all message tensors for a given ITensornetwork and its partition into sub graphs
 """
-function update_all_mts(
-  tn::ITensorNetwork,
-  mts::DataGraph;
-  contraction_sequence::Function=tn -> contraction_sequence(tn; alg="optimal"),
-)
+function update_all_mts(tn::ITensorNetwork, mts::DataGraph; contract_kwargs=(;))
```
mtfishman, line 106: Can we just call this `belief_propagation_step` or `belief_propagation_iteration` instead of `update_all_mts`? I guess this is a single step of BP? And can we call the outer loop just `belief_propagation`?

src/beliefpropagation.jl
```diff
@@ -109,27 +137,45 @@ function get_environment(tn::ITensorNetwork, mts::DataGraph, verts::Vector; dir=
...
+    end
+  end
 end
 
 """
 Calculate the contraction of a tensor network centred on the vertices verts. Using message tensors.
 Defaults to using tn[verts] as the local network but can be overriden
 """
-function calculate_contraction(
+function calculate_contraction_network(
```
mtfishman, line 170: What about `approx_subnetwork` or `approx_network_region`, as in it outputs an approximation of a subnetwork or region of the network?

src/beliefpropagation.jl
```diff
@@ -47,53 +72,56 @@ DO a single update of a message tensor using the current subgraph and the incomi
 function update_mt(
```
mtfishman, line 72: What about `update_message_tensor` or `message_tensor_update`?

src/beliefpropagation.jl
```diff
@@ -109,27 +137,45 @@ function get_environment(tn::ITensorNetwork, mts::DataGraph, verts::Vector; dir=
...
+    itn = reduce(⊗, itns)
+    if output_type == ITensorNetwork
+      return itn
+    elseif output_type == Vector{ITensor}
+      return ITensor[itn[v] for v in vertices(itn)]
+    else
+      error("Output Type for get_environment not Supported!")
+    end
+  end
 end
```
mtfishman, line 164: This is a bit confusing and hard to follow, could you explain what this is doing?

Also, it is preferable for code organization to dispatch on types instead of using if-statements like this, for example:
```julia
function get_environment(
  output_type::Type, tn::ITensorNetwork, mts::DataGraph, verts::Vector; kwargs...
)
  itns = get_environment(tn::ITensorNetwork, mts::DataGraph, verts::Vector; kwargs...)
  return _convert_itensornetworks(output_type, itns)
end

_convert_itensornetworks(::Type{<:Vector{<:ITensorNetwork}}, itns) = itns

_convert_itensornetworks(::Type{<:ITensorNetwork}, itns) = reduce(⊗, itns)

function _convert_itensornetworks(::Type{Vector{ITensor}}, itns)
  itn = _convert_itensornetworks(ITensorNetwork, itns)
  return ITensor[itn[v] for v in vertices(itn)]
end
```
though it doesn't seem like this requires a devoted `get_environment` interface to do all of this (it seems like someone could just do the conversions themselves if they want to, as a post-processing of the output of `get_environment`).

mtfishman:
@JoeyT1994 besides the function renaming I raised in the  comments above, over all it looks good! Nice to see it involves pretty minimal changes. Is there anything in particular you think I should look at?

src/beliefpropagation.jl
JoeyT1994: Good idea on the name. I believe we need the tensor network for getting the correct indices on the message tensors.

JoeyT1994:
No I think it all seems straightforward and there is nothing else in particular for you to look at. Just let me know what you want to do with the output type of `get_environment`.

src/beliefpropagation.jl
mtfishman: Say we define:
```julia
ITensorNetwork(itns::Vector{ITensorNetwork}) = reduce(⊗, itns)
```
which seems reasonable to have anyway.

In that case, it doesn't seem any harder to write:
```julia
envs = ITensorNetwork(get_environment(tn, mts, verts))
```
vs.
```julia
envs = get_environment(ITensorNetwork, tn, mts, verts)
```
so it's not clear to me that conversion functionality should be in the `get_environment` function.

For now let's put aside the issue of outputting as `Vector{ITensor}`, which is already easy to convert to from an `ITensorNetwork` with:
```julia
tn = ITensorNetwork(named_grid((2, 2)); link_space=2)
ts = Vector{ITensor}(tn)
```
and I don't want to encourage people to use `Vector{ITensor}` in general anyway.

Part of my reason for trying to keep `get_environment` simpler is anticipating how that function will work for other contraction backends beyond BP, like Linjian's algorithm or CTMRG. The more features we add to it, the harder it is to design it to support various backends going forward.

Some other related questions:

1. When would you actually want `Vector{ITensorNetwork}` output instead of just an `ITensorNetwork`?
2. If there is a case where `Vector{ITensorNetwork}` is useful, is it better to output it as a partitioned graph in the standard format we use for that, i.e. as a `DataGraph`? I guess for BP they would be disconnected graphs anyway so it doesn't matter much, but that also points to the fact that in BP the format `Vector{ITensorNetwork}` doesn't seem so useful anyway, since we could just output `ITensorNetwork` and users could partition it into disconnected components if they want to distinguish between message tensor networks.

So what I'm getting it is ultimately if we can simplify all of this just by outputting an `ITensorNetwork`.

src/beliefpropagation.jl
mtfishman: I understand that in practice it may be easier given the tensor network, but in principle isn't all that information in the subgraphs, i.e. you can just look at the common indices between subgraphs? I think in fact the `partition` function may even output that information as the edge metadata, if I remember correctly.

src/beliefpropagation.jl
mtfishman: Right, and also I think that is a case where `full_update` could just accept both as inputs and do the conversion itself.

JoeyT1994: Okay let me change that in `full_update`

src/beliefpropagation.jl
JoeyT1994: Okay let me took a look into that and see if it does have that info and I can use it

JoeyT1994: Yeah you are right it's all in there so I have removed `tn` as an argument.

test/test_belief_propagation.jl
```diff
@@ -95,4 +106,44 @@ using SplitApplyCombine
...
+  Oψ[v] = apply(op("Sz", s[v]), ψ[v])
+  ψOψ = flatten_networks(ψ, dag(Oψ); combine_linkinds=false, map_bra_linkinds=prime)
+
+  combiners = linkinds_combiners(ψψ)
+  ψψ = combine_linkinds(ψψ, combiners)
+  ψOψ = combine_linkinds(ψOψ, combiners)
+
+  nsites = 1
+  vertex_groups = nested_graph_leaf_vertices(partition(ψψ, group(v -> v[1], vertices(ψψ))))
+  mts = compute_message_tensors(
```
mtfishman, line 130: Can we rename `compute_message_tensors(...)` to `belief_propagation(...)`? It looks like it is the same function except it picks some default for the initial message tensors, in general when that is the case (i.e. two functions are the same but one sets some default inputs) I would name them the same thing.

JoeyT1994: Good idea.

examples/belief_propagation/bpexample.jl
```diff
@@ -46,23 +47,62 @@ function main()
...
+      output_structure=path_graph_structure,
+      maxdim,
+      contraction_sequence_alg="optimal",
+    ),
+    init_contract_kwargs=(;
+      alg="density_matrix",
+      output_structure=path_graph_structure,
+      cutoff=1e-16,
+      contraction_sequence_alg="optimal",
+    ),
```
mtfishman, line 91: Why are there two sets of keyword arguments here?

JoeyT1994: One is for the construction of the initial message tensors (where you generally want to contract the rank 1 boundary network down to the desired `output_structure` which specifies the initial structure of the message tensors) and the other is for the message tensor update step where you are going to want to think harder about contracting the resulting network down to the desired structure.

JoeyT1994: Wasn't entirely sure how to handle this part. It comes down to how one does the initialization of the message tensors.

JoeyT1994: Okay yeah I can see the logic behind that. Let me remove that function and have a go at writing some wrappers.

Do you think the default should be MPS or just a single tensor for the message tensors? I would lean towards a single tensor just as that is the original formulation of BP for a tensornetwork.

mtfishman: I would vote for defaulting to an MPS since that scales to large partition sizes. Also it has the nice limit that for a rank-1 MPS, the message tensors of generalized BP reduce to a product of message tensors, which is the same as simple BP.

src/beliefpropagation.jl
```diff
@@ -1,41 +1,67 @@
...
+          inds(subgraphs[subgraph][v]), vcat(inds_boundary, inds_internal)
+        )
+        push!(
+          boundary_tensors,
+          itensor(
+            [init(Tuple(I)...) for I in CartesianIndices(tuple(dim.(new_inds)...))],
+            new_inds,
+          ),
+        )
+      end
```
mtfishman, line 69: So there is a bit of a "secret" feature in `partition` where it stores the boundary edges of the original partitioned network as the edge metadata of the `DataGraph` representing the partitioned network:
```julia
julia> using Metis

julia> using ITensorNetworks

julia> tn = ITensorNetwork(named_grid((4, 4)); link_space=2);

julia> tnp = partition(tn; nvertices_per_partition=4);

julia> vertices(tnp)
4-element Vector{Int64}:
 1
 2
 3
 4

julia> vertices(tnp[1])
4-element Vector{Tuple{Int64, Int64}}:
 (1, 1)
 (2, 1)
 (1, 2)
 (2, 2)

julia> vertices(tnp[2])
4-element Vector{Tuple{Int64, Int64}}:
 (3, 1)
 (4, 1)
 (3, 2)
 (4, 2)

julia> tnp[1 => 2][:edges]
2-element Vector{NamedGraphs.NamedEdge{Tuple{Int64, Int64}}}:
 (2, 1) => (3, 1)
 (2, 2) => (3, 2)
```
Is that helpful here?

You can see the logic here for extracting those edges here: https://github.com/mtfishman/ITensorNetworks.jl/blob/741a28f8c3c31fd0161816ae1c7fcc379e001cea/src/partition.jl#L235-L257

examples/belief_propagation/bpexample.jl
```diff
@@ -28,41 +28,77 @@ function main()
...
-    partition(partition(ψψ, group(v -> v[1], vertices(ψψ))); nvertices_per_partition=nsites)
+  Z = partition(
+    partition(ψψ, group(v -> v[1], vertices(ψψ))); nvertices_per_partition=nsites
   )
-  mts = compute_message_tensors(ψψ; vertex_groups=vertex_groups)
-  sz_bp =
-    calculate_contraction(
-      ψψ, mts, [(v, 1)]; verts_tensors=ITensor[apply(op("Sz", s[v]), ψ[v])]
-    )[] / calculate_contraction(ψψ, mts, [(v, 1)])[]
+  mts = message_tensors(ψψ, Z; contract_kwargs=(; alg="exact"))
```
mtfishman, line 49: I think this could still be simplified a little bit (for example making a version of `message_tensors` that does the partitioning internally, and also a different interface for the keyword arguments since `contract_kwargs` is a bit confusing in this context), but I would be fine leaving that for a future PR.

src/beliefpropagation.jl
JoeyT1994: Yeah I think that might be useful actually and help simplify that section if I can easily extract the vertices on the `subgraph` side of the `subgraph -> subgraph_neighbor` partition. I would still need to prune the unnecessary indices off of the tensors on those vertices.

mtfishman: It sounds like there are some functions we could make for `ITensorNetwork`s or partitioned `ITensorNetwork`s that would make implementing this kind of function easier.

Ultimately, given two neighboring subgraph `ITensorNetwork`s (say `tnp[1]` and `tnp[2]` above), don't you just need the common indices between those two networks to make the message tensors?

It sounds like if there was a function `commoninds(tn1::ITensorNetwork, tn2::ITensorNetwork)` that output a `Vector{Index}` with the indices shared between two networks, then we could do something like the following:
```julia
function message_tensors(subgraphs::DataGraph; maxdim=1)
  mts = DataGraph{<:Any,<:Any,ITensorNetwork}(directed_graph(underlying_graph(subgraphs)))
  for e in edges(subgraphs)
    inds_e = commoninds(subgraphs[src(e)], subgraphs[dst(e)])
    indsnetwork_e = IndsNetwork(
      path_graph(length(inds_e));
      site_space=Dictionary(eachindex(inds_e), inds_e),
      link_space=maxdim,
    )
    mts[e] = randomITensorNetwork(indsnetwork_e)
    mts[reverse(e)] = dag(mts[e])
  end
  return mts
end
```

JoeyT1994: In principle yes. The only caveat here is that this doesn't account for the internal structure of the boundary tensornetwork? For example if you want to initialise an MPS as the message tensor then you want the ordering of the boundary indices to reflect the order of the tensors of the boundary (say like to match the column of a PEPS) and to not be arbitrary?

JoeyT1994: Hence why I am currently keeping track of the `internal_inds` and constructing a tensor network as the message tensor which has both the correct internal and boundary inds?

JoeyT1994: Although that function above is useful and could be used to make things more streamlined. So should I add it to the PR?

mtfishman: Ah, I didn't realize that's what your function is doing. That seems pretty subtle to do in general, how are you determining the linear ordering of the MPS?

It still seems like you could use the function above, but then have a separate step that determines the linear order of `inds_e`.

That seems like a good reason to just use bond dimension one MPS message tensors to start with, since then the linear ordering doesn't matter. And then Linjian's `"density_matrix"` algorithm will determine a good ordering of the MPS message tensors after the first update step (so it sounds like you are essentially reproducing some of the work Linjian does for determining the linear ordering). It just seems like a lot of work and code for what could be a very simple function, if we simplified our goal a little bit.

mtfishman: Update: We've decided to just handle product state (rank 1) message tensors for now, so something like:
```julia
function message_tensors(subgraphs::DataGraph; itensor_constructor=randomITensor)
  mts = DataGraph{<:Any,<:Any,ITensorNetwork}(directed_graph(underlying_graph(subgraphs)))
  for e in edges(subgraphs)
    inds_e = commoninds(subgraphs[src(e)], subgraphs[dst(e)])
    mts[e] = ITensorNetwork(map(itensor_constructor, inds_e))
    mts[reverse(e)] = dag(mts[e])
  end
  return mts
end
```
where we are offering some choice about how to construct the ITensors making up the rank 1 message tensor networks with the `itensor_constructor` keyword argument.

JoeyT1994: Okay great, this is what I have now gone for with initialisation -> with the default being `delta()` for `itensor_constructor`.

JoeyT1994:
Okay I think I have made the belief propagation interface much simpler following your suggestions. Initialisation now always gives a rank 1 MPS with the on-site tensor constructor specificable by the user. 
The `datagraph` passed to `message_tensor()` must always have `ITensorNetworks` directly as the vertex data and thus should be the output of the `partition()` function.
There is also the option of `message_tensors(tn::ITensorNetwork; partition_args)` for handling the partitioning within `belief_propagation.jl`

mtfishman:
Looks great, thanks!

PR #89, opened 2023-05-21
Fix a small bug related to caching in `approx_itensornetwork`

LinjianMa:
The previous implementation has a small bug in caching so there is some duplicated computations. This PR fixes this and improves the performance


mtfishman:
Thanks!

PR #91, opened 2023-06-22
Introduce `contraction_sequence_to_digraph` and simplify `contraction_sequence_to_graph`

LinjianMa:
Given a contraction sequence, the output of  `contraction_sequence_to_graph` gives the partition information of the contraction sequence. However, multiple important information such as the contraction order, as well as the root of the contraction sequence, is lost. The output of `contraction_sequence_to_digraph` is a directed graph that keeps such informations.

Example: when the input sequence is [[1,2], [3,4]], the output of the `contraction_sequence_to_digraph` will be a graph with the structure of 
```
    [1,2,3,4]
     /    \
[1,2]  [3,4]
 /  \    /  \
1    2   3   4
```
while the output of `contraction_sequence_to_graph` will be an undirected graph with the structure of 
```
[1,2]----[3,4]
 /  \    /  \
1    2   3   4
```

### DIFF
--- a/src/contraction_tree_to_graph.jl
+++ b/src/contraction_tree_to_graph.jl
@@ -1,7 +1,7 @@
 """
 Take a contraction sequence and return a directed graph.
 """
-function contraction_sequence_to_directed_graph(contract_sequence)
+function contraction_sequence_to_graph(contract_sequence)
   g = NamedDiGraph()
   leaves = collect(Leaves(contract_sequence))
   seq_to_v = Dict()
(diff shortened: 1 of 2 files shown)


mtfishman:
@LinjianMa, @JoeyT1994 said that he thinks it's fine to just change the output format of `contraction_sequence_to_graph` to the one you propose, so then we don't need `contraction_sequence_to_directed_graph`. He had used it in some experimental code but it will be easy enough to use the new format you are proposing.

LinjianMa:
@mtfishman sure, then is `contraction_tree_leaf_bipartition` still useful?

mtfishman:
Joey was using that to implement an algorithm for approximating the norm of a tensor network given sets of environments. I guess it is dependent on the previous format so maybe we can comment it out for now.

LinjianMa:
@mtfishman are we able to merge this one?

PR #92, opened 2023-06-22
Introduce `approx_itensornetwork` with algorithm `ttn_svd`

LinjianMa:
Plus some reorganization of the files


mtfishman:
Looks good, thanks!

PR #94, opened 2023-06-28
Added Efficiency Improvements in the code. Including fast versions of…

JoeyT1994:
This update adds efficiency improvements to the `apply` and `gauge` functions by minimising the amount of graph analysis that is done. 

Additionally:
1) The apply function for a tn in the Vidal gauge places the site-index on the QR decomp relative to whether a gate is being applied.
2) A fast version of `dag(tn::AbstractITensorNetwork)` has been added
3) A function `unflattened_norm_network(tn::AbstractITensorNetwork)` has been added which is faster than using `⊗`
4) An example has been added to demonstrate different gauging routines

### DIFF
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -331,6 +331,14 @@ end
 adjoint(tn::Union{IndsNetwork,AbstractITensorNetwork}) = prime(tn)
 
 dag(tn::AbstractITensorNetwork) = map_vertex_data(dag, tn)
+#Faster version!
+function fast_dag(tn::AbstractITensorNetwork)
+  tndag = copy(tn)
+  for v in vertices(tndag)
+    setindex_preserve_graph!(tndag, dag(tndag[v]), v)
+  end
+  return tndag
+end
 
 # TODO: should this make sure that internal indices
 # don't clash?
@@ -644,6 +652,23 @@ function flatten_networks(
   kwargs...,
 )
   return flatten_networks(flatten_networks(tn1, tn2; kwargs...), tn3, tn_tail...; kwargs...)
+end
+
+#Ideally this will dispatch to inner_network but this is a temporary fast version for now
+function unflattened_norm_network(tn::AbstractITensorNetwork)
+  tnbra = rename_vertices(v -> (v, 1), data_graph(tn))
+  tndag = copy(tn)
+  for v in vertices(tndag)
+    ITensorNetworks.setindex_preserve_graph!(tndag, dag(tndag[v]), v)
+  end
+  tnket = rename_vertices(v -> (v, 2), data_graph(prime(tndag; sites=[])))
+  tntn = ITensorNetwork(union(tnbra, tnket))
+  for v in vertices(tn)
+    if !isempty(commoninds(tntn[(v, 1)], tntn[(v, 2)]))
+      add_edge!(tntn, (v, 1) => (v, 2))
+    end
+  end
+  return tntn
 end
 
 # TODO: Use or replace with `flatten_networks`
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -123,15 +123,18 @@ end
 _gate_vertices(o::ITensor, ψ) = neighbor_vertices(ψ, o)
 _gate_vertices(o::AbstractEdge, ψ) = [src(o), dst(o)]
 
-function _contract_factorized_gate(o::ITensor, ψv1, ψv2)
-  G1, G2 = factorize(o, Index[commonind(ψv1, o), commonind(ψv1, o)']; cutoff=1e-16)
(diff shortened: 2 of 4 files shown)


src/abstractitensornetwork.jl
```diff
@@ -331,6 +331,14 @@ end
 adjoint(tn::Union{IndsNetwork,AbstractITensorNetwork}) = prime(tn)
 
 dag(tn::AbstractITensorNetwork) = map_vertex_data(dag, tn)
+#Faster version!
+function fast_dag(tn::AbstractITensorNetwork)
```
mtfishman, line 335: Could just call this `dag` and keep the other one commented out in the code as a reminder to make a faster version/mode for `map_vertex_data`.

src/abstractitensornetwork.jl
```diff
@@ -646,6 +654,23 @@ function flatten_networks(
   return flatten_networks(flatten_networks(tn1, tn2; kwargs...), tn3, tn_tail...; kwargs...)
 end
 
+#Ideally this will dispatch to inner_network but this is a temporary fast version for now
+function unflattened_norm_network(tn::AbstractITensorNetwork)
```
mtfishman, line 658: Maybe just `norm_network`? We could add some options to choose if it outputs flattened or unflattened in the future.

src/abstractitensornetwork.jl
```diff
@@ -646,6 +654,23 @@ function flatten_networks(
...
+  tnbra = rename_vertices(v -> (v, 1), data_graph(tn))
+  tndag = copy(tn)
+  for v in vertices(tndag)
+    ITensorNetworks.setindex_preserve_graph!(tndag, dag(tndag[v]), v)
```
mtfishman, line 662: Shouldn't need `ITensorNetworks.` since we are in that module so `setindex_preserve_graph!` it should be in the namespace.

src/apply.jl
```diff
@@ -123,15 +123,18 @@ end
...
+  return Qᵥ₁, Rᵥ₁, Qᵥ₂, Rᵥ₂, theta
 end
 
-function _contract_factorized_gate(o::AbstractEdge, ψv1, ψv2)
-  return ψv1, ψv2
+function _contract_gate(o::AbstractEdge, ψv1, ψv2)
+  Qᵥ₁, Rᵥ₁ = qr(ψv1, uniqueinds(ψv1, ψv2))
+  Qᵥ₂, Rᵥ₂ = qr(ψv2, uniqueinds(ψv2, ψv1))
+  theta = Rᵥ₁ * Rᵥ₂
+  return Qᵥ₁, Rᵥ₁, Qᵥ₂, Rᵥ₂, theta
```
mtfishman, line 137: Nice how simple and minimal it is to choose between these two version of the algorithm!

JoeyT1994:
Not really sure what is going on in `test_apply` here (which is why the tests are failing?). The changes to 
`function ITensors.apply(
  o::Union{ITensor,NamedEdge},
  ψ::AbstractITensorNetwork,
  bond_tensors::DataGraph;
  normalize=false,
  apply_kwargs...,
)` appear to have made the apply function extremely accurate for the simple update case and hence its accuracy is far higher than the belief propagation version (whereas they were identical before).

There must be a bug there as they should probably not be that accurate..

mtfishman:
What if you switch back from `qr(...)` to `factorize(...; cutoff=1e-16)`, do the tests pass? Otherwise it seems like the rest of the changes are either independent of those tests or just code reoarganization.

mtfishman:
I'm confused by why you are saying SU gauging is now more accurate, it looks less accurate to me with the latest commit.

src/gauging.jl
```diff
@@ -23,6 +23,7 @@ function vidal_gauge(
 
   for e in edges(ψ_vidal)
     vsrc, vdst = src(e), dst(e)
+    ψvsrc, ψvdst = copy(ψ_vidal[vsrc]), copy(ψ_vidal[vdst])
```
mtfishman, line 26: The `copy` calls don't seem necessary here.

JoeyT1994:
Okay those bugs should be fixed now and all tests should pass.

src/apply.jl
```diff
@@ -199,8 +199,8 @@ function ITensors.apply(
     end
 
     if normalize
-      normalize!(ψv1)
-      normalize!(ψv2)
+      ψv1 ./= norm(ψv1)
+      ψv2 ./= norm(ψv2)
```
mtfishman, line 203: This is still an in-place operation.

If you want to make it out-of-place use:
```julia
      ψv1 /= norm(ψv1)
      ψv2 /= norm(ψv2)
```
(without the `.`).

mtfishman:
Thanks!

Issue #62, opened 2023-02-04
`OpSum` to `TTN` broken for multiple on-site operators


JoeyT1994:
Was playing around with `TTN()` and found the following code

```
using NamedGraphs
using ITensorNetworks
using ITensors

grid_dims = (2, 1)
g = named_grid(grid_dims)
s = siteinds("S=1/2", g)

os = OpSum()
os += 1.0, "Sx", (1,1)
os += 0.0, "Sy", (1,1)
H = TTN(os, s)

@show H[(1,1)]

os = OpSum()
os += 1.0, "Sx", (1,1)
H = TTN(os, s)

@show H[(1,1)]

```

realizes two different `TTN`s which contract to different values when applied to some `ψ`. Specifically the output of the above is:

<img width="392" alt="image" src="https://user-images.githubusercontent.com/51231103/216791623-299b1d1d-8324-4167-b627-54a274b833e2.png">.

This is despite the fact the `TTN`s should be identical (up to arbitrary index ids). In general, there appears to be something strange happening with the `"Sx"` operator for spin half systems --- when one adds mixes in other terms with a local `"Sx"` field the wrong `TTN` appears to get constructed.

mtfishman:
If you contract the TTN in both cases (`H[1, 1] * H[2, 1]`) does it give the correct Hamiltonian?

mtfishman:
Does it create the correct Hamiltonian for:
```julia
os = OpSum()
os += 1.0, "Sx", (1,1)
os += 1.0, "Sy", (1,1)
H = TTN(os, s)
```
? If not then it is probably an issue with having multiple on-site operators.

JoeyT1994:
I believe it returns `KroneckerProduct(Sy, IdentityMatrix)` in this case and ignore the `Sx` part.
I agree this would make sense and be consistent with what I have seen (it does not have a problem when you mix multiple two-site operators)

mtfishman:
Ok, that's helpful.

@leburgel or @emstoudenmire could you take a look at this?

PR #95, opened 2023-06-29
Fix `OpSum` to `TTN` converter for multiple on-site operators

leburgel:
In the spirit of better late than never, a fix for #62. I still need to figure out the problem with the manual svd compression so we can default to that and remove the current redundancy, but this might take some time so I just patched the current working version for now.

The issue was that the finite state machine representation just overwrote the on-site entry every time an on-site term was added; I just didn't think of on-site sums writing it. Wrapping everything in a `Sum{...}` fixes it. I added a small test.

### DIFF
--- a/src/treetensornetworks/opsum_to_ttn.jl
+++ b/src/treetensornetworks/opsum_to_ttn.jl
@@ -32,12 +32,11 @@ function crosses_vertex(t::Scaled{C,Prod{Op}}, g::AbstractGraph, v) where {C}
   return v ∈ span(t, g)
 end
 
-# annoying thing to allow sparse arrays with ITensors.Op entries
-# TODO: get rid of this
-function Base.zero(::Type{Scaled{C,Prod{Op}}}) where {C}
-  return zero(C) * Prod([Op("0")])
+# allow sparse arrays with ITensors.Sum entries
+function Base.zero(::Type{S}) where {S<:Sum}
+  return S()
 end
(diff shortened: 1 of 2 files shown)


mtfishman:
Thanks for circling back to this @leburgel!

Do you remember what still needed to be done to make the `OpSum` to `TTN` conversion feature complete with `OpSum` to `MPO` conversion in ITensors.jl? I seem to recall that it doesn't support QNs, and also that it may have had some issues for longer range interactions, even in the dense case.

leburgel:
The default current default implementation should work for any long range interactions (at least for anything I've tried so far), the only problem is that the resulting network has a bond dimension that is larger than necessary. There is an issue when I manually compress the network by merging interaction channels (similar to what is done in ITensors.jl) that I haven't figured out, so I've just switched off the compression by default for now until this is fixed. So right now everything should work correctly as is, it's just not optimal. I should have time to try and fix this soon though.

Everything with QN conservation errors by default right now I think. For fermions, there's definitely things in the converter that are just plain broken. It's hard for me to tell how easily this can be fixed though, since I'm not really familiar with how ITensors.jl handles fermions. For regular symmetries however, I don't think there should be any fundamental issues. I've just never tested any of the tree tensor network code with QN number support so this errors more as a precaution. It should be a matter of trying it and seeing whether things break at some point. I can also try to have a look at this soon.

PR #97, opened 2023-07-06
[WIP] Introduce `partitioned_contract`

LinjianMa:


src/approx_itensornetwork/density_matrix.jl
```diff
@@ -335,6 +328,97 @@ function _rem_vertex!(
   return U
 end
 
+function _update_cache_w_low_rank_projector!(
+  ::Algorithm"direct_eigen",
+  alg_graph::_DensityMartrixAlgGraph,
```
mtfishman, line 333: Just noticed this should be `_DensityMatrixAlgGraph`, not `_DensityMartrixAlgGraph`. Also I think it's best to remove the underscore.

src/approx_itensornetwork/approx_itensornetwork.jl
```diff
@@ -22,6 +22,8 @@ function approx_itensornetwork(
   partition_wo_deltas = _contract_deltas_ignore_leaf_partitions(
     binary_tree_partition; root=root
   )
+  density_matrix_alg =
+    alg isa Algorithm"density_matrix_direct_eigen" ? "direct_eigen" : "qr_svd"
```
mtfishman, line 26: It seems like the main use for `Algorithm"density_matrix_direct_eigen"` is this selection of the `density_matrix_alg`. Maybe simpler to just make `density_matrix_alg` a keyword argument directly and remove `Algorithm"density_matrix_direct_eigen"`.

src/approx_itensornetwork/approx_itensornetwork.jl
```diff
@@ -5,7 +5,7 @@ with the same binary tree structure. `root` is the root vertex of the
 pre-order depth-first-search traversal used to perform the truncations.
 """
 function approx_itensornetwork(
```
mtfishman, line 7: I forget why this was called `approx_itensornetwork`. Can this just be `contract`? It seems like it is just called from `contract` anyway.

src/contract.jl
```diff
@@ -10,10 +10,45 @@ function contract(
...
+
+function contract(
+  alg::Algorithm"partitioned_contract",
+  tn::AbstractITensorNetwork;
+  nvertices_per_partition::Integer=2,
+  backend::String="KaHyPar",
+  kwargs...,
+)
+  sequence = _partitioned_contraction_sequence(
+    tn; nvertices_per_partition=nvertices_per_partition, backend=backend
```
mtfishman, line 29:
```suggestion
    tn; nvertices_per_partition, backend
```
In the more recent versions of Julia, you don't have to repeat the argument if it has the same name as the keyword in a keyword argument.

mtfishman:
Sorry for the very long delay in reviewing this, finally finding some time to take a look.

In terms of the API, instead of the `partitioned_contract` you propose above, what if it was the following:
```julia
function contract(
  alg::Algorithm"partitioned_contract",
  tn::AbstractITensorNetwork;
  sequence,
  backend::String="KaHyPar",
  kwargs...,
)
```
`sequence` would be a partial contraction sequence, by which I mean if the vertices are `1:10`, it could be something like:
```julia
[[[1, 2, 3], [4, 5, 6]], [7, 8, 9, 10]]
```
This would complement the version that accepts `nvertices_per_partition::Integer`, which would generate a `sequence` with that format.

PR #100, opened 2023-07-07
Add square root BP and BP-based simple update gate evolution

mtfishman:
Add square root BP based on QR decompsitions, as proposed in Eq. (25) of [^1].

So far it is pretty simple. It only supports 1-site unit cells (simple BP), doesn't support early stopping based on a convergence criteria, and doesn't have an interface for outputting the square root fixed point message tensors. Mostly it's used for testing but those could all be improved pretty easily.

@JoeyT1994

I've also added support for simple update gate evolution in an arbitrary gauge using square root message tensors as proposed in Fig. 5 ii) of [^1] to this PR.

[^1]: https://arxiv.org/abs/2306.17837


src/beliefpropagation/sqrt_beliefpropagation.jl
```diff
@@ -0,0 +1,127 @@
...
+      tn, sqrt_mts[src(e)], environment_tensornetworks;
+    )
+
+    # if compute_norm
+    #   LHS, RHS = ITensors.contract(ITensor(sqrt_mts[src(e) => dst(e)])),
+    #   ITensors.contract(ITensor(new_sqrt_mts[src(e) => dst(e)]))
+    #   LHS /= sum(diag(LHS))
+    #   RHS /= sum(diag(RHS))
+    #   c += 0.5 * norm(LHS - RHS)
+    # end
```
JoeyT1994, line 50: Does the same calculation of C not apply to sqrt_mts? I.e. we can't assess convergence criteria in the same way? If not then I guess for now we should just drop these lines of code and not have this function return `c`

examples/belief_propagation/sqrt_bp.jl
```diff
@@ -0,0 +1,76 @@
...
+    2 * contract(
+      numerator_network; sequence=contraction_sequence(numerator_network; alg="optimal")
+    )[] / contract(
+      denominator_network; sequence=contraction_sequence(denominator_network; alg="optimal")
+    )[]
+
+  println("Sqrt Belief Propagation Gives Sz on Site " * string(v) * " as " * string(sz_bp))
+
+  return (; mts, mts_sqrt)
+end
```
JoeyT1994, line 76: When this example is run by tests will it actually run anything? Or do we need to call `main()` somewhere?

src/beliefpropagation/sqrt_beliefpropagation.jl
mtfishman: It's just a bit more subtle since the square root message tensors can have an arbitrary unitary gauge degree of freedom. We should be able to just square them of fix the gauge in some other way and then compare them from step to step, I just didn't implement that yet.

examples/belief_propagation/sqrt_bp.jl
mtfishman: Good point, we'll have to call `main()`. Also I should return something we can actually test, such as Sz for both BP and square root BP.

JoeyT1994: Yeah I think that would be worth while, returning both values and checking they are within a given tolerance of each other.

src/beliefpropagation/sqrt_beliefpropagation.jl
JoeyT1994: That's fair. I suppose if we square them all up to test at every iteration that could be a fairly significant computational overhead so can see why you would leave it out.

mtfishman: It's only `O(χ²)` so it should be sub-leading cost (except for linear networks) so I think it's ok, I just hadn't implemented it yet. I think it's good to put in, but I presume this will be an internal feature mostly for testing at first anyway since regular BP is faster.

examples/belief_propagation/sqrt_bp.jl
mtfishman: Fixed.

src/beliefpropagation/sqrt_beliefpropagation.jl
mtfishman: I'm going to leave this for a future PR.

PR #101, opened 2023-07-10
Heavy hex example

JoeyT1994:
This PR adds an example of doing the dynamics of the Kicked Ising model on the heavy-hex geometry.
This geometry is that seen on IBM's eagle processor.

I have also added the 2D_imaginary_TEBD example to the same examples folder and renamed it.


mtfishman:
Looks good! Does it take a long time to run (say more than a few minutes)? If it's not too long, would be nice to run it in the tests.

JoeyT1994:
Nope takes like 10 seconds once compiled. Have updated to add it to tests.

examples/dynamics/heavy_hex_ising_real_tebd.jl
```diff
@@ -0,0 +1,190 @@
...
+  vidal_to_symmetric_gauge,
+  vidal_gauge,
+  approx_network_region,
+  optimal_contraction_sequence,
+  contract,
+  symmetric_to_vidal_gauge,
+  norm_network
+
+#Remove edges from a graph involving the vertex v (either at src or dst)
+function rem_edges_involving_vertex(g::NamedGraph, v)
```
mtfishman, line 19: Doesn't this happen automatically if the vertex is removed with `rem_vertex!`?

JoeyT1994: You're right it does! Changed this now and removed the function

PR #104, opened 2023-07-14
Fix channel compression in `OpSum` to `TTN` converter

leburgel:
Fixes the manual compression in the `OpSum` to `TTN` converter. I switched the default `OpSum` to `TTN` converter to always use the compression now. As far as I can tell this now works for any combination of interactions (I extended the corresponding tests a bit), and always gives the correct compressed bond dimension.

I left the explicit finite state machine construction for now since this might be useful for debugging purposes, but it's no longer exported. Maybe this might be useful if we ever want to use a fully sparse Hamiltonian representation instead of always making a dense operator, but I can also remove it if you think that's better.

### DIFF
--- a/src/treetensornetworks/opsum_to_ttn.jl
+++ b/src/treetensornetworks/opsum_to_ttn.jl
@@ -398,7 +398,6 @@ function TTN(eltype::Type{<:Number}, os, sites::IndsNetwork; kwargs...)
   return NDTensors.convert_scalartype(eltype, TTN(os, sites; kwargs...))
 end
 
-
 # 
 # Tree adaptation of functionalities in ITensors.jl/src/physics/autompo/matelem.jl
 # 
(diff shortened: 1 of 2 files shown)


mtfishman:
Thanks @leburgel! I like the idea of keeping the finite state machine construction, there were some applications I had in mind for that (for example using symbolic representations of TTNOs).

@b-kloss will be interested in this.

Would this code also work with QNs? We could ignore fermions for now, since we are moving to an automatic fermion sign system anyway. @emstoudenmire

leburgel:
I still need to write the explicit QN version (similar to src/physics/autompo/opsum_to_mpo_qn.jl), I just wanted to fix the regular compression first. It doesn't look like there would be any real issues with this if I can ignore fermions, but it might take a while to actually do it.

mtfishman:
No problem, definitely no pressure for you to do it (obviously feel free to if you have the time and motivation, but reach out to us to check if someone else is also working on that since one of us might get started on it).

PR #106, opened 2023-07-23
Pass truncation kwargs to `insert_local_tensor`

leburgel:
Explicitly pass the truncation keyword arguments when inserting a local tensor back into the network in `alternating_update`. These were probably pulled out at a higher level at some point and no longer made it to `insert_local_tensor`, so the 2-site sweeping routines weren't actually truncating.


mtfishman:
Thanks! Could you add a simple test that it is truncating properly? For example setting `maxdim` and checking the output is at or below the `maxdim`.

leburgel:
Done.

PR #109, opened 2023-08-04
Use simpler constructors for heavy hex lattice

JoeyT1994:
Further update to Heavy Hex dynamics example by simplifying the graph construction procedure based on the new features in the PR https://github.com/mtfishman/NamedGraphs.jl/pull/37 in NamedGraphs.jl

Also fixed a few bugs in beliefpropagation.jl which cropped up in testing.

### DIFF
--- a/examples/dynamics/heavy_hex_ising_real_tebd.jl
+++ b/examples/dynamics/heavy_hex_ising_real_tebd.jl
@@ -32,7 +32,6 @@ function ibm_processor_graph(n::Int64, m::Int64)
   return g
 end
 
-#Create the graph of the same structure as  https://www.nature.com/articles/s41586-023-06096-3.
 eagle_processor_graph() = ibm_processor_graph(3, 6)
 hummingbird_processor_graph() = ibm_processor_graph(2, 4)
 osprey_processor_graph() = ibm_processor_graph(6, 12)
--- a/examples/dynamics/heavy_hex_ising_real_tebd.jl
+++ b/examples/dynamics/heavy_hex_ising_real_tebd.jl
@@ -15,10 +15,15 @@ using ITensorNetworks:
   symmetric_to_vidal_gauge,
   norm_network
 
-#Create the heavy_hex grid of the same structure as  https://www.nature.com/articles/s41586-023-06096-3.
-function create_eagle_processor()
-  g = hexagonal_lattice_graph(3, 6)
+function heavy_hex_lattice(n::Int64, m::Int64)
+  g = hexagonal_lattice_graph(n, m)
   g = decorate_graph_edges(g)
+  return g
+end
+
+#Create the graph of the same structure as  https://www.nature.com/articles/s41586-023-06096-3.
+function eagle_processor_graph()
+  g = heavy_hex_lattice(3, 6)
   add_vertices!(g, [(1, 8), (7, 1)])
   add_edge!(g, (1, 7) => (1, 8))
   add_edge!(g, (7, 2) => (7, 1))
@@ -48,7 +53,7 @@ end
 function main(θh::Float64, no_trotter_steps::Int64; apply_kwargs...)
 
   #Build the graph
-  g = create_eagle_processor()
+  g = eagle_processor_graph()
 
   #Do this if measuring a Z based expectation value (i.e. ignore ZZ_gates in final layer as they are irrelevant)
   shortened_final_layer = true
--- a/examples/dynamics/heavy_hex_ising_real_tebd.jl
+++ b/examples/dynamics/heavy_hex_ising_real_tebd.jl
@@ -17,11 +17,11 @@ using ITensorNetworks:
 
 #Create the heavy_hex grid of the same structure as  https://www.nature.com/articles/s41586-023-06096-3.
 function create_eagle_processor()
-  g = hexagonal_lattice_graph(3,6)
+  g = hexagonal_lattice_graph(3, 6)
   g = decorate_graph_edges(g)
   add_vertices!(g, [(1, 8), (7, 1)])
-  add_edge!(g, (1,7) => (1,8))
-  add_edge!(g, (7,2) => (7,1))
+  add_edge!(g, (1, 7) => (1, 8))
+  add_edge!(g, (7, 2) => (7, 1))
   return g
 end
 
--- a/examples/dynamics/heavy_hex_ising_real_tebd.jl
+++ b/examples/dynamics/heavy_hex_ising_real_tebd.jl
@@ -1,7 +1,7 @@
 using ITensors
 using ITensorNetworks
 using NamedGraphs
-using NamedGraphs: rem_edges!, add_edge!
+using NamedGraphs: rem_edges!, add_edge!, decorate_graph_edges, hexagonal_lattice_graph
 using Graphs
 
 using ITensorNetworks:
@@ -15,75 +15,13 @@ using ITensorNetworks:
   symmetric_to_vidal_gauge,
   norm_network
 
-#Create the heavy_hex grid of the same structure and labelling as  https://www.nature.com/articles/s41586-023-06096-3. With up to no_qubits <= 127 qubits.
-function create_heavy_hex_grid(no_qubits::Int64)
-  g = named_grid((no_qubits, 1))
-  g = rename_vertices(v -> v[1] - 1, g)
-  rung_qubits = [
-    14,
-    15,
-    16,
-    17,
-    33,
-    34,
-    35,
-    36,
-    52,
-    53,
-    54,
-    55,
-    71,
-    72,
-    73,
-    74,
-    90,
-    91,
-    92,
-    93,
-    109,
-    110,
-    111,
-    112,
-  ]
-  qubit_pairs = [
-    (0, 18),
-    (4, 22),
-    (8, 26),
-    (12, 30),
-    (20, 39),
-    (24, 43),
-    (28, 47),
-    (32, 51),
-    (37, 56),
-    (41, 60),
-    (45, 64),
-    (49, 68),
-    (58, 77),
-    (62, 81),
-    (66, 85),
-    (70, 89),
-    (75, 94),
-    (79, 98),
-    (83, 102),
-    (87, 106),
-    (96, 114),
-    (100, 118),
-    (104, 122),
-    (108, 126),
-  ]
-
-  for (i, qp) in enumerate(qubit_pairs)
-    rq = rung_qubits[i]
-    rem_vertex!(g, rq)
-    add_vertex!(g, rq)
-    if qubit_pairs[i][1] <= (no_qubits - 1) && rq <= (no_qubits - 1)
-      add_edge!(g, qubit_pairs[i][1] => rq)
-    end
-    if qubit_pairs[i][2] <= (no_qubits - 1) && rq <= (no_qubits - 1)
-      add_edge!(g, rq => qubit_pairs[i][2])
-    end
-  end
-
+#Create the heavy_hex grid of the same structure as  https://www.nature.com/articles/s41586-023-06096-3.
+function create_eagle_processor()
+  g = hexagonal_lattice_graph(3,6)
+  g = decorate_graph_edges(g)
+  add_vertices!(g, [(1, 8), (7, 1)])
+  add_edge!(g, (1,7) => (1,8))
+  add_edge!(g, (7,2) => (7,1))
   return g
 end
 
@@ -110,8 +48,7 @@ end
 function main(θh::Float64, no_trotter_steps::Int64; apply_kwargs...)
 
   #Build the graph
-  no_qubits = 127
-  g = create_heavy_hex_grid(no_qubits)
+  g = create_eagle_processor()
 
   #Do this if measuring a Z based expectation value (i.e. ignore ZZ_gates in final layer as they are irrelevant)
   shortened_final_layer = true
--- a/examples/dynamics/heavy_hex_ising_real_tebd.jl
+++ b/examples/dynamics/heavy_hex_ising_real_tebd.jl
@@ -26,8 +26,8 @@ function ibm_processor_graph(n::Int64, m::Int64)
   dims = maximum(vertices(hexagonal_lattice_graph(n, m)))
   v1, v2 = (1, dims[2]), (dims[1], 1)
   add_vertices!(g, [v1, v2])
-  add_edge!(g, v1 => (v1[1], v1[2] - 1))
-  add_edge!(g, v2 => (v2[1], v2[2] + 1))
+  add_edge!(g, v1 => v1 .- (0, 1))
+  add_edge!(g, v2 => v2 .+ (0, 1))
 
   return g
 end
--- a/examples/dynamics/heavy_hex_ising_real_tebd.jl
+++ b/examples/dynamics/heavy_hex_ising_real_tebd.jl
@@ -15,20 +15,27 @@ using ITensorNetworks:
   symmetric_to_vidal_gauge,
   norm_network
 
-function heavy_hex_lattice(n::Int64, m::Int64)
+function heavy_hex_lattice_graph(n::Int64, m::Int64)
   g = hexagonal_lattice_graph(n, m)
   g = decorate_graph_edges(g)
   return g
 end
 
-#Create the graph of the same structure as  https://www.nature.com/articles/s41586-023-06096-3.
-function eagle_processor_graph()
-  g = heavy_hex_lattice(3, 6)
-  add_vertices!(g, [(1, 8), (7, 1)])
-  add_edge!(g, (1, 7) => (1, 8))
-  add_edge!(g, (7, 2) => (7, 1))
+function ibm_processor_graph(n::Int64, m::Int64)
+  g = heavy_hex_lattice_graph(n, m)
+  dims = maximum(vertices(hexagonal_lattice_graph(n, m)))
+  v1, v2 = (1, dims[2]), (dims[1], 1)
+  add_vertices!(g, [v1, v2])
+  add_edge!(g, v1 => (v1[1], v1[2] - 1))
+  add_edge!(g, v2 => (v2[1], v2[2] + 1))
+
   return g
 end
+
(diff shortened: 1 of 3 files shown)


examples/dynamics/heavy_hex_ising_real_tebd.jl
```diff
@@ -15,75 +15,13 @@ using ITensorNetworks:
...
-    if qubit_pairs[i][1] <= (no_qubits - 1) && rq <= (no_qubits - 1)
-      add_edge!(g, qubit_pairs[i][1] => rq)
-    end
-    if qubit_pairs[i][2] <= (no_qubits - 1) && rq <= (no_qubits - 1)
-      add_edge!(g, rq => qubit_pairs[i][2])
-    end
-  end
-
+#Create the heavy_hex grid of the same structure as  https://www.nature.com/articles/s41586-023-06096-3.
+function create_eagle_processor()
```
mtfishman, line 19: I think `eagle_processor_graph` would be a better name.

Also, I guess it is easy enough to make it generic to any numbers of loops:
```julia
function eagle_processor_graph(dims::Tuple)
  g = hexagonal_lattice_graph(dims)
  g = decorate_graph_edges(g)
  # Define new_x, new_Y
  add_vertices!(g, [(1, new_y), (new_x, 1)])
  add_edge!(g, (1, new_y - 1) => (1, new_y))
  add_edge!(g, (new_x, 2) => (new_x, 1))
end

eagle_processor_graph(dims::Integer...) = eagle_processor_graph(dims)
```
(I didn't work out what the dangling vertices should be called in general).

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -99,7 +99,7 @@ function belief_propagation_iteration(
       ITensors.contract(ITensor(new_mts[src(e) => dst(e)]))
       LHS /= sum(diag(LHS))
       RHS /= sum(diag(RHS))
-      c += 0.5 * norm(LHS - RHS)
+      c += 0.5 * norm(dense(LHS) - dense(RHS))
```
mtfishman, line 102: I guess there was a problem subtracting `Diag` tensors?

Also, `denseblocks` would be better to use since that works for block sparse tensors as well.

examples/dynamics/heavy_hex_ising_real_tebd.jl
mtfishman: Ah, I guess I forgot that they refer to "Eagle" specifically as the 127 qubit version.

Maybe we could make a generic `ibm_processor_graph(dims)` function, and then define `eagle_processor_graph() = ibm_processor_graph(3, 6)`. I suppose that relies on the processors of different sizes having dangling indices in the same generic location, but we could always modify that part as needed.

src/beliefpropagation/beliefpropagation.jl
JoeyT1994: Yeah there was. I have changed to use `denseblocks`

examples/dynamics/heavy_hex_ising_real_tebd.jl
mtfishman: Also, the convention is usually that the dimensions are listed as `(nrows, ncols)`, for example for matrices or grid graphs:
```julia
julia> randn(3, 6)
3×6 Matrix{Float64}:
 1.75374    0.733839  -2.31459   0.229901   0.38511    0.763495
 1.07339   -0.246599   1.71149   0.534501   1.16299    1.27957
 0.703778  -1.62359    0.095052  0.68742   -0.662432  -0.0861288

julia> t = ITensorNetwork(named_grid((3, 6)); link_space=2);

julia> @visualize t;
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀t(1, 6)⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀2⡰⠁⠈⠉t(1, 5)⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⡜⠀⠀⠀⠀⠀⠀⢀⠎⠉⠒⠢2⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀t(2, 6)⡀⠀⠀⢀2⠀⠀⠀⠀⠀t(1, 4)⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⢀2⠀⠀⠀⠀t(2, 5)⠀⠀⠀⠀⠀2⠃⠀⠀⠉2⠤⢄⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀t(3, 6)⠀⠀⠀2⠎⠀⠀⠈⠉2⠤⢄⡀⡰⠁⠀⠀⠀⠀⠀t(1, 3)⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠈⠑2⠤⣀⢀⠎⠀⠀⠀⠀⠀⠀t(2, 4)⠀⠀⠀⠀⠀2⠀⠀⠀⠀2t(1, 2)⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀t(3, 5)2⠀⠀⠀⠀2⠀⠀⠀⠀⠀2t(2, 3)⠀⠀⠀⠀⠀⢠⠃⠉⠒2⢄⡀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈t(3, 4)⠀⠀⠀⠀⠀⠀⡜⠈⠉⠒⠤2⡀⠀⠀⢠2⠀⠀⠀t(1, 1)⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠑⠒2⣀⠀⠀⢀2⠀⠀⠀⠀⠀t(2, 2)⠀⠀⠀⠀2⠁⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀t(3, 3)⠀⠀⠀⠀⠀2⠁⠀⠀⠈t(2, 1)⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉2⠢⠤⣀⡔⠁⠀⠀⠀⠀⠀⠀⡰⠁⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀t(3, 2)⣀⡀⠀⡔2⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀t(3, 1)⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 
    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀ 

```
Following that, shouldn't the dimensions of `hexagonal_lattice_graph` be switched? It seems like they are listed as `(ncols, nrows)`.

examples/dynamics/heavy_hex_ising_real_tebd.jl
```diff
@@ -15,75 +15,18 @@ using ITensorNetworks:
...
-    rem_vertex!(g, rq)
-    add_vertex!(g, rq)
-    if qubit_pairs[i][1] <= (no_qubits - 1) && rq <= (no_qubits - 1)
-      add_edge!(g, qubit_pairs[i][1] => rq)
-    end
-    if qubit_pairs[i][2] <= (no_qubits - 1) && rq <= (no_qubits - 1)
-      add_edge!(g, rq => qubit_pairs[i][2])
-    end
-  end
+function heavy_hex_lattice(n::Int64, m::Int64)
```
mtfishman, line 18: I would call this `heavy_hex_graph` with the convention that the function is named after what type of object it outputs (when that might be ambiguous).

examples/dynamics/heavy_hex_ising_real_tebd.jl
JoeyT1994: Yeah good name for the function. I have made it a bit more generic and changed the name, although I think making the dangling vertices generic might not be needed as that graph doesn't have much meaning beyond the specific 3 x 6 IBM processor.

examples/dynamics/heavy_hex_ising_real_tebd.jl
JoeyT1994: Sure!

JoeyT1994: What about `heavy_hexagonal_lattice_graph` for consistency with `hexagonal_lattice_graph`?

mtfishman: I agree that would be more consistent but also pretty long. Maybe `heavy_hex_lattice_graph` as a compromise between the two?

examples/dynamics/heavy_hex_ising_real_tebd.jl
JoeyT1994: Yeah you're right that the rows and cols naming in `hexagonal_lattice_graph` should be switched (it names the vertices as `(col, row)` which is a mistake I made). But at the level here it is all symmetric (one could also call `heavy_hex_lattice(6,  3)` with appropriate vertex naming). It all just depends in which orientation you look at the processor.

mtfishman: Hmm I don't follow that, aren't the lattices with 6 x 3 heavy hexes or 3 x 6 heavy hexes different?

JoeyT1994: Well they're just rotations of each other by 90 degrees? So they're only different in terms of vertex naming convention. Which is dealt with in `NamedGraphs.jl` and I should change there

JoeyT1994:
Okay I implemented a more generic naming protocol and a way to determine the dangling indices. Now have functions for creating 3 of the named IBM processors, see:
https://www.oezratty.net/wordpress/wp-content/IBM-hex-lattice.png

examples/dynamics/heavy_hex_ising_real_tebd.jl
```diff
@@ -15,21 +15,28 @@ using ITensorNetworks:
...
-  add_vertices!(g, [(1, 8), (7, 1)])
-  add_edge!(g, (1, 7) => (1, 8))
-  add_edge!(g, (7, 2) => (7, 1))
+function ibm_processor_graph(n::Int64, m::Int64)
+  g = heavy_hex_lattice_graph(n, m)
+  dims = maximum(vertices(hexagonal_lattice_graph(n, m)))
+  v1, v2 = (1, dims[2]), (dims[1], 1)
+  add_vertices!(g, [v1, v2])
+  add_edge!(g, v1 => (v1[1], v1[2] - 1))
+  add_edge!(g, v2 => (v2[1], v2[2] + 1))
```
mtfishman, line 30: A bit simpler:
```suggestion
  add_edge!(g, v1 => v1 .- (0, 1))
  add_edge!(g, v2 => v2 .+ (0, 1))
```

examples/dynamics/heavy_hex_ising_real_tebd.jl
mtfishman: Maybe I'm misunderstanding, but it seems like you are implying that `hexagonal_lattice_graph(3, 6)` and `hexagonal_lattice_graph(6, 3)` output exactly the same lattice up to vertex renaming. To me that implies the hexagonal lattice is invariant under rotation of 90 degrees which doesn't sound correct to me.

mtfishman: But I do agree we should change the convention in `NamedGraphs.jl` such that `hexagonal_lattice_graph(6, 3)` corresponds to the Eagle processor layout.

JoeyT1994: I guess it depends what you mean by invariant but it's invariant in the way that `hexagonal_lattice_graph(6, 3)` is a 90 degree planar rotation of `hexagonal_lattice_graph(3, 6)`. I.e. it's just 
image
versus
image

mtfishman: So then how would you make:

6x3 hex

JoeyT1994: Okay I see what you mean and they aren't the same. In it's current implementation that drawing of your corresponds to `hexagonal_lattice_graph(6, 3)` whereas the above pictures I sent are `hexagonal_lattice_graph(3, 6)`. 

Obviously that makes it seem like we have rows and columns swapped in the function arguments but if you look at everything from a 90 degree orientation then we actually have them correct. I.e. 
image
is `hexagonal_lattice_graph(3, 6)`

mtfishman: Yeah, I agree we could have different interpretations of what the "rows" and "columns" of a hexagonal lattice mean, depending on the orientation we choose, and that would change what `hexagonal_lattice_graph(nrows, ncolumns)` means.

It seems like it is usually drawn how you just drew it in your last post, i.e. consistent with https://en.wikipedia.org/wiki/Hexagonal_lattice#Honeycomb_point_set and https://mathworld.wolfram.com/HexagonalGrid.html. So maybe we could go with that definition.

mtfishman: So with that definition, I guess the function in NamedGraphs.jl is fine? Or is there still a vertex naming issue, i.e. do the row and column labels of the vertices need to be switched?

JoeyT1994: Yeah that makes sense as I took the function from `networkx` which probably adopts the `correct' convention.
Nonetheless there is still a vertex naming issue in `NamedGraphs.jl` because I named the vertices as `(col, row)` instead of `(row, col)`. Which I will make a PR to change.

JoeyT1994: Should be addressed in: https://github.com/mtfishman/NamedGraphs.jl/pull/38

mtfishman:
Looks great, thanks! Nice that it is so simple to make these exotic lattices.

PR #110, opened 2023-08-07
`add(::ITensorNetwork, ::ITensorNetwork)` with `directsum` backend

JoeyT1994:
This PR adds the capability to add two `itensornetworks` provided they are of the same structure (edges, vertices).

Tests are included.

### DIFF
--- a/src/ITensorNetworks.jl
+++ b/src/ITensorNetworks.jl
@@ -46,7 +46,6 @@ using ITensors:
   dim,
   orthocenter,
   ProjMPS,
-  add,
   set_nsite!
 using KrylovKit: exponentiate, eigsolve, linsolve
 using NamedGraphs:
--- a/src/ITensorNetworks.jl
+++ b/src/ITensorNetworks.jl
@@ -46,6 +46,7 @@ using ITensors:
   dim,
   orthocenter,
   ProjMPS,
+  add,
   set_nsite!
 using KrylovKit: exponentiate, eigsolve, linsolve
 using NamedGraphs:
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -870,21 +870,28 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
   return inds
 end
 
-"""Add two itensornetworks together by growing the bond dimension. The network structures need to be identical (same edges, vertex names) and the local
-tensors must have the same site indices but can have different link indices"""
+"""Add two itensornetworks together by growing the bond dimension. The network structures need to be have the same vertex names, same site index on each vertex """
 function add_itensornetworks(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
   @assert issetequal(vertices(tn1), vertices(tn2))
+  @assert issetequal(edges(tn1), edges(tn2))
 
   tn12 = copy(tn1)
-  edges_tn12 = unique(vcat(edges(tn1), edges(tn2)))
+  edges_tn12 = edges(tn1)
+  #Create vertices of tn12 as direct sum of tn1[v] and tn2[v]. Make index tags those of tn1[v]
   for v in vertices(tn1)
+    tn1v_linkinds = uniqueinds(tn1[v], tn2[v])
+    tn2v_linkinds = uniqueinds(tn2[v], tn1[v])
+
+    @assert siteinds(tn1, v) == siteinds(tn2, v)
+    @assert length(tn1v_linkinds) == length(tn2v_linkinds)
     tn12[v], _ = ITensors.directsum(
-      tn1[v] => Tuple(uniqueinds(tn1[v], tn2[v])),
-      tn2[v] => Tuple(uniqueinds(tn2[v], tn1[v]));
-      tags=tags.(Tuple(uniqueinds(tn1[v], tn2[v]))),
+      tn1[v] => Tuple(tn1v_linkinds),
+      tn2[v] => Tuple(tn2v_linkinds);
+      tags=tags.(Tuple(tn1v_linkinds)),
     )
   end
 
+  #Rematch the indices
   for e in edges_tn12
     src_inds, dst_inds = inds(tn12[src(e)]), inds(tn12[dst(e)])
     ctags = intersect(tags.(src_inds), tags.(dst_inds))
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -870,6 +870,37 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
   return inds
 end
 
+"""Add two itensornetworks together by growing the bond dimension. The network structures need to be identical (same edges, vertex names) and the local
+tensors must have the same site indices but can have different link indices"""
+function add_itensornetworks(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
+  @assert issetequal(vertices(tn1), vertices(tn2))
+
+  tn12 = copy(tn1)
+  edges_tn12 = unique(vcat(edges(tn1), edges(tn2)))
+  for v in vertices(tn1)
+    tn12[v], _ = ITensors.directsum(
+      tn1[v] => Tuple(uniqueinds(tn1[v], tn2[v])),
+      tn2[v] => Tuple(uniqueinds(tn2[v], tn1[v]));
+      tags=tags.(Tuple(uniqueinds(tn1[v], tn2[v]))),
+    )
+  end
+
+  for e in edges_tn12
+    src_inds, dst_inds = inds(tn12[src(e)]), inds(tn12[dst(e)])
+    ctags = intersect(tags.(src_inds), tags.(dst_inds))
+    src_inds = filter(x -> tags(x) ∈ ctags, src_inds)
+    dst_inds = filter(x -> tags(x) ∈ ctags, dst_inds)
+
+    @assert length(src_inds) === length(dst_inds)
+
+    for (i, ind) in enumerate(src_inds)
+      tn12[dst(e)] = replaceind(tn12[dst(e)], dst_inds[i], ind)
+    end
+  end
+
+  return tn12
+end
+
 ## # TODO: should this make sure that internal indices
 ## # don't clash?
 ## function hvncat(
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -884,8 +884,8 @@ is_multi_edge(tn::AbstractITensorNetwork, e) = length(linkinds(tn, e)) > 1
 function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
   @assert issetequal(vertices(tn1), vertices(tn2))
 
-  tn1 = combine_linkinds(tn1; edges=filter(e -> length(linkinds(tn1, e)) > 1, edges(tn1)))
-  tn2 = combine_linkinds(tn2; edges=filter(e -> length(linkinds(tn2, e)) > 1, edges(tn2)))
+  tn1 = combine_linkinds(tn1; edges=filter(e -> is_multi_edge(tn1, e), edges(tn1)))
+  tn2 = combine_linkinds(tn2; edges=filter(e -> is_multi_edge(tn2, e), edges(tn2)))
 
   edges_tn1, edges_tn2 = edges(tn1), edges(tn2)
 
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -913,7 +913,7 @@ function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
     e1_v = NamedEdge[NamedEdge(v => vn) for vn in neighbors(tn1, v)]
     e2_v = NamedEdge[NamedEdge(v => vn) for vn in neighbors(tn2, v)]
 
-    @assert Set(e1_v) == Set(e2_v)
+    @assert issetequal(e1_v, e2_v)
     tn1v_linkinds = Index[first(linkinds(tn1, e)) for e in e1_v]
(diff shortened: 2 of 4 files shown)


mtfishman:
Nice to see how simple it is!

I guess we could support networks that don't have the same structure pretty easily by adding trivial indices to edges where one TN has indices and the other doesn't, I think I even have a function somewhere in the library for adding trivial indices to some set of edges.

When we generalize `alternating_update` to ITensorNetworks, we can also provide a "fitting" algorithm as another backend for adding tensor networks, where we optimize the fidelity between a sum of ITensorNetworks and a new ITensorNetwork.

src/abstractitensornetwork.jl
```diff
@@ -870,6 +870,44 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
   return inds
 end
 
+"""Add two itensornetworks together by growing the bond dimension. The network structures need to be have the same vertex names, same site index on each vertex """
+function add_itensornetworks(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
```
mtfishman, line 874: We could just call this `add(...)`, and then also provide an alias `+(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork) = add(tn1, tn2)` (where `+` would be imported from `Base`.

mtfishman: Note that ITensors.jl exports `add` so probably we also want to import `add` from ITensors.

JoeyT1994: Okay cool I believe I did this.

src/ITensorNetworks.jl
```diff
@@ -46,6 +46,7 @@ using ITensors:
   dim,
   orthocenter,
   ProjMPS,
+  add,
```
mtfishman, line 49: To import `add` to overload it, you need to use `import ITensors: add`. `using ITensors: add` just brings it into scope but doesn't make it overloadable.

JoeyT1994:
> Nice to see how simple it is!
> 
> I guess we could support networks that don't have the same structure pretty easily by adding trivial indices to edges where one TN has indices and the other doesn't, I think I even have a function somewhere in the library for adding trivial indices to some set of edges.

Okay I think I added this support now. So the networks need to have the same vertices and siteindexes but can have different different edge structures. The link indices need to be named properly (i.e. using standard `itensornetworks' edge tag convention) though as the function needs tag info to determine how to properly do the direct sum etc (which indexes should be combined between the two networks).

src/ITensorNetworks.jl
JoeyT1994: I see, changed that now

mtfishman:
Ideally we would do it without involving tags, just direct summing the corresponding edges. Why do you need to involve tags?

mtfishman:
Is the situation you are addressing one where there are multiple link indices on an edge of one of the input tensor networks? If that's the case and the question is about how to sort those indices to combine them into a single index, I would say just combine them in whatever order you find them on the network (say using the `combine_linkinds` function).

src/abstractitensornetwork.jl
```diff
@@ -870,6 +870,59 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
...
 
+"""Add two itensornetworks together by growing the bond dimension. The network structures need to be have the same vertex names, same site index on each vertex """
+function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
+  @assert issetequal(vertices(tn1), vertices(tn2))
+
+  if !issetequal(edges(tn1), edges(tn2))
+    new_edges = union(edges(tn1), edges(tn2))
+    tn1 = insert_missing_internal_inds(tn1, new_edges)
+    tn2 = insert_missing_internal_inds(tn2, new_edges)
+    return add(tn1, tn2)
```
mtfishman, line 881: I would remove this line and let it continue executing the rest of `add`, so it doesn't do these checks again and avoids any potential infinite loops.

src/abstractitensornetwork.jl
```diff
@@ -870,6 +870,59 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
...
+
+    for (i, ind) in enumerate(src_inds)
+      tn12[dst(e)] = replaceind(tn12[dst(e)], dst_inds[i], ind)
+    end
+  end
+
+  return tn12
+end
+
+Base.:+(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork) = add(tn1, tn2)
```
mtfishman, line 924: Let's instead import `+` from `Base` with `import Base: +`.

JoeyT1994:
Yeah I think I can see now that this can all be worked out based on edges and need not rely on matching tags anywhere. Let me re-write the function that way!

mtfishman:
> Yeah I think I can see now that this can all be worked out based on edges and need not rely on matching tags anywhere. Let me re-write the function that way!

My goal is basically to always avoid using tags in Index logic, and essentially just treat them like they are just used for printing (there are a few exceptions to that but for most code they can be treated that way). It basically always makes the code simpler and also much more general, since we want to allow users to input networks with indices that arbitrary tags.

src/abstractitensornetwork.jl
```diff
@@ -870,6 +870,77 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
...
 
+"""Check if an itensornetwork has multiple indices along at least one of its edges"""
+function has_multi_edge(tn::AbstractITensorNetwork)
+  for e in edges(tn)
+    if length(linkinds(tn, e)) > 1
+      return false
+    end
+  end
+
+  return true
```
mtfishman, line 881:
```suggestion
  any(e -> length(linkinds(tn, e)) > 1, edges(tn))
```

src/abstractitensornetwork.jl
```diff
@@ -870,6 +870,77 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
...
+  tn12 = copy(tn1)
+  edge_index_map = Dict{NamedEdge,Index}()
+  #Create vertices of tn12 as direct sum of tn1[v] and tn2[v]. Work out the matching indices by matching edges. Make index tags those of tn1[v]
+  for v in vertices(tn1)
+    @assert siteinds(tn1, v) == siteinds(tn2, v)
+    e1_v = NamedEdge[NamedEdge(v => vn) for vn in neighbors(tn1, v)]
+    e2_v = NamedEdge[NamedEdge(v => vn) for vn in neighbors(tn2, v)]
+
+    @assert Set(e1_v) == Set(e2_v)
+    tn1v_linkinds = Index[first(linkinds(tn1, e)) for e in e1_v]
```
mtfishman, line 917: Could use `only(linkinds(tn1, e))` instead of `first(linkinds(tn1, e))` to add an assertion that the edge has only one Index.

src/abstractitensornetwork.jl
```diff
@@ -870,6 +870,77 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
...
+
+  tn12 = copy(tn1)
+  edge_index_map = Dict{NamedEdge,Index}()
+  #Create vertices of tn12 as direct sum of tn1[v] and tn2[v]. Work out the matching indices by matching edges. Make index tags those of tn1[v]
+  for v in vertices(tn1)
+    @assert siteinds(tn1, v) == siteinds(tn2, v)
+    e1_v = NamedEdge[NamedEdge(v => vn) for vn in neighbors(tn1, v)]
+    e2_v = NamedEdge[NamedEdge(v => vn) for vn in neighbors(tn2, v)]
+
+    @assert Set(e1_v) == Set(e2_v)
```
mtfishman, line 916:
```suggestion
    @assert issetequal(e1_v, e2_v)
```

mtfishman:
@JoeyT1994 I see now where some of the complication comes from, that when you direct sum indices of neighboring tensors then the direct-summed indices don't match anymore since each call to `directsum` generates new indices internally. I think I have a nice solution for that. In ITensors.jl, I'm working on a PR where you can provide a direct-summed Index to `directsum`, so then you can make them beforehand for each edge so then they are guaranteed to already match, so then you don't have to make them match after the fact.

I'll try to finish that ASAP so you can use it here to simplify your code.

mtfishman:
Will be implemented by https://github.com/ITensor/ITensors.jl/pull/1176.

mtfishman:
@JoeyT1994 I merged https://github.com/ITensor/ITensors.jl/pull/1176, hopefully that helps here to make this function a bit simpler.

JoeyT1994:
Okay great, thanks! Let me edit based on that and use this as a test for that new interface.

JoeyT1994:
Okay I just took advantage of that new functionality! Makes writing the function easier.

src/abstractitensornetwork.jl
```diff
@@ -870,6 +870,75 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
...
+  return any(e -> length(linkinds(tn, e)) > 1, edges(tn))
+end
+
+"""Add two itensornetworks together by growing the bond dimension. The network structures need to be have the same vertex names, same site index on each vertex """
+function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
+  @assert issetequal(vertices(tn1), vertices(tn2))
+
+  #Check for any edges with multiple indices and combine to avoid
+  if has_multi_edge(tn1)
+    tn1 = combine_linkinds(tn1)
```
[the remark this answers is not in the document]
JoeyT1994: Yeah I was thinking about whether this logic should be handled by `combine_linkinds` and I think that is the cleanest. Lets do that.

src/abstractitensornetwork.jl
```diff
@@ -591,20 +591,22 @@ function neighbor_vertices(ψ::AbstractITensorNetwork, T::ITensor)
   return first.(v⃗)
 end
 
-function linkinds_combiners(tn::AbstractITensorNetwork)
+function linkinds_combiners(tn::AbstractITensorNetwork; edges_to_combine=edges(tn))
```
mtfishman, line 594: I think calling it `edges` is fine, I don't think people will get confused about what that is referring to.

src/abstractitensornetwork.jl
```diff
@@ -879,14 +881,18 @@ end
...
-
-  if has_multi_edge(tn2)
-    tn2 = combine_linkinds(tn2)
-  end
+  tn1 = combine_linkinds(
+    tn1,
+    linkinds_combiners(
+      tn1; edges_to_combine=filter(e -> length(linkinds(tn1, e)) > 1, edges(tn1))
+    ),
+  )
```
mtfishman, line 889: Seems simpler to just add an `edges` keyword argument directly to `combine_linkinds`.

src/abstractitensornetwork.jl
```diff
@@ -591,20 +591,22 @@ function neighbor_vertices(ψ::AbstractITensorNetwork, T::ITensor)
...
+    if e ∈ edges_to_combine
+      C = combiner(linkinds(tn, e); tags=edge_tag(e))
+      combiners[e] = C
+      combiners[reverse(e)] = dag(C)
+    end
   end
   return combiners
 end
 
 function combine_linkinds(tn::AbstractITensorNetwork, combiners=linkinds_combiners(tn))
```
mtfishman, line 606: One way to organize this would be to define two versions:
```julia
function combine_linkinds(tn::AbstractITensorNetwork, combiners::DataGraph)
  # Implementation of combining the link indices
end

function combine_linkinds(tn::AbstractITensorNetwork; edges::Vector{<:Union{Pair,AbstractEdge}}=edges(tn))
  combiners = linkinds_combiners(tn; edges)
  return combine_linkinds(tn, combiners)
end
```

src/abstractitensornetwork.jl
JoeyT1994: Mmmm but `combine_linkinds` by default populates a `combiner` along every edge of the network and I want to avoid it doing that? So like I originally had
`combine_linkinds(tn::AbstractITensorNetwork, combiners=linkinds_combiners(tn); edges = edges(tn)` but then I don't want the user to call `combine_linkinds(tn; edges = edge_subset)` which would create combiners on every edge and end up not using them, making them redundant. Unless it's possible to do
`combine_linkinds(tn, combiners=linkinds_combiners(tn; edges); edges = edges(tn)`?

src/abstractitensornetwork.jl
JoeyT1994: Okay yeah this resolves my above concern

JoeyT1994: Although this is technically not allowed surely because `combine_linkinds(tn)' then points to both functions?

mtfishman: The version `function combine_linkinds(tn::AbstractITensorNetwork, combiners::DataGraph) [...] end` I proposed above wouldn't provide a default value so they are distinct functions.

src/abstractitensornetwork.jl
```diff
@@ -870,6 +877,67 @@ function ITensors.commoninds(tn1::AbstractITensorNetwork, tn2::AbstractITensorNe
   return inds
 end
 
+"""Check if the edge of an itensornetwork has multiple indices"""
+is_multi_edge(tn::AbstractITensorNetwork, e) = length(linkinds(tn, e)) > 1
```
mtfishman, line 881: Something neat you can do is define a single-argument version:
```julia
is_multi_edge(tn::AbstractITensorNetwork) = Base.Fix1(is_multi_edge, tn)
```
Then, `is_multi_edge(tn)(e)` is the same as `is_multi_edge(tn, e)`. This can simplify code where you want to pass `is_multi_edge` as a map or filter function, for example:
```julia
filter(is_multi_edge(tn), edges(tn))
```

JoeyT1994: Oh nice! That's cool. I have added that now.

PR #111, opened 2023-10-02
Belief propagation order flexibility

JoeyT1994:
This PR adds the ability to do iterate the belief propagation self-consistent equations in parallel or in sequence. It also adds a keyword argument for iterating over only a subset of edges instead of all edges of the network.

Tests are updated to reflect this. 
The BP gauging example in `examples/gauging/gauging_itns.jl` now also benchmarks for parallel and sequential updates.


src/beliefpropagation/beliefpropagation.jl
```diff
@@ -81,17 +81,25 @@ function belief_propagation_iteration(
   mts::DataGraph;
   contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
   compute_norm=false,
+  update_order::String="Parallel",
```
mtfishman, line 84: Let's make it lower case (i.e. `"parallel"`).

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -81,17 +81,25 @@ function belief_propagation_iteration(
...
+  es=edges(mts),
```
mtfishman, line 85: I would call it `edges`, `es` is a bit cryptic. You can use `Graphs.edges` to refer to the function.

src/beliefpropagation/beliefpropagation.jl
mtfishman: Also `"Sequential"` to `"sequential"`.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -81,17 +81,25 @@ function belief_propagation_iteration(
   mts::DataGraph;
   contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
   compute_norm=false,
+  update_order::String="parallel",
```
mtfishman, line 84: I think it would be better to call it `update_sequence` to match our convention of using the term "contract sequence" in other parts of the code.

JoeyT1994:
Okay I have added commits to this to take advantage of the PR: https://github.com/mtfishman/NamedGraphs.jl/pull/39 in `NamedGraphs`. I have also made it so parallel vs sequential is specified by `edges = Vector{Vector{NamedEdge{}}}` where the internal vectors in the nest are done sequentially and the external ones in parallel. Currently the parallel updates involve a full copy of the message tensors `DataGraph` for each parallel instance. When we move to a better partition structure for them I will optimise this that to avoid over allocating memory - which can be excessive in the case of a large number of parallel groups (e.g. in a fully parallel update sequence).
For now I think it is okay and the default is all updates done sequentially anyway.
I have provided plenty of sequence examples in `examples/belief_propagation/bpsequences.jl` and you can see the power of our custom sequence finder there as it gives pretty significant speedups in every single case.

mtfishman:
If they don't take too long to run, I think we should run `examples/belief_propagation/bpsequences.jl` from the tests. I feel uncomfortable accumulating all of these example files which are not being run in the tests, it's easy for things to become outdated.

mtfishman:
There does seem to be an issue that this test suite takes a long time to run...

JoeyT1994:
Yeah it doesn't too long so I will add it. Is there a reason for there being a separate `test_sqrt_bp.jl` file whilst it is also included in the `test_examples.jl` file? Which means the test gets run twice. If not, I will delete the `test_sqrt_bp.jl` file.

I also agree that we do have a lot of tests in the package and these can take a long time to run! We may have to think about streamlining tests a bit/ removing some which we consider less important from the `runtests.jl` list.

mtfishman:
> Yeah it doesn't too long so I will add it. Is there a reason for there being a separate `test_sqrt_bp.jl` file whilst it is also included in the `test_examples.jl` file? Which means the test gets run twice. If not, I will delete the `test_sqrt_bp.jl` file.

Hmm I guess `test_sqrt_bp.jl` is actually testing the output so seems better to keep that one and not include it in `test_examples.jl`.

examples/gauging/gauging_itns.jl
```diff
@@ -23,6 +22,7 @@ using ITensorNetworks:
 using NamedGraphs
 using NamedGraphs: add_edges!, rem_vertex!, hexagonal_lattice_graph
 using Graphs
+using MKL
```
JoeyT1994, line 25:
```suggestion
```

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -172,3 +212,24 @@ function approx_network_region(
...
+  for forest in forests
+    trees = NamedGraph[forest[vs] for vs in connected_components(forest)]
+    for tree in trees
+      tree_edges = post_order_dfs_edges(tree, root_vertex(tree))
+      push!(edges, vcat(tree_edges, reverse(reverse.(tree_edges)))...)
+    end
+  end
+
+  return edges
+end
```
JoeyT1994, line 235: Let's move this to `beleiefpropagation/beliefpropagation_schedules.jl` and have a range of functions for building schedules.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -172,3 +212,24 @@ function approx_network_region(
 
   return environment_tn ⊗ verts_tn
 end
+
+"""
+Return a custom edge order for how how to update all BP message tensors on a general undirected graph.
```
JoeyT1994, line 217:
```suggestion
Return a custom edge order for how to update all BP message tensors on a general undirected graph. 
```

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -102,7 +103,43 @@ function belief_propagation_iteration(
...
+    updated_mts, ct = belief_propagation_iteration(
+      tn, mts, edges; contract_kwargs, compute_norm
+    )
+    for e in edges
+      new_mts[e] = updated_mts[e]
+    end
+    c += ct
+  end
+  return new_mts, c / (length(edge_groups))
+end
```
JoeyT1994, line 131: Assert that each `Vector{E}` is disjoint from every other

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -102,7 +103,43 @@ function belief_propagation_iteration(
...
+  mts::DataGraph,
+  edge_groups::Vector{Vector{E}};
+  contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
+  compute_norm=false,
+) where {E<:NamedEdge}
+  new_mts = copy(mts)
+  c = 0
+  for edges in edge_groups
+    updated_mts, ct = belief_propagation_iteration(
+      tn, mts, edges; contract_kwargs, compute_norm
```
JoeyT1994, line 123: Here we are `sending` the full `mts` structure over the whole graph but only need a subset of this structure. So we should think about improving the structure and only sending the relevant parts (probably for a future PR)

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -102,7 +103,43 @@ function belief_propagation_iteration(
...
+  return new_mts, c / (length(edges))
+end
+
+"""
+Do parallel updates between groups of edges of all message tensors for a given ITensornetwork and its partition into sub graphs
+"""
+function belief_propagation_iteration(
+  tn::ITensorNetwork,
+  mts::DataGraph,
+  edge_groups::Vector{Vector{E}};
```
mtfishman, line 115:
```suggestion
  edge_groups::Vector{Vector{<:AbstractEdge}};
```
In general I think we should use `AbstractEdge` instead of `NamedEdge` for dispatch where possible to make the code more generic.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -102,7 +103,43 @@ function belief_propagation_iteration(
...
+  contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
+  compute_norm=false,
+) where {E<:NamedEdge}
```
mtfishman, line 118:
```suggestion
)
```

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -111,16 +148,18 @@ function belief_propagation(
   contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
   niters=20,
   target_precision::Union{Float64,Nothing}=nothing,
-)
-  compute_norm = target_precision == nothing ? false : true
+  edges::Union{Vector{Vector{E}},Vector{E}}=belief_propagation_edge_sequence(
```
mtfishman, line 151:
```suggestion
  edges::Union{Vector{<:Vector{<:AbstractEdge}},<:Vector{<:AbstractEdge}}=belief_propagation_edge_sequence(
```

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -135,9 +174,10 @@ function belief_propagation(
   subgraph_vertices=nothing,
   niters=20,
   target_precision::Union{Float64,Nothing}=nothing,
```
mtfishman, line 176:
```suggestion
  target_precision=nothing,
```
I generally don't put type assertions on keyword arguments.

mtfishman: Again, to make things more generic, for example maybe you want to use single precision in which case it makes sense to pass something that is `Float32`.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -172,3 +212,24 @@ function approx_network_region(
...
+On a tree this will yield a sequence which only needs to be performed once. Based on forest covers and depth first searches amongst the forests.
+"""
+function belief_propagation_edge_sequence(
+  g::NamedGraph; root_vertex=NamedGraphs.default_root_vertex
+)
+  @assert !is_directed(g)
+  forests = NamedGraphs.forest_cover(g)
+  edges = NamedEdge[]
+  for forest in forests
+    trees = NamedGraph[forest[vs] for vs in connected_components(forest)]
```
mtfishman, line 227:
```suggestion
    trees = [forest[vs] for vs in connected_components(forest)]
```
Julia will figure out the element type of the collection automatically.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -172,3 +212,24 @@ function approx_network_region(
...
+"""
+Return a custom edge order for how to update all BP message tensors on a general undirected graph. 
+On a tree this will yield a sequence which only needs to be performed once. Based on forest covers and depth first searches amongst the forests.
+"""
+function belief_propagation_edge_sequence(
+  g::NamedGraph; root_vertex=NamedGraphs.default_root_vertex
+)
+  @assert !is_directed(g)
+  forests = NamedGraphs.forest_cover(g)
+  edges = NamedEdge[]
```
mtfishman, line 225:
```suggestion
  edges = edgetype(g)[]
```
Again, better for generic code, I think we should try to anticipate wanting to pass any type of graph into most of these functions, which could have their own edge types.

src/beliefpropagation/sqrt_beliefpropagation.jl
```diff
@@ -1,16 +1,70 @@
...
+    end
+    c += ct
+  end
+  return new_sqrt_mts, c / (length(edges))
+end
+
+function sqrt_belief_propagation_iteration(
+  tn::ITensorNetwork,
+  sqrt_mts::DataGraph;
+  edges::Union{Vector{Vector{E}},Vector{E}}=belief_propagation_edge_sequence(
```
mtfishman, line 48:
```suggestion
  edges=belief_propagation_edge_sequence(...)
```

src/beliefpropagation/sqrt_beliefpropagation.jl
```diff
@@ -1,16 +1,70 @@
...
+  ),
+) where {E<:NamedEdge}
+  return sqrt_belief_propagation_iteration(tn, sqrt_mts, edges)
+end
+
 function sqrt_belief_propagation(
   tn::ITensorNetwork,
   mts::DataGraph;
   niters=20,
+  edges::Union{Vector{Vector{E}},Vector{E}}=belief_propagation_edge_sequence(
```
mtfishman, line 59:
```suggestion
  edges=belief_propagation_edge_sequence(
```

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -102,7 +103,43 @@ function belief_propagation_iteration(
...
+  end
+  return new_mts, c / (length(edge_groups))
+end
+
+function belief_propagation_iteration(
+  tn::ITensorNetwork,
+  mts::DataGraph;
+  contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
+  compute_norm=false,
+  edges::Union{Vector{Vector{<:NamedEdge}},Vector{<:NamedEdge}}=belief_propagation_edge_sequence(
```
mtfishman, line 138:
```suggestion
  edges=belief_propagation_edge_sequence(
```

src/beliefpropagation/beliefpropagation.jl
mtfishman: Also I think we can just drop this type assertion altogether.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -146,14 +147,15 @@ function belief_propagation(
   tn::ITensorNetwork,
   mts::DataGraph;
   contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
-  niters=20,
-  target_precision::Union{Float64,Nothing}=nothing,
-  edges::Union{Vector{Vector{E}},Vector{E}}=belief_propagation_edge_sequence(
-    undirected_graph(underlying_graph(mts))
-  ),
+  niters::Union{Int64,Nothing}=default_bp_niters(mts),
```
mtfishman, line 150:
```suggestion
  niters=default_bp_niters(mts),
```

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -0,0 +1,51 @@
+abstract type EdgeSequenceAlgorithm end
+
+struct ForestCover <: EdgeSequenceAlgorithm end
+struct Parallel <: EdgeSequenceAlgorithm end
```
mtfishman, line 4: Can we use the `Algorithm` type from `NDTensors` instead?

Basically you can just replace your usage of `::Parallel` with `::Algorithm"parallel"` when dispatching on `edge_sequence`, and then use `Algorithm"parallel"()` or `Algorithm("parallel")` for instances.

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -0,0 +1,51 @@
...
+
+default_edge_sequence_alg() = ForestCover()
+
+default_bp_niters(g::NamedGraph) = is_tree(g) ? 1 : nothing
+function default_bp_niters(g::Union{DataGraph,AbstractITensorNetwork})
```
mtfishman, line 9: Seems like this could be simplified with:
```julia
@traitfn default_bp_niters(g::AbstractGraph::::(!IsDirected)) = is_tree(g) ? 1 : nothing

@traitfn default_bp_niters(g::AbstractGraph::::IsDirected) = default_bp_niters(undirected_graph(g))
```
or something like that.

mtfishman: Also looks like something similar could be used for `edge_sequence`.

src/beliefpropagation/sqrt_beliefpropagation.jl
```diff
@@ -43,26 +43,23 @@ function sqrt_belief_propagation_iteration(
...
 end
 
 function sqrt_belief_propagation(
   tn::ITensorNetwork,
   mts::DataGraph;
-  niters=20,
-  edges::Union{Vector{Vector{E}},Vector{E}}=belief_propagation_edge_sequence(
-    undirected_graph(underlying_graph(mts))
-  ),
+  niters::Union{Int64,Nothing}=default_bp_niters(mts),
```
mtfishman, line 54:
```suggestion
  niters=default_bp_niters(mts),
```

src/beliefpropagation/beliefpropagation_schedule.jl
JoeyT1994: I could make everything just dispatch on abstract graph types by doing, say:

`default_bp_niters(g::AbstractGraph) = is_tree(undirected_graph(underlying_graph(g))) ? 1 : nothing` ?

And similarly with `edge_sequence` ?

mtfishman: Hmm looks like it is defined here: https://github.com/JuliaGraphs/Graphs.jl/blob/cc3052f9e3c7486d006f3864fe42da0d4e03e5d7/src/trees/prufer.jl#L11-L13, does it break for `AbstractDataGraph`?

examples/belief_propagation/bpsequences.jl
```diff
@@ -53,7 +53,7 @@ function main()
       contract_kwargs=(; alg="exact"),
       target_precision=1e-10,
       niters=100,
-      edges=edge_sequence(mts_init; alg=ITensorNetworks.Parallel()),
+      edges=edge_sequence(mts_init; alg=ITensors.Algorithm("Parallel")),
```
mtfishman, line 56: You can add `using ITensors: Algorithm` to `src/imports.jl` to bring it `Algorithm` into the namespace.

Also you can just use `alg="parallel"` (note it should be lower case, everything should be lower case except types and modules).

Then we can wrap `alg` in `Algorithm` inside `edge_sequence`.

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -1,25 +1,24 @@
...
 end
 
 function edge_sequence(g::NamedGraph; alg=default_edge_sequence_alg())
   return edge_sequence(alg, g)
 end
 
-function edge_sequence(mts::DataGraph; alg=default_edge_sequence_alg())
-  return edge_sequence(alg, undirected_graph(underlying_graph(mts)))
+function edge_sequence(g::AbstractGraph; alg=default_edge_sequence_alg())
+  return edge_sequence(alg, undirected_graph(underlying_graph(g)))
```
mtfishman, line 13:
```suggestion
  return edge_sequence(Algorithm(alg), undirected_graph(underlying_graph(g)))
```
This change will allow you to just use string names not wrapped in `Algorithm` in other functions.

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -1,25 +1,24 @@
...
+end
+
+function edge_sequence(alg::Algorithm, g::AbstractGraph; kwargs...)
+  return edge_sequence(alg, g, kwargs...)
 end
 
 function edge_sequence(
-  ::ForestCover, g::NamedGraph; root_vertex=NamedGraphs.default_root_vertex
+  ::Algorithm"ForestCover", g::NamedGraph; root_vertex=NamedGraphs.default_root_vertex
```
mtfishman, line 21:
```suggestion
  ::Algorithm"forest_cover", g::NamedGraph; root_vertex=NamedGraphs.default_root_vertex
```

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -35,17 +34,7 @@ function edge_sequence(
...
 end
 
-function edge_sequence(
-  ::ForestCover, mts::DataGraph; root_vertex=NamedGraphs.default_root_vertex
-)
-  return edge_sequence(ForestCover, undirected_graph(underlying_graph(mts)); root_vertex)
-end
-
-function edge_sequence(::Parallel, g::NamedGraph)
+function edge_sequence(::Algorithm"Parallel", g::NamedGraph)
```
mtfishman, line 37:
```suggestion
function edge_sequence(::Algorithm"parallel", g::NamedGraph)
```

src/beliefpropagation/beliefpropagation_schedule.jl
JoeyT1994: Yeah it appears to break for `AbstractDataGraph`

mtfishman: Looks ok to me:
```julia
julia> using Graphs

julia> using NamedGraphs

julia> using DataGraphs

julia> is_tree(DataGraph(grid((4, 1))))
true

julia> is_tree(DataGraph(grid((4, 2))))
false

julia> is_tree(DataGraph(named_grid((4, 1))))
true

julia> is_tree(DataGraph(named_grid((4, 2))))
false
```

JoeyT1994: See 
image

mtfishman: Right well that's just that you are inputting a directed graph, which isn't inherently an issue with calling `is_tree` on a `DataGraph`, just that it's not defined for directed graphs. I'm being pedantic on purpose to try to hone in on what the actual issue is, i.e. `is_tree` is working for _undirected_ `DataGraph`s.

So then I understand why it probably makes sense to use `is_tree(undirected_graph(underlying_graph(g)))` since you want to make it undirected, but it probably doesn't make sense to use `undirected_graph(g)` on a directed graph, since you wouldn't know what to do with the edge data if there are edges in both directions shared between two vertices.

JoeyT1994: Yeah I see what you mean. So now I have made it so all versions of the function which take `g::AbstractGraph` type do `undirected_graph(underlying_graph(g))` and then pass to the version of the which requires an`undirected graph` (which I have made a `traitfn`)

mtfishman: Seems like there might be a better pattern for this kind of thing, like defining `undirected_is_tree` and/or `undirected_underlying_graph(g))`. But this seems like a good generic solution for now.

JoeyT1994: Yeah agreed it would be optimal to handle such things in `NamedGraphs`. Thanks for making the issue.

examples/gauging/gauging_itns.jl
```diff
@@ -45,7 +45,10 @@ end
 
 """Bring an ITN into the Vidal gauge, various methods possible. Result is timed"""
 function benchmark_state_gauging(
-  ψ::ITensorNetwork; mode="BeliefPropagation", no_iterations=50
+  ψ::ITensorNetwork;
+  mode="BeliefPropagation",
```
mtfishman, line 49:
```suggestion
  mode="belief_propagation",
```

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -0,0 +1,40 @@
+default_edge_sequence_alg() = "forest_cover"
+
+@traitfn default_bp_niters(g::NamedGraph::(!IsDirected)) = is_tree(g) ? 1 : nothing
```
mtfishman, line 3:
```suggestion
@traitfn default_bp_niters(g::::(!IsDirected)) = is_tree(g) ? 1 : nothing
```
Can just be for any undirected graph.

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -0,0 +1,40 @@
...
+function default_bp_niters(g::AbstractGraph)
+  return default_bp_niters(undirected_graph(underlying_graph(g)))
+end
+
+function edge_sequence(g::NamedGraph; alg=default_edge_sequence_alg())
```
mtfishman, line 8:
```suggestion
function edge_sequence(g::AbstractGraph; alg=default_edge_sequence_alg())
```

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -0,0 +1,40 @@
...
+  return edge_sequence(alg, g)
```
mtfishman, line 9:
```suggestion
  return edge_sequence(Algorithm(alg), g)
```
so that strings representing the algorithms can be passed as keyword arguments.

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -0,0 +1,40 @@
...
+  return edge_sequence(Algorithm(alg), undirected_graph(underlying_graph(g)))
+end
+
+function edge_sequence(alg::Algorithm, g::AbstractGraph; kwargs...)
+  return edge_sequence(alg, g, kwargs...)
+end
+
+@traitfn function edge_sequence(
+  ::Algorithm"forest_cover",
+  g::NamedGraph::(!IsDirected);
```
mtfishman, line 22:
```suggestion
  g::::(!IsDirected);
```

src/beliefpropagation/beliefpropagation_schedule.jl
```diff
@@ -0,0 +1,40 @@
...
+    for tree in trees
+      tree_edges = post_order_dfs_edges(tree, root_vertex(tree))
+      push!(edges, vcat(tree_edges, reverse(reverse.(tree_edges)))...)
+    end
+  end
+
+  return edges
+end
+
+@traitfn function edge_sequence(::Algorithm"parallel", g::NamedGraph::(!IsDirected))
```
mtfishman, line 38:
```suggestion
@traitfn function edge_sequence(::Algorithm"parallel", g::::(!IsDirected))
```

src/beliefpropagation/beliefpropagation_schedule.jl
JoeyT1994: This seems to result in an infinite loop and the trait function never gets called?
```
@traitfn default_bp_niters(g::NamedGraph::(!IsDirected)) = is_tree(g) ? 1 : nothing
function default_bp_niters(g::AbstractGraph)
  return default_bp_niters(undirected_graph(underlying_graph(g)))
end
```

src/beliefpropagation/beliefpropagation_schedule.jl
JoeyT1994: Currently `NamedGraphs.forest_cover` (a function called by this functions) acts only on an `AbstractNamedGraph` ?

JoeyT1994: So this would raise an error if you passed it an undirected `datagraph` or something

src/beliefpropagation/beliefpropagation_schedule.jl
JoeyT1994: Also should we really allow the trait function to dispatch on any type which is not directed or should we at least constraint it to abstractgraphs via
`@traitfn default_bp_niters(g::AbstractGraph::(!IsDirected)) = is_tree(g) ? 1 : nothing`

JoeyT1994: What about

```
@traitfn undirected_default_bp_niters(g::AbstractGraph::(!IsDirected)) =
  is_tree(g) ? 1 : nothing
function default_bp_niters(g::AbstractGraph)
  return undirected_default_bp_niters(undirected_graph(underlying_graph(g)))
end
```

src/beliefpropagation/beliefpropagation.jl
mtfishman: Looks like this was resolved but I don't see the type assertion removed.

PR #114, opened 2023-11-03
Fix some issues around keyword arguments

mtfishman:


mtfishman:
@JoeyT1994 this fixes the test errors caused by stricter keyword argument handling introduced in https://github.com/ITensor/ITensors.jl/pull/1226.

PR #118, opened 2024-01-10
Fix test failures

mtfishman:
This is an attempt to fix test failures.


mtfishman:
This fixes tests on Julia 1.10 but there is still a mysterious issue on Julia 1.7, I'll merge we can investigate that later.

PR #116, opened 2024-01-08
[Enhancement] Add quantum number support in TTNO constructor.

b-kloss:
This PR generalizes the existing constructor of a `TTN` from an `OpSum` and `IndsNetwork` (`svdTTN`) to physical indices with `QN`s (`qn_svdTTN`). This implementation closely follows the parallel logic in `ITensors.svdMPO` and `ITensors.qn_svdMPO`. Tests of the functionality (at the same level as tests of the existing constructor) have been added.

ToDo:
 

- [x]  test/generalize for TTN's with vertices without physical indices
- [ ] implement autoFermion support
- [x] test whether QN code can be used also for non-QN siteinds with the goal to remove non-QN code
- [ ] implement profiling/instrumentation for future performance optimizations 
- [x] add function barrier for type stability
- [x] change to snaking case
- [x] move non-QN functionality into a `deprecated.jl` file

PR #119, opened 2024-01-10
Use `ITensorNetworks.heisenberg` in PEPS example

mtfishman:
Use `ITensorNetworks.heisenberg` in PEPS example.


mtfishman:
Hopefully helps a test issue that popped up in #116.

PR #120, opened 2024-01-10
Contraction path optimization with EinExprs

mofeing:
This is an ongoing work with @JoeyT1994 to add support for contraction path optimization using [EinExprs.jl](https://github.com/bsc-quantic/EinExprs.jl). It is a lightweight package ONLY for contraction path optimization. Inspired by `opt_einsum` and `cotengra` Python packages, but with more effort spent on simplicity and performance (algorithms occupy 10s of lines and are easy to comprehend).

- [x] Fix lingering Julia 1.10 test errors on Ubuntu and macOS (package extension is loading properly but there are some bugs in the code).
- [x] Fix Windows tests by making `KaHyPar.jl` an optional dependency of `EinExprs.jl`, since it is not available for Windows so the extension fails to load since it is looking for a library that doesn't exist.
- [x] Add [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl/tree/main) to make sure the extension will work for Julia versions below 1.9.

~`EinExprs` is not yet in the General registry, so the @bsc-quantic registry has to be added. We are working towards moving it to the General registry for the next release.~
`EinExprs` has been moved to the General registry.

~This is a WIP, not yet ready to be merged.~

@JoeyT1994 I have a couple of questions:
- EinExprs represents indices using `Symbol`s. Since an `Index` has no name, it is ok if I use the `id`?
- How can I obtain the open indices in a `ITensorNetwork`?
- What is the expected return type of `contraction_sequence`?


JoeyT1994:
Thanks for opening this!

- Yes, the `id` is a unique identifier for each index so this is a good way to keep track of them. 

- The function `externalinds(tn)` returns a vector of all of the open indices in an `ITensorNetwork`. Although definitely let us know if you see any cases where it doesn't return what you expect.

- `contraction_sequence(tn)` returns a nested list. In the case of a vector `[A,B,C]` of `ITensors`, `contraction_sequence([A,B,C])` might return `[3, [1,2]]` telling you that you should contract tensors `1` and `2` (`A` and `B` in the list) first and then contract the result onto the third tensor in the list (`C`).  In the case of `tn`, an `ITensorNetwork`, the return is similar but the entries of the nested list correspond to the vertices `v` of the `ITensorNetwork`. These are wrapped as `Key(v)` to avoid ambiguity when you have vertex names which are lists themselves.

mtfishman:
Thanks @mofeing, seems like this will be good to have. Currently we use [OMEinsumContractionOrders.jl](https://github.com/TensorBFS/OMEinsumContractionOrders.jl) as a backend, how does that compare to `EinExprs.jl`?

Adding on to Joey's comments:

1. I think this should be designed as a package extension so it is in optional dependency.
2. ITensor indices are also identified by their tags and prime levels along with their ID. One strategy to handle that could be to make a dictionary mapping each ITensor Index in the network to a symbol that can be used by `EinExprs.jl`. The hash of an ITensor Index is based on the id, tags, and prime level.

mtfishman:
The only reason we don't use package extensions right now for `OMEinsumContractionOrders` is because we added that dependency before they were available and haven't updated it yet. Ideally for that and this PR we would organize any conditionally loaded code as package extensions and use [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl) for backwards compatibility with older Julia versions.

Yeah, ITensor indices can have the same ID but if they have different prime levels or tags they won't compare equal. Distinguishing indices by either IDs, prime levels, or tags can be convenient in different situations (for example we have a convention that operators distinguish bra and ket spaces using prime levels, which is much more convenient than distinguishing them by IDs which are randomly generated).

mofeing:
ah ok, then should I use `hash(index)`?

mtfishman:
`hash` alone wouldn't be guaranteed to not give collisions (i.e. return the same hash value even if the indices are different), but using a dictionary would help handle hash collisions.

mofeing:
mmm then the `objectid` of the index?

mtfishman:
Maybe, though it seems like there can still be hash collisions: https://docs.julialang.org/en/v1/base/base/#Base.objectid.

mtfishman:
I'm picturing something like this:
```julia
julia> using ITensors

julia> i, j, k = Index.((2, 2, 2))
((dim=2|id=45), (dim=2|id=388), (dim=2|id=489))

julia> symbols = Dict{Index,Symbol}()
Dict{Index, Symbol}()

julia> using Random

julia> get!(symbols, i, Symbol(randstring()))
:maCJxbai

julia> get!(symbols, j, Symbol(randstring()))
:bfRnBZjQ

julia> get!(symbols, k, Symbol(randstring()))
:wiM0i2Oe

julia> get!(symbols, i, Symbol(randstring()))
:maCJxbai

julia> symbols
Dict{Index, Symbol} with 3 entries:
  (dim=2|id=489) => :wiM0i2Oe
  (dim=2|id=388) => :bfRnBZjQ
  (dim=2|id=45)  => :maCJxbai
```
Another option is to make a unique vector of all ITensor indices (say with `unique`) and then use symbols that correspond to each index deterministically based on the position in the vector, i.e.:
```julia
julia> inds = unique([i, j, k, k])
3-element Vector{Index{Int64}}:
 (dim=2|id=45)
 (dim=2|id=388)
 (dim=2|id=489)

julia> symbols = Symbol.(eachindex(inds))
3-element Vector{Symbol}:
 Symbol("1")
 Symbol("2")
 Symbol("3")

julia> inds[parse(Int, String(symbols[2]))]
(dim=2|id=388)
```

mofeing:
okay, will try to refactor base on your examples. what should I use to get the full list of indices of the TN?

JoeyT1994:
`collect(vertex_data(siteinds(ψ)))` will give you the open indices.
`collect(edge_data(linkinds(ψ)))` will give you the internal indices. 
So the concatenation of those two lists should give you them all. @mtfishman maybe there is a function to do this explicitly but I can't see it.

mtfishman:
@JoeyT1994 does that make a flat list of indices (`Vector{<:Index}`), or a nested list `Vector{<:Vector{<:Index}}`? I would have guessed that latter, you probably also need to flatten it with something like `collect(Iterators.flatten(v))` or `reduce(vcat, v)`.

I think that is probably the most succinct way for now.

mtfishman:
Though the definition of `externalinds` doesn't look correct to me, I think it should be looping over the vertices instead of the edges.

JoeyT1994:
@mtfishman #123 should hopefully fix that and export those functions?

mofeing:
I parametrized the label type in `EinExprs` v0.6 so now you can use your own `Index` type. I've updated the code. Can you give it a check @JoeyT1994?

ext/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,37 @@
...
+  optimizer=EinExprs.Exhaustive(),
+) where {T}
+  IndexType = Any
+
+  tensors = EinExpr{IndexType}[]
+  tensor_map = Dict{Set{IndexType},T}()
+  sizedict = Dict{IndexType,Int}()
+
+  for (key, tensor) in pairs(vertex_data(tn))
+    _inds = collect(ITensorNetworks.inds(tensor))
```
JoeyT1994, line 19: I think `inds()` should just work as it is exported by `ITensors`.

ext/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,37 @@
...
+    push!(tensors, EinExpr{IndexType}(; head=_inds))
+    tensor_map[Set(_inds)] = key
+    merge!(sizedict, Dict(_inds .=> size(tensor)))
+  end
+
+  _openinds = collect(ITensorNetworks.externalinds(tn))
```
JoeyT1994, line 25: `ITensorNetworks.externalinds() -> externalinds()`

JoeyT1994 (commented):
Thanks @mofeing ! I think we could follow the pattern we have for `OMEinsumContractionOrders` here and put this code in a file `src/requires/EinExprs.jl` 

Then you can write a function in `src/contraction_sequences` which checks the `EinExprs.jl` module exist and then passes to this function ? 

That way we don't need `EinExprs.jl` in the dependencies at all and can just add it to `src/test/Project.toml` as a required dependency for testing.

ext/ITensorNetworksEinExprsExt.jl
mofeing: The problem is that `EinExprs` also exports `inds`, so I put it to make sure which `inds` function it's using.

ext/ITensorNetworksEinExprsExt.jl
mofeing: I don't think it is automatically exported.
```julia
julia> using ITensorNetworks

julia> externalinds
ERROR: UndefVarError: `externalinds` not defined
```

mofeing:
> Thanks @mofeing ! I think we could follow the pattern we have for `OMEinsumContractionOrders` here and put this code in a file `src/requires/EinExprs.jl`
> 
> Then you can write a function in `src/contraction_sequences` which checks the `EinExprs.jl` module exist and then passes to this function ?
> 
> That way we don't need `EinExprs.jl` in the dependencies at all and can just add it to `src/test/Project.toml` as a required dependency for testing.

From https://github.com/mtfishman/ITensorNetworks.jl/pull/120#issuecomment-1887602761, I understood you wanted it as a package extension. Currently it's set as a weakdep.

ext/ITensorNetworksEinExprsExt.jl
JoeyT1994: Strange, are you on `ITensorNetworks 0.4.0`? This should have been added with #123 .

ext/ITensorNetworksEinExprsExt.jl
JoeyT1994: `tensor` here is presumably an `ITensor` type? Is `EinExprs.inds()` defined for an `ITensor` too? If so then I understand keeping `ITensorNetworks.inds`

JoeyT1994:
Ah yes good point. @mtfishman what do you think? Shall we have this as a package extension as it is currently? And then in a future PR also switch `OMEEinsum` to a package extension like @mofeing has done.

ext/ITensorNetworksEinExprsExt.jl
mofeing: Nope, I'm on v0.3.10. Let me rebase then.

ext/ITensorNetworksEinExprsExt.jl
mofeing: Yes, here `tensor` is an `ITensor`.
Nope, `EinExprs.inds` is not defined for `ITensor`.

mofeing:
Update: `EinExprs` is getting registered in the General registry. The 3-day countdown finishes tomorrow, so I'll update tests then. If not, the CI will fail because it won't find `EinExprs`.

mtfishman:
This should be a package extension, and there should be a package extension for `OMEinsumContractionOrders` as well (we can do that in a future PR).

mtfishman:
@mofeing I think we should put the code for this into a directory structured like `ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl`, to anticipate having other extensions that may have more sophisticated file structures.

ext/ITensorNetworksEinExprsExt.jl
JoeyT1994: I see, `ITensors` exports `inds` so shouldn't `inds(it::ITensor)` just work without the prefix as there is no ambiguity with `EinExprs`?

ext/ITensorNetworksEinExprsExt.jl
JoeyT1994, line 1: Can we create the folder `ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl` for this? I know it's currently just one file but I think we want to anticipate having other `ext` dependencies with large file structures. So this gives us a good idea for how that should be structured.

JoeyT1994 (commented):
Looks great @mofeing ! What do you think @mtfishman ? Once it is merged we can do some testing of the library on the runtime for finding sequences via the various methods. That might help us decide what we want to use as default etc

ext/ITensorNetworksEinExprsExt.jl
mofeing: ah right, I forget about this

mofeing: done

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
+module ITensorNetworksEinExprsExt
+
+using ITensors
+using ITensorNetworks
+using ITensorNetworks: Index
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function EinExprs.einexpr(tn::ITensorNetwork{T}; optimizer::EinExprs.Optimizer) where {T}
```
mtfishman, line 8:
```suggestion
function EinExprs.einexpr(tn::ITensorNetwork; optimizer::EinExprs.Optimizer)
```
and then use `vertextype(tn)` instead of `T` in the body of the function. Generally I try to avoid using `where` when possible since it can usually be replaced by function calls which make code more generic, though we haven't moved all of the code in `ITensorNetworks` over to that style yet.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
+module ITensorNetworksEinExprsExt
+
+using ITensors
+using ITensorNetworks
+using ITensorNetworks: Index
```
mtfishman, line 5: Let's list the functions and types we are using explicitly, i.e.:
```julia
using ITensors: Index, ...
using ITensorNetworks: ITensorNetworks, ITensorNetwork, ...
```
Also best to include a function or type from the module/package where it originated, i.e. `using ITensors: Index` instead of `using ITensorNetworks: Index`.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,66 @@
...
+  alg::ITensorNetworks.Algorithm"einexpr", tn::Vector{ITensor}; kwargs...
+)
+  return contraction_sequence(alg, ITensorNetwork(tn); kwargs...)
+end
+
+function ITensorNetworks.contraction_sequence(
+  ::ITensorNetworks.Algorithm"einexpr",
+  tn::ITensorNetwork{T};
+  optimizer=EinExprs.Exhaustive(),
+) where {T}
```
mtfishman, line 39: Same comment on `where` as above.

mofeing:
Btw, if there is something that you would like to have in `EinExprs`, we are open to suggestions.

I've added the `EinExprs.einexpr` method to `ITensorNetwork` so you can get back the `EinExpr` and plot it, or compute some indices to slice...

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,66 @@
...
+
+  _openinds = collect(externalinds(tn))
+  expr = SizedEinExpr(sum(tensors; skip=_openinds), sizedict)
+
+  path = einexpr(optimizer, expr)
+
+  function _convert_to_contraction_sequence(subpath)
+    EinExprs.nargs(subpath) == 0 && return tensor_map[Set(subpath.head)]
+    return map(_convert_to_contraction_sequence, EinExprs.args(subpath))
+  end
```
mtfishman, line 61: I'm confused why this needs to be a function.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mtfishman: Also, it seems like you may as well include more things into the namespace, like `ITensors.Algorithm`, `EinExprs.Optimizer`, etc. I wasn't clear on the logic of why certain things were included into the namespace while others were not.

mtfishman: Unless they are indicating that they are low-level or private parts of the API, perhaps that is the case with `EinExprs.nargs`? If that's the case it seems worth prepending with the module name to remind us that the code is relying on internals of `EinExprs.jl`.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,67 @@
+module ITensorNetworksEinExprsExt
+
+using ITensors
+using ITensorNetworks
+using ITensorNetworks: Index
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function EinExprs.einexpr(tn::ITensorNetwork; optimizer::EinExprs.Optimizer)
+  IndexType = Any
+  T = vertextype(tn)
```
mtfishman, line 16: Maybe `VertexType`?

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: Because it is recursive: it calls itself on line 60.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,67 @@
...
+
+  tensors = EinExpr{IndexType}[]
+  tensor_map = Dict{Set{IndexType},T}()
+  sizedict = Dict{IndexType,Int}()
+
+  for (key, tensor) in pairs(vertex_data(tn))
```
mtfishman, line 22: I think it is better to write this as:
```julia
for vertex in vertices(tn)
  tensor = tn[vertex]
  ...
end
```
since `vertex_data` should be considered an internal detail of the `ITensorNetwork` type.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,67 @@
...
+    _inds = collect(inds(tensor))
```
mtfishman, line 22:
```suggestion
    inds_v = collect(inds(tensor))
```
to indicate it is the indices of the tensor on vertex `v`. I don't really like using variable names that start with `_`.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,67 @@
...
+using ITensors
+using ITensorNetworks
+using ITensorNetworks: Index
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function EinExprs.einexpr(tn::ITensorNetwork; optimizer::EinExprs.Optimizer)
+  IndexType = Any
+  T = vertextype(tn)
+
+  tensors = EinExpr{IndexType}[]
```
mtfishman, line 17: I think it is confusing using the variable name `tensor` below to refer to an ITensor, but then `tensors` here is a collection of `EinExpr`. Maybe this could be called `einexprs`?

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: `EinExprs.nargs` is "public" but not exported by default. I didn't find a use-case outside of these integrations, but it's solid and there shouldn't be any changes.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,67 @@
...
+  sizedict = Dict{IndexType,Int}()
+
+  for (key, tensor) in pairs(vertex_data(tn))
+    _inds = collect(inds(tensor))
+    push!(tensors, EinExpr{IndexType}(; head=_inds))
+    tensor_map[Set(_inds)] = key
+    merge!(sizedict, Dict(_inds .=> size(tensor)))
+  end
+
+  _openinds = collect(externalinds(tn))
```
mtfishman, line 28:
```suggestion
  externalinds_tn = collect(externalinds(tn))
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,67 @@
...
+  return contraction_sequence(alg, ITensorNetwork(tn); kwargs...)
+end
+
+function ITensorNetworks.contraction_sequence(
+  ::ITensorNetworks.Algorithm"einexpr",
+  tn::ITensorNetwork{T};
+  optimizer=EinExprs.Exhaustive(),
+)
+  IndexType = Any
+  T = vertextype(tn)
```
mtfishman:
```suggestion
  VertexType = vertextype(tn)
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,67 @@
...
+
+  tensors = EinExpr{IndexType}[]
+  tensor_map = Dict{Set{IndexType},T}()
+  sizedict = Dict{IndexType,Int}()
+
+  for (key, tensor) in pairs(vertex_data(tn))
```
mtfishman, line 50: Same as above in terms of naming variables and iterating over vertices and tensors.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: Same thing with `EinExprs.Optimizer`: it's public but not exported.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,70 @@
...
+using ITensorNetworks:
+  ITensorNetworks,
+  ITensorNetwork,
+  vertextype,
+  vertex_data,
+  externalinds,
+  contraction_sequence
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function EinExprs.einexpr(tn::ITensorNetwork; optimizer::EinExprs.Optimizer)
```
mtfishman, line 13: It seems like the code for `EinExprs.einexpr` and `ITensorNetworks.contraction_sequence` are pretty similar, is there some common code logic we can split off and share between them?

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: I usually just call them `exprs`.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mtfishman: Oh, I see. Maybe we could split this off into a function `to_contraction_sequence(tensor_map, subpath)`? I usually prefer not defining functions within other functions like this since generally they can be factored out into separate functions that may be useful in other places with a bit of thought.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: I wanted `ITensorNetworks.contraction_sequence` to call `EinExprs.einexpr` but then duplicated them because `contraction_sequence` required `tensor_map`. Now I've refactored the common code to `prepare_einexpr` function.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: Renamed them to `tensor_exprs`.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,59 @@
...
+  externalinds,
+  contraction_sequence
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function prepare_einexpr(tn::ITensorNetwork)
+  IndexType = Any
+  VertexType = vertextype(tn)
+
+  tensors = EinExpr{IndexType}[]
+  tensor_map = Dict{Set{IndexType},VertexType}()
```
mtfishman, line 18: Could we come up with a better name for this mapping? It required me to read to type signature to understand what it might represent. Maybe `tensor_inds_to_vertex` or `inds_to_vertex` (I guess it is identifying which vertex a tensor is on based on the indices?).

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: Yep, no problem.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,59 @@
...
+function ITensorNetworks.contraction_sequence(
+  alg::Algorithm"einexpr", tn::Vector{ITensor}; kwargs...
+)
+  return contraction_sequence(alg, ITensorNetwork(tn); kwargs...)
+end
+
+function ITensorNetworks.contraction_sequence(
+  ::Algorithm"einexpr", tn::ITensorNetwork{T}; optimizer=EinExprs.Exhaustive()
+)
+  path, tensor_map = prepare_einexpr(tn)
```
mtfishman, line 49:
```suggestion
  expr, tensor_inds_to_vertex = prepare_einexpr(tn)
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,58 @@
...
+using ITensorNetworks:
+  ITensorNetworks,
+  ITensorNetwork,
+  vertextype,
+  vertex_data,
+  externalinds,
+  contraction_sequence
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function prepare_einexpr(tn::ITensorNetwork)
```
mtfishman, line 13: Would prefer `einexpr_and_inds_to_vertex` or whatever name we decide to give the current `tensor_map` data structure to be clearer about what this is outputting, since `prepare` is pretty vague.

mofeing: What do you think of `to_einexpr`?

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,58 @@
...
+function ITensorNetworks.contraction_sequence(
+  alg::Algorithm"einexpr", tn::Vector{ITensor}; kwargs...
+)
+  return contraction_sequence(alg, ITensorNetwork(tn); kwargs...)
+end
+
+function ITensorNetworks.contraction_sequence(
+  ::Algorithm"einexpr", tn::ITensorNetwork{T}; optimizer=EinExprs.Exhaustive()
+)
+  path, tensor_inds_to_vertex = prepare_einexpr(tn)
```
mtfishman, line 49:
```suggestion
  expr, tensor_inds_to_vertex = prepare_einexpr(tn)
```
since it seems like that's what this object is called in other functions. Same change in `to_contraction_sequence`.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,58 @@
...
+  return to_contraction_sequence(tensor_inds_to_vertex, path)
+end
+
+function to_contraction_sequence(tensor_inds_to_vertex, path)
```
mtfishman, line 53:
```suggestion
function to_contraction_sequence(expr, tensor_inds_to_vertex)
```
That ordering makes more sense to me, since this function is converting the `expr`. The `tensor_inds_to_vertex` is just used to help with that conversion.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mtfishman: For this I would still prefer a longer more explicit name saying what it is outputting, since it is being used internally so the priority would be on readability/understanding the code later on rather than having a short name.

mofeing:
Tests on Julia 1.7 seem to be failing because `EinExprs` minimum supported version is Julia 1.9.

mtfishman:
I think there are a few issues on Julia 1.7:

1. `ITensorNetworks` tests were already failing on Julia 1.7 for some unknown reason, though ideally we would support it.
2. Package extensions were only introduced in Julia 1.9: https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions), we can use [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl) to make it work on earlier versions.
3. `EinExprs.jl` only supporting Julia 1.9 and later.

We can investigate issue 1. separately from this PR. Could you address issue 2. in this PR?

Regarding 3., why doesn't `EinExprs.jl` support earlier Julia versions?

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,58 @@
...
+function EinExprs.einexpr(tn::ITensorNetwork; optimizer::EinExprs.Optimizer)
+  expr, _ = to_einexpr(tn)
+  return einexpr(optimizer, expr)
+end
+
+function ITensorNetworks.contraction_sequence(
+  alg::Algorithm"einexpr", tn::Vector{ITensor}; kwargs...
+)
+  return contraction_sequence(alg, ITensorNetwork(tn); kwargs...)
+end
```
mtfishman, line 44: It seems like this should already be defined generically for any `Algorithm` input, is that not the case? If not this should be added to the main package.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,58 @@
...
+function ITensorNetworks.contraction_sequence(
+  ::Algorithm"einexpr", tn::ITensorNetwork{T}; optimizer=EinExprs.Exhaustive()
+)
+  expr, tensor_inds_to_vertex = to_einexpr(tn)
+  return to_contraction_sequence(expr, tensor_inds_to_vertex)
+end
+
+function to_contraction_sequence(expr, tensor_inds_to_vertex)
+  EinExprs.nargs(expr) == 0 && return tensor_inds_to_vertex[Set(expr.head)]
+  return map(_convert_to_contraction_sequence, EinExprs.args(expr))
```
mtfishman, line 55: Seems like this should use `to_contraction_sequence` instead of `_convert_to_contraction_sequence`.

mtfishman:
Regarding using the `Index` directly in the `EinExpr`, it's definitely nice that `EinExpr` can now handle generic index labels. A potential issue using `Index` is that comparison between `Index` objects is slower than comparison between simpler objects like `Int` or `Symbol`. If contraction sequence finding algorithms in `EinExprs.jl` is dominated by index comparisons you may see a performance hit there so you may want to define a map to and from a simple index label object to speed things up. But that seems like something that could be handled either on the `EinExprs.jl` side or in another PR, if that is seen to be necessary for performance.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: 🤦 my bad

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: I think that is not the case. What I've seen is that `contraction_sequence` is called with a `Vector{ITensor}` by default.

JoeyT1994: Perhaps we should make this in line with the other functions in `contraction_sequences.jl` to resolve it. Specifically, we currently have the logic:
```
function contraction_sequence(tn::AbstractITensorNetwork; kwargs...)
  seq_linear_index = contraction_sequence(Vector{ITensor}(tn); kwargs...)
  # TODO: Use Functors.fmap?
  return deepmap(n -> Key(vertices(tn)[n]), seq_linear_index)
end
```
So here really we should have `to_einexpr(tn::ITensorNetwork)` and `contraction_sequence(
  ::Algorithm"einexpr", tn::ITensorNetwork{T}; kwargs..)` take a `Vector{ITensor}` as the argument instead? 

I think the logic of `to_einexpr` would be almost unchanged, except the call `externalinds(tn)` should be replaced by `noncommondinds(ts)` where `ts::Vector{ITensor}`?

mtfishman: Ah I understand. This is the standard logic for other `contraction_sequence` backends:
https://github.com/mtfishman/ITensorNetworks.jl/blob/3fc272e7ff7bc1650b76695d2bf883ef7ef7d618/src/contraction_sequences.jl#L5-L9
i.e. it converts from `ITensorNetwork` to `Vector{ITensor}`, gets a sequence in terms of integer vertex labels, then converts that sequence back to the vertex labels/names of the `ITensorNetwork` (wrapped in the `Key` type to disambiguate the case where the vertices themselves are of type `Vector`).

It seems like we could in principle do that here, though the current code makes use of some `ITensorNetwork` functions like `externalinds`. We could just keep that in mind for the future.

mtfishman: Just saw your post @JoeyT1994. Seems reasonable to rewrite these functions in terms of `Vector{ITensor}` instead in this PR.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,58 @@
...
+function ITensorNetworks.contraction_sequence(
+  ::Algorithm"einexpr", tn::ITensorNetwork{T}; optimizer=EinExprs.Exhaustive()
+)
+  expr, tensor_inds_to_vertex = to_einexpr(tn)
+  return to_contraction_sequence(expr, tensor_inds_to_vertex)
+end
+
+function to_contraction_sequence(expr, tensor_inds_to_vertex)
+  EinExprs.nargs(expr) == 0 && return tensor_inds_to_vertex[Set(expr.head)]
+  return map(to_contraction_sequence, EinExprs.args(expr))
```
mtfishman, line 55: Probably should be:
```suggestion
  return map(expr -> to_contraction_sequence(expr, tensor_inds_to_vertex), EinExprs.args(expr))
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,58 @@
...
+    push!(tensor_exprs, EinExpr{IndexType}(; head=inds_v))
+    tensor_inds_to_vertex[Set(inds_v)] = key
+    merge!(inds_dims, Dict(inds_v .=> size(tensor_v)))
+  end
+
+  externalinds_tn = collect(externalinds(tn))
+  expr = SizedEinExpr(sum(tensor_exprs; skip=externalinds_tn), inds_dims)
+
+  return expr, tensor_inds_to_vertex
+end
```
mtfishman, line 33: I realized that there isn't really a need for this one function to make both `expr` and `tensor_inds_to_vertex`, can we just split it in two? The part building `expr` could be called `expr(tn)` (or whatever seems appropriate for that) and the part building `tensor_inds_to_vertex` could be it's own function called `tensor_inds_to_vertex(tn)`. I think that would help settle the question of what to call it.

mofeing: Done!

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: should i refactor it to work with `Vector{ITensor}` or is it ok if i just transform to an `ITensorNetwork`?

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
...
+using ITensorNetworks:
+  ITensorNetworks,
+  ITensorNetwork,
+  vertextype,
+  vertex_data,
+  externalinds,
+  contraction_sequence
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function to_einexpr(tn::AbstractITensorNetwork)
```
JoeyT1994, line 13:
```suggestion
function to_einexpr(ts::Vector{ITensor})
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
...
+  inds_dims = Dict{IndexType,Int}()
+
+  for v in vertices(tn)
+    tensor_v = tn[v]
+    inds_v = collect(inds(tensor_v))
+    push!(tensor_exprs, EinExpr{IndexType}(; head=inds_v))
+    merge!(inds_dims, Dict(inds_v .=> size(tensor_v)))
+  end
+
+  externalinds_tn = collect(externalinds(tn))
```
JoeyT1994, line 26:
```suggestion
  externalinds_ts = noncommoninds(ts...)
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
...
+  contraction_sequence
+using EinExprs: EinExprs, EinExpr, einexpr, SizedEinExpr
+
+function to_einexpr(tn::AbstractITensorNetwork)
+  IndexType = Any
+
+  tensor_exprs = EinExpr{IndexType}[]
+  inds_dims = Dict{IndexType,Int}()
+
+  for v in vertices(tn)
```
JoeyT1994, line 19:
```suggestion
  for tensor_v in ts
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
...
+    tensor_v = tn[v]
```
JoeyT1994, line 20:
```suggestion
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
...
+    inds_v = collect(inds(tensor_v))
+    push!(tensor_exprs, EinExpr{IndexType}(; head=inds_v))
+    merge!(inds_dims, Dict(inds_v .=> size(tensor_v)))
+  end
+
+  externalinds_tn = collect(externalinds(tn))
+  return SizedEinExpr(sum(tensor_exprs; skip=externalinds_tn), inds_dims)
```
JoeyT1994, line 27:
```suggestion
  return SizedEinExpr(sum(tensor_exprs; skip=externalinds_ts), inds_dims)
```

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
JoeyT1994: I just made suggestions on `to_einexpr()` for how to change it to take `Vector{ITensor}` instead. The same logic can be done with `tensor_inds_to_vertex` and then you should be able to get rid of the `contraction_sequence` definitions in this file.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,65 @@
...
+    mapping[Set(inds_v)] = v
+  end
+
+  return mapping
+end
+
+function EinExprs.einexpr(tn::ITensorNetwork; optimizer::EinExprs.Optimizer)
+  expr = to_einexpr(tn)
+  return einexpr(optimizer, expr)
+end
```
JoeyT1994, line 45: Is this function necessary? Or can we just combine it directly with `einexpr()` above?

mofeing: Now that we have `to_einexpr`, I think it's no longer needed.

mtfishman:
`ITensors.jl` supports Julia 1.6, since that is the current long term support version of Julia. I've found that with a combination of [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl) and [Compat.jl](https://github.com/JuliaLang/Compat.jl), it is pretty easy to support back to Julia 1.6 but still use modern functionality.

The goal would be for `ITensorNetworks.jl` to support Julia 1.6, currently we only test back to Julia 1.7 but that is mysteriously broken so we'll have to look into that.

test/test_contraction_sequence.jl
```diff
@@ -23,7 +24,19 @@ ITensors.disable_warn_order()
...
+  res_einexprs_greedy = contract(tn; sequence=seq_einexprs_exhaustive)[]
+  seq_einexprs_kahypar = contraction_sequence(tn; alg="einexpr", optimizer=HyPar())
+  res_einexprs_kahypar = contract(tn; sequence=seq_einexprs_kahypar)[]
+  @test res_optimal ≈
+    res_greedy ≈
+    res_tree_sa ≈
+    res_sa_bipartite ≈
+    res_einexprs_exhaustive ≈
+    res_einexprs_greedy ≈
+    res_einexprs_kahypar
```
mtfishman, line 39: I think it is better to split into separate tests, like:
```julia
  @test res_optimal ≈ res_greedy
  @test res_optimal ≈ res_tree_sa
  #...
```
The way it is written, if one failed the whole test would fail but we wouldn't know which one caused the failure.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,50 @@
...
+  tensor_exprs = EinExpr{IndexType}[]
+  inds_dims = Dict{IndexType,Int}()
+
+  for tensor_v in ts
+    inds_v = collect(inds(tensor_v))
+    push!(tensor_exprs, EinExpr{IndexType}(; head=inds_v))
+    merge!(inds_dims, Dict(inds_v .=> size(tensor_v)))
+  end
+
+  externalinds_tn = collect(noncommoninds(ts...))
```
mtfishman, line 20:
```suggestion
  externalinds_tn = reduce(noncommoninds, ts)
```
Splatting a big vector can cause trouble for the compiler.

mofeing:
Okay. Let me add these packages to `EinExprs` and release a new patch version.

mofeing:
EinExprs v0.6.3 released!

mtfishman:
Great, thanks! Looks like tests are still failing, it seems like the new package extension isn't being loaded properly, in the Julia 1.10 test logs I see:
```julia
┌ Error: Error during loading of extension ITensorNetworksEinExprsExt of ITensorNetworks, use `Base.retry_load_extensions()` to retry.
│   exception =
│    1-element ExceptionStack:
│    ArgumentError: Package ITensorNetworksEinExprsExt [2789bce2-5c52-55b3-96f1-0f8e29cdd121] is required but does not seem to be installed:
│     - Run `Pkg.instantiate()` to install all recorded dependencies.
│    
│    Stacktrace:
│      [1] _require(pkg::Base.PkgId, env::Nothing)
│        @ Base ./loading.jl:1920
│      [2] __require_prelocked(uuidkey::Base.PkgId, env::Nothing)
│        @ Base ./loading.jl:1806
│      [3] #invoke_in_world#3
│        @ Base ./essentials.jl:921 [inlined]
...
```

mofeing:
I think we need to replicate weakdeps into either deps or extras.

mofeing:
mmm could it be that KaHyPar is not built for Windows? I have `KaHyPar` as a strong dependency, but I can move to weakdeps if needed.

mtfishman:
Yes, KaHyPar doesn't have binaries for Windows: https://github.com/kahypar/KaHyPar.jl/issues/26, which is consistent with the Windows CI error message:
```julia
ERROR: LoadError: UndefVarError: `libkahypar` not defined
Stacktrace:
 [1] include
   @ Base .\Base.jl:495 [inlined]
 [2] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
   @ Base .\loading.jl:2216
 [3] top-level scope
   @ stdin:3
in expression starting at C:\Users\runneradmin\.julia\packages\KaHyPar\63UEh\src\KaHyPar.jl:1
in expression starting at stdin:3
ERROR: LoadError: Failed to precompile KaHyPar [2a6221f6-aa48-11e9-3542-2d9e0ef01880] to "C:\\Users\\runneradmin\\.julia\\compiled\\v1.10\\KaHyPar\\jl_5150.tmp".
#...
```

However, the macOS and Ubuntu CI are also failing, for different reasons:
```julia
┌ Error: Error during loading of extension ITensorNetworksEinExprsExt of ITensorNetworks, use `Base.retry_load_extensions()` to retry.
│   exception =
│    1-element ExceptionStack:
│    ArgumentError: Package ITensorNetworksEinExprsExt [2789bce2-5c52-55b3-96f1-0f8e29cdd121] is required but does not seem to be installed:
│     - Run `Pkg.instantiate()` to install all recorded dependencies.
│    
│    Stacktrace:
│      [1] _require(pkg::Base.PkgId, env::Nothing)
│        @ Base ./loading.jl:1920
│      [2] __require_prelocked(uuidkey::Base.PkgId, env::Nothing)
│        @ Base ./loading.jl:1806
│      [3] #invoke_in_world#3
│        @ Base ./essentials.jl:921 [inlined]
│      [4] invoke_in_world
│        @ Base ./essentials.jl:918 [inlined]
#...
```
I'm not sure what's going on there, I don't see anything obviously wrong with the setup. Do tests pass if you run them locally?

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mtfishman, line 1: Maybe this needs to be `ext/ITensorNetworksEinExprsExt/ITensorNetworksEinExprsExt.jl` instead of `ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl`, according to https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions).

mtfishman:
Reminder to add [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl/tree/main) to make sure the extension will work for Julia versions below 1.9.

ext/ITensorNetworksEinExprsExt/src/ITensorNetworksEinExprsExt.jl
mofeing: I guess this is the case because tests pass locally.

ext/ITensorNetworksEinExprsExt/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,50 @@
+module ITensorNetworksEinExprsExt
+
+using ITensors: Index, ITensor, Algorithm, noncommoninds
```
mtfishman, line 3:
```suggestion
using ITensors: Index, ITensor, Algorithm, @Algorithm_str, noncommoninds
```
should fix the latest CI failures.

mtfishman:
The error that CI is throwing now is:
```julia
ERROR: LoadError: UndefVarError: `@Algorithm_str` not defined
Stacktrace:
 [1] top-level scope
   @ :0
 [2] include
   @ Base ./Base.jl:495 [inlined]
 [3] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
   @ Base ./loading.jl:2216
 [4] top-level scope
   @ stdin:3
in expression starting at /home/runner/work/ITensorNetworks.jl/ITensorNetworks.jl/ext/ITensorNetworksEinExprsExt/ITensorNetworksEinExprsExt.jl:36
in expression starting at /home/runner/work/ITensorNetworks.jl/ITensorNetworks.jl/ext/ITensorNetworksEinExprsExt/ITensorNetworksEinExprsExt.jl:1
in expression starting at stdin:3
#...
```
since `@Algorithm_str` isn't being loaded into the namespace, I committed an attempted fix.

ext/ITensorNetworksEinExprsExt/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,50 @@
+module ITensorNetworksEinExprsExt
+
+using ITensors: Index, ITensor, Algorithm, @Algorithm_str, noncommoninds
```
mtfishman, line 3:
```suggestion
using ITensors: Index, ITensor, @Algorithm_str, noncommoninds
```
Not needed, doesn't appear to be used.

ext/ITensorNetworksEinExprsExt/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,50 @@
+module ITensorNetworksEinExprsExt
+
+using ITensors: Index, ITensor, @Algorithm_str, noncommoninds
```
mtfishman, line 3:
```suggestion
using ITensors: Index, ITensor, @Algorithm_str, inds, noncommoninds
```
Fix another namespace issue.

mtfishman:
@mofeing at least now `ITensorNetworksEinExprsExt`  is getting loaded properly on Ubuntu and macOS but there are some bugs introduced by the refactor, could you take a look?

mofeing:
I moved KaHyPar to a package extension and bumped the patch version.
https://github.com/JuliaRegistries/General/pull/100771

mtfishman:
@mofeing still more test failures due to bugs caused by the refactor.

mtfishman:
> I moved KaHyPar to a package extension and bumped the patch version. [JuliaRegistries/General#100771](https://github.com/JuliaRegistries/General/pull/100771)

Great, thanks. I think that is a good move.

Project.toml
```diff
@@ -40,6 +46,7 @@ DataStructures = "0.18"
 Dictionaries = "0.3.15"
 Distributions = "0.25.86"
 DocStringExtensions = "0.8, 0.9"
+EinExprs = "0.6"
```
mtfishman, line 49:
```suggestion
EinExprs = "0.6.4"
```
when the `KaHyPar` extension was added.

mofeing:
> @mofeing still more test failures due to bugs caused by the refactor.

The error in "Julia 1" should be solved, but I can't figure out why this error is happening in "Julia 1.7".

https://github.com/mtfishman/ITensorNetworks.jl/actions/runs/7887034139/job/21524035311

ext/ITensorNetworksEinExprsExt/ITensorNetworksEinExprsExt.jl
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  externalinds_tn = reduce(noncommoninds, ts)
+  return SizedEinExpr(sum(tensor_exprs; skip=externalinds_tn), inds_dims)
+end
+
+function tensor_inds_to_vertex(ts::Vector{ITensor})
+  IndexType = Any
+
+  mapping = Dict{Set{IndexType},VertexType}()
```
mtfishman, line 27: Looks like `VertexType` is also not defined.

mofeing: Ah, damn.

mtfishman:
> > @mofeing still more test failures due to bugs caused by the refactor.
> 
> The error in "Julia 1" should be solved, but I can't figure out why this error is happening in "Julia 1.7".
> 
> https://github.com/mtfishman/ITensorNetworks.jl/actions/runs/7887034139/job/21524035311

From what I can see you haven't added and set up [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl/tree/main) in this PR so the package extension won't work on Julia versions below 1.9.

mofeing:
> From what I can see you haven't added and set up [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl/tree/main) in this PR so the package extension won't work on Julia versions below 1.9.

Ahh, I misunderstood you. I thought you refer to `EinExprs`, not to this package extension.

mtfishman:
> > From what I can see you haven't added and set up [PackageExtensionCompat.jl](https://github.com/cjdoris/PackageExtensionCompat.jl/tree/main) in this PR so the package extension won't work on Julia versions below 1.9.
> 
> Ahh, I misunderstood you. I thought you refer to `EinExprs`, not to this package extension.

Right, both `ITensorNetworks.jl` and `EinExprs.jl` need it in order for the extensions to work on older Julia versions.

JoeyT1994:
Thanks @mtfishman for the reviewing and @mofeing for the PR. Lets set up a call to discuss running some timing tests to compare `einexprs.jl` to the other sequence backends

PR #121, opened 2024-01-11
Refactor interfaces built around alternating update

b-kloss:
This PR aims to make the solver interface more flexible without changing the existing structure substantially. Examples of target use cases:
a) A call to a local subspace expansion preceding every local solve. This would be achieved by defining a solver that calls a local expander routine before e.g. `exponentiate` (both of which follow the solver function and return signature).
b) A call to a global/sweep subspace expansion preceding every n sweeps of TDVP with `exponentiate` as the local solver. This would be accomplished by defining a custom sweeps function that assembles the above pattern, passing the solver as a `step_kwarg`.

- [x] Pass the wavefunction, `psi::TTN`, and the projected Hamiltonian, `PH::AbstractProjTTN`, as a keyword arg by reference all the way to the `solver` functions which is called by `local_update`. This allows the solver function to modify the latter out-of-place, a functionality needed e.g. for subspace expansions accessed as a solver.
- [x] Restructuring and renaming of kwargs for better readability.
- [x] Decide on design:
a) Future goal: Pass the solver function as `step_kwarg`, s.t. the decision about what gets done locally is accessible by modifying the sweeps pattern (which is defined at the same level as the solver).
b) Currently: Implement all of the logic of choosing what solver to use locally inside the solver function (which should have the necessary information about the position in the sweep etc).
- [x] Update `dmrg` solver interface and tests.
- [x] Update `dmrg-x` solver inteface and tests. Also implement linsolve?
- [ ] Implement what used to be triggered by `noise` as updater (for full DMRG functionality, the alternative being subspace expansion (or both) )

Naming updates:
- [ ] Rename `region_kwargs` to `region_update_kwargs` (or combine with `updater_kwargs`?).
- [x] Rename `psi_ref!` to `state!` and `PH_ref!` to `projected_operator!`.
- [x] Rename `psi0` to `init_state`, `psi` to `state`, `H` to `operator`, `PH` to `projected_operator`.

Open questions
- [ ] Role of `extract_local_tensor` when region is an edge (single-site TDVP). Subspace expansion requires truncation at this step, but could be done inside solver, at the cost of performing SVD of a single-vertex-tensor twice. 
- [ ] refactor `extract/insert_local_tensor` using `region` and neighbouring `region`?


src/solvers/exponentiate.jl
```diff
@@ -0,0 +1,42 @@
...
+      normalize,
+      time_step,
+    )
+        #H=copy(PH_ref![])
+        H=PH_ref![] ###since we are not changing H we don't need the copy
+        # let's test whether given region and sweep regions we can find out what the previous and next region were
+        # this will be needed in subspace expansion
+        region_ind=sweep_step
+        next_region=region_ind==length(sweep_regions) ? nothing : first(sweep_regions[region_ind+1])
+        previous_region=region_ind==1 ? nothing : first(sweep_regions[region_ind-1])
```
mtfishman, line 25: I realize this is a work in progress, but I'm a bit confused why this code logic is here since it isn't used by `exponentiate`.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,40 @@
...
+    solver_tol=1e-14,
+    solver_krylovdim=3,
+    solver_maxiter=1,
+    solver_verbosity=0,
+  )
+    
+    function solver(
+        init;
+        psi_ref!,
+        PH_ref!,
```
mtfishman, line 14: I don't think it's obvious but I would vote for just calling these `psi!` and `PH!` (and ideally more descriptive names like `state!` and `projected_operator!`). We can just have a convention that keyword arguments ending in `!` are `Ref`s.

mtfishman:
Thanks for getting started with this. Could you format? It makes it easier to review the PR.

src/solvers/exponentiate.jl
mtfishman: I guess I see from the comment above that this was just a proof of principle and will only be used in the subspace expansion function.

b-kloss: Yes, this was just a placeholder to check whether the information required for subspace information can be passed to the solver. I'll remove it / move it to subspace expansion solver.

src/solvers/eigsolve.jl
b-kloss: I am ok with that. I may keep it around for a while to be more explicit, but will ultimately rename them.

b-kloss:
`format(".")` throws an error, not sure what's going on there ...
I have implemented most of the changes we talked about earlier today, but I may have missed a few things. At least it's functional and a good place to work from.
One issue i was wondering about is how to name the updater_kwargs. I chose to use the underlying function's kwargs (so `krylovdim` instead of `solver_krylovdim` --- but we may decide that it's better to provide a uniform naming scheme from ITensorNetworks side and translate at the level of the solver (I did that for `outputlevel` vs `verbosity`.)

src/solvers/applyexp.jl
```diff
@@ -0,0 +1,25 @@
+function applyexp_solver()
```
mtfishman, line 1: This wrapper function should be removed.

src/solvers/dmrg_x_solver.jl
```diff
@@ -0,0 +1,19 @@
+function dmrg_x_solver(
+  init;
+  psi_ref!,
+  PH_ref!,
+  normalize=nothing,
+  region,
+  sweep_regions,
+  sweep_step,
+  half_sweep,
```
mtfishman, line 9: Do these need to be updated?

src/solvers/dmrg_x_solver.jl
```diff
@@ -0,0 +1,19 @@
...
+  step_kwargs...,
```
mtfishman, line 10: Is this right? I thought we were planning to avoid slurping extra keyword arguments in this way.

src/solvers/dmrg_x_solver.jl
```diff
@@ -0,0 +1,19 @@
+function dmrg_x_solver(
```
mtfishman, line 1:
```suggestion
function dmrg_x_updater(
```

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,38 @@
+
```
mtfishman, line 1:
```suggestion
```

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,38 @@
...
+    ishermitian=true,
+    tol=1e-14,
+    krylovdim=3,
+    maxiter=1,
+    outputlevel=0,
+    eager=false,
+  )
+  updater_kwargs = merge(default_updater_kwargs, updater_kwargs)  #last collection has precedence
+  howmany = 1
+  which = updater_kwargs[:solver_which_eigenvalue]
```
mtfishman, line 24:
```suggestion
  which = updater_kwargs.solver_which_eigenvalue
```
and same thing elsewhere, I like that style better.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,38 @@
+
+function eigsolve_updater(
+  init;
+  psi_ref!,
+  PH_ref!,
+  outputlevel,
+  which_sweep,
+  region_updates,
+  which_region_update,
+  region_kwargs,
```
mtfishman, line 10: I think `region_update_kwargs` sounds better to me.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,38 @@
...
+  updater_kwargs,
+)
+  default_updater_kwargs = (;
+    solver_which_eigenvalue=:SR,
```
mtfishman, line 14:
```suggestion
    which_eigenvalue=:SR,
```
seems better to me.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,38 @@
...
+    howmany,
+    which;
+    ishermitian=updater_kwargs[:ishermitian],
+    tol=updater_kwargs[:tol],
+    krylovdim=updater_kwargs[:krylovdim],
+    maxiter=updater_kwargs[:maxiter],
+    verbosity=updater_kwargs[:outputlevel],
+    eager=updater_kwargs[:eager],
+  )
+  return vecs[1], (; info, energies=vals)
```
mtfishman, line 37:
```suggestion
  return vecs[1], (; info, eigvals=vals)
```
since they won't always be energies (that would depend on the operator).

src/solvers/exponentiate.jl
```diff
@@ -0,0 +1,48 @@
+function exponentiate_updater(
+  init;
+  psi_ref!,
+  PH_ref!,
+  outputlevel,
+  which_sweep,
+  region_updates,
+  which_region_update,
+  region_kwargs,
+  updater_kwargs,
```
mtfishman, line 10: Is the plan to combine these into `updater_kwargs`?

src/solvers/exponentiate.jl
```diff
@@ -0,0 +1,48 @@
...
+  region_ind = which_region_update
+  next_region =
+    region_ind == length(region_updates) ? nothing : first(region_updates[region_ind + 1])
+  previous_region = region_ind == 1 ? nothing : first(region_updates[region_ind - 1])
+
+  phi, exp_info = KrylovKit.exponentiate(
+    H,
+    time_step,
+    init;
+    ishermitian=updater_kwargs[:ishermitian],
```
mtfishman, line 39:
```suggestion
    ishermitian=updater_kwargs.ishermitian,
```
and same below.

src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -46,55 +46,60 @@ function alternating_update(
...
-    checkdone(; psi, sweep, outputlevel, kwargs...) && break
+    checkdone(; psi, which_sweep, outputlevel, kwargs...) && break
   end
-  select!(sweep_observer!, Observers.DataFrames.Not("sweep_printer")) # remove sweep_printer
+  select!(sweep_observer!, Observers.DataFrames.Not("sweep_printer"))
   return psi
 end
 
-function alternating_update(solver, H::AbstractTTN, psi0::AbstractTTN; kwargs...)
+function alternating_update(updater, H::AbstractTTN, psi0::AbstractTTN; kwargs...)
```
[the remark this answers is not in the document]
mtfishman: Also `PH` to `projected_operator`.

src/treetensornetworks/solvers/dmrg.jl
```diff
@@ -1,58 +1,34 @@
...
-end
-
 """
 Overload of `ITensors.dmrg`.
 """
+
+function dmrg_sweep(
+  nsite::Int, graph::AbstractGraph; root_vertex=default_root_vertex(graph)
+)
+  return tdvp_sweep(2, nsite, Inf, graph; root_vertex, reverse_step=false)
```
mtfishman, line 8:
```suggestion
  order = 2
  time_step = Inf
  return tdvp_sweep(order, nsite, time_step, graph; root_vertex, reverse_step=false)
```
so we remember what `2` and `Inf` mean.

src/treetensornetworks/solvers/dmrg.jl
```diff
@@ -1,58 +1,34 @@
...
-  ishermitian=true,
-  solver_tol=1e-14,
-  solver_krylovdim=3,
-  solver_maxiter=1,
-  solver_verbosity=0,
+  nsweeps,  #it makes sense to require this to be defined
+  nsite=2,
+  (sweep_observer!)=observer(),
+  root_vertex=default_root_vertex(init),
+  updater_kwargs=NamedTuple(;),
```
mtfishman, line 19:
```suggestion
  updater_kwargs=(;),
```

mtfishman:
> `format(".")` throws an error, not sure what's going on there

That's too bad. If you think it is a bug with the formatter (i.e. the code you wrote is correct Julia syntax and runs properly, but the formatter errors on it), you can try to track which line of code is giving the formatter issues and then tell the formatter to skip those lines of code with https://domluna.github.io/JuliaFormatter.jl/stable/skipping_formatting/#Skipping-Formatting. Then ideally we could make a minimal example of what is giving the formatter issues and raise an issue.

> One issue i was wondering about is how to name the updater_kwargs. I chose to use the underlying function's kwargs (so `krylovdim` instead of `solver_krylovdim` --- but we may decide that it's better to provide a uniform naming scheme from ITensorNetworks side and translate at the level of the solver (I did that for `outputlevel` vs `verbosity`.)

I would generally bias towards naming the keyword arguments after the keyword argument names of the solver, so then we can just point to the documentation of those solvers for a list of supported keyword arguments. `outputlevel` seems like a good example of one that we translate. Though even for that one, I would consider not passing `outputlevel` to the solver itself (i.e. `KrylovKit.eigsolve`) and if users want to print from within the solver they need to use whatever keyword arguments are defined for that solver backend, and maybe `outputlevel` is just used in code we write ourselves, since each solver may have it's own interface for outputting different things from the solver and it could get messy trying to mesh that with our convention for how `outputlevel` works.

src/solvers/exponentiate.jl
b-kloss: `updater_kwargs` are `updater` specific kwargs, while `region_args` are among the things that we expose to all updaters. in principle, we can nest the region_args into updater_kwargs in the call to `region_update` but I am not sure if that's actually preferable.

src/solvers/dmrg_x_solver.jl
b-kloss: I haven't yet gotten to update `dmrg_x`

src/solvers/exponentiate.jl
```diff
@@ -0,0 +1,48 @@
...
+  updater_kwargs = merge(default_updater_kwargs, updater_kwargs)  #last collection has precedence
+  #H=copy(PH_ref![])
+  H = PH_ref![] ###since we are not changing H we don't need the copy
+  # let's test whether given region and sweep regions we can find out what the previous and next region were
+  # this will be needed in subspace expansion
+  #@show step_kwargs
+  substep = get(region_kwargs, :substep, nothing)
+  time_step = get(region_kwargs, :time_step, nothing)
+  @assert !isnothing(time_step) && !isnothing(substep)
+  region_ind = which_region_update
```
mtfishman, line 30:
```suggestion
  region_ind = which_region_update
```
I think we should avoid renaming variables like this, it makes it harder to read the code later on (i.e. fewer names are better).

src/solvers/exponentiate.jl
mtfishman: Hmm ok, I just have found it hard to keep track of the logic of why certain keyword arguments are bundled in certain ways, how they will be used, etc.

For example, from the perspective of this function, the only argument I can see that is being used here from `region_kwargs` is `time_step`, which I don't think is really any different conceptually from the arguments being passed in `updater_kwargs` (it's just another thing being used by the solver/updater). So it makes sense to me to just bundle those together in one flat `NamedTuple` called `updater_kwargs`.

src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -46,55 +46,60 @@ function alternating_update(
...
-        mindim=mindim[sweep],
-        cutoff=cutoff[sweep],
-        noise=noise[sweep],
+        which_sweep,
+        sweep_params=(;
+          maxdim=maxdim[which_sweep],
+          mindim=mindim[which_sweep],
+          cutoff=cutoff[which_sweep],
+          noise=noise[which_sweep],
+        ),
```
mtfishman, line 81: I would define this up above before the call to `sweep_update` in a `NamedTuple` called `sweep_params`, and then pass that `sweep_params` as a keyword argument, to make this function call simpler.

Also, `sweep_params` doesn't seem like a good name for this, I think we decided on a different name for it but I forget what it was. Maybe these can be bundled together in the `updater_kwargs`, since they are just another set of arguments being passed to the updater?

src/treetensornetworks/solvers/dmrg.jl
```diff
@@ -1,58 +1,34 @@
...
   H,
   init::AbstractTTN;
-  solver_which_eigenvalue=:SR,
-  ishermitian=true,
-  solver_tol=1e-14,
-  solver_krylovdim=3,
-  solver_maxiter=1,
-  solver_verbosity=0,
+  nsweeps,  #it makes sense to require this to be defined
+  nsite=2,
```
mtfishman, line 16: I think it would be better to call this `nsites` instead of `nsite`, since that is closer to what we would say (i.e. "number of sites"). Same change other places where `nsite` is used.

src/solvers/eigsolve.jl
mtfishman: Though the same discussion on `exponentiate_updater` applies here, I think we should discuss how these are being passed and maybe merge them with `updater_kwargs`.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,38 @@
...
+    tol=1e-14,
+    krylovdim=3,
+    maxiter=1,
+    outputlevel=0,
+    eager=false,
+  )
+  updater_kwargs = merge(default_updater_kwargs, updater_kwargs)  #last collection has precedence
+  howmany = 1
+  which = updater_kwargs[:solver_which_eigenvalue]
+  vals, vecs, info = KrylovKit.eigsolve(
```
mtfishman, line 25:
```suggestion
  vals, vecs, info = eigsolve(
```
Does that work, or does it need to be qualified with `KrylovKit.`?

src/solvers/applyexp.jl
mtfishman: And the inner function renamed `applyexp_updater`.

Though maybe we should just get rid of `applyexp` at this point, I don't think there is any advantage of it over `KrylovKit.exponentiate`.

src/solvers/dmrg_x_solver.jl
```diff
@@ -0,0 +1,19 @@
...
+  half_sweep,
+  step_kwargs...,
+)
+  H = contract(PH_ref![], ITensor(1.0))
+  D, U = eigen(H; ishermitian=true)
+  u = uniqueind(U, H)
+  max_overlap, max_ind = findmax(abs, array(dag(init) * U))
+  U_max = U * dag(onehot(u => max_ind))
+  # TODO: improve this to return the energy estimate too
+  return U_max, NamedTuple()
```
mtfishman, line 18:
```suggestion
  return U_max, (;)
```
I prefer using that shorthand for constructing `NamedTuple`.

src/treetensornetworks/solvers/dmrg.jl
```diff
@@ -1,58 +1,34 @@
...
-  end
-  return solver
-end
-
 """
 Overload of `ITensors.dmrg`.
 """
+
+function dmrg_sweep(
+  nsite::Int, graph::AbstractGraph; root_vertex=default_root_vertex(graph)
```
mtfishman, line 6:
```suggestion
  nsites::Int, graph::AbstractGraph; root_vertex=default_root_vertex(graph)
```

src/treetensornetworks/solvers/update_step.jl
```diff
@@ -14,11 +14,12 @@ function default_sweep_regions(nsite, graph::AbstractGraph; kwargs...)
   )
 end
 
-function step_printer(;
-  cutoff, maxdim, mindim, outputlevel::Int=0, psi, region, spec, sweep_step
+function region_printer(;
```
mtfishman, line 17:
```suggestion
function region_update_printer(;
```

src/treetensornetworks/solvers/dmrg.jl
```diff
@@ -1,58 +1,34 @@
...
-    return psi, (; solver_info=info, energies=vals)
-  end
-  return solver
-end
-
 """
 Overload of `ITensors.dmrg`.
 """
+
+function dmrg_sweep(
```
mtfishman, line 5: Maybe we should rename to something more descriptive, like `dmrg_sweep_updates` or `dmrg_sweep_plan` (same for renaming `tdvp_sweep`).

I think I like `dmrg_sweep_plan` since it indicates it is a plan for what to do during the sweep, rather than actually performing the sweep/updates.

src/treetensornetworks/solvers/dmrg.jl
```diff
@@ -1,58 +1,34 @@
...
-      ishermitian,
-      solver_tol,
-      solver_krylovdim,
-      solver_maxiter,
-      solver_verbosity,
-    ),
-    H,
-    init;
-    kwargs...,
+  region_updates = dmrg_sweep(nsite, init; root_vertex)
```
mtfishman, line 22:
```suggestion
  sweep_plan = dmrg_sweep_plan(nsite, init; root_vertex)
```
(Maybe, up for discussion).

src/treetensornetworks/solvers/dmrg_x.jl
```diff
@@ -1,16 +1,4 @@
...
-  u = uniqueind(U, H)
-  max_overlap, max_ind = findmax(abs, array(dag(init) * U))
-  U_max = U * dag(onehot(u => max_ind))
-  # TODO: improve this to return the energy estimate too
-  return U_max, NamedTuple()
-end
-
 function dmrg_x(PH, init::AbstractTTN; kwargs...)
-  psi = alternating_update(dmrg_x_solver, PH, init; kwargs...)
+  psi = alternating_update(ITensorNetworks.dmrg_x_solver, PH, init; kwargs...)
```
mtfishman, line 2:
```suggestion
  psi = alternating_update(dmrg_x_solver, PH, init; kwargs...)
```
That doesn't seem like it should be necessary.

src/treetensornetworks/solvers/tdvp.jl
```diff
@@ -122,17 +68,18 @@ function tdvp(
   (sweep_observer!)=observer(),
   root_vertex=default_root_vertex(init),
   reverse_step=true,
+  updater_kwargs=NamedTuple(;),
   kwargs...,
 )
   nsweeps = _compute_nsweeps(nsteps, t, time_step, order)
-  sweep_regions = tdvp_sweep(order, nsite, time_step, init; root_vertex, reverse_step)
+  region_updates = tdvp_sweep(order, nsite, time_step, init; root_vertex, reverse_step)
```
mtfishman, line 75:
```suggestion
  sweep_plan = tdvp_sweep_plan(order, nsite, time_step, init; root_vertex, reverse_step)
```
Maybe, up for discussion.

src/treetensornetworks/solvers/dmrg.jl
mtfishman: Also, I'm not sure how I feel about calling this `dmrg_sweep` since it could be used by other solvers like `linsolve`. I'm also not a huge fan of calling it `tdvp_sweep` for the same reason.

Maybe we can rethink this API a bit, for example rename the current `tdvp_sweep` to `default_sweep_plan` and then just call that with different inputs from within the different solvers. I think a good design for that would be to choose default values that "just work" for DMRG, including defaulting to `order=2`, and then provide optional inputs like `reverse_step=true` and `time_step` for use with TDVP. That may mean moving some of the inputs like order and time step to keyword arguments, which seems like a better design anyway.

emstoudenmire: Yes, and this is what the `default_sweep_regions` function was trying to be. That is, a sort of "typical" sweep plan that most algorithms (dmrg, linsolve, etc.) could use. We could definitely think about moving that function out of the `update_step.jl` file but the intent there was for it to be a pretty generic default.

mtfishman: I see, I forgot about `default_sweep_regions`, I'm not sure what the fate of that is in the current PR since a lot of things are getting changed in this PR. But it seems like `default_sweep_regions`, `tdvp_sweep`, and `dmrg_sweep` should all get consolidated into a single function, which tentatively we are thinking of calling something like `default_sweep_plan`, `default_region_updates`, or `default_region_update_plans` (since it contains both the regions that will be updated but also information about how to do the update), which I think should just get called directly by the different solvers like `dmrg`, `tdvp`, `linsolve`, etc. and passed to `alternating_update`.

Opinions on what to name that function would be welcome. We want to come up with a good name for that function and what it outputs which we can then use consistently throughout the rest of the code. The list output by `tdvp_sweep`/`dmrg_sweep` is called `region_updates` in the current version of this PR but we are discussing alternative names, which hopefully will be consistent with the name of the new function.

src/solvers/applyexp.jl
b-kloss: I'd vote for removing it.

src/treetensornetworks/solvers/alternating_update.jl
b-kloss: They are also required inside `region_update` (specifically inside `extract/insert local_tensor`). We could decide to move them into updater_kwargs, and then extract them when needed.
With a plan to move to allow to specify each function in region_update (extract, updater, insert) and it's kwargs via NamedTuple attached to the region, I'd for now avoid to putting them all into `updater_kwargs`.

src/solvers/applyexp.jl
mtfishman: Ok, let's remove it in this PR.

src/solvers/exponentiate.jl
mtfishman: Ok, let's discuss this later, this part of the design is hard to follow/keep track of so indicates to me that there is something too complicated about it.

src/solvers/eigsolve.jl
b-kloss: Since this will be obsolete soon, I am in favor of sticking with `region_kwargs`.

src/solvers/exponentiate.jl
b-kloss: As discussed, these will eventually be bundled together in an upcoming PR.

src/treetensornetworks/solvers/dmrg.jl
b-kloss: I incorporated the suggestion, and the sweep_regions constructor is planend to be refactored in an upcoming PR

b-kloss:
Ready to merge from my side.

src/imports.jl
```diff
@@ -34,7 +34,7 @@ import .DataGraphs:
 
 import Graphs: SimpleGraph, is_directed, weights
 
-import KrylovKit: eigsolve, linsolve
+import KrylovKit: eigsolve, linsolve, exponentiate
```
mtfishman, line 37: I don't see why `exponentiate` needs to get imported since from what I can see it isn't being overloaded.

src/solvers/contract.jl
```diff
@@ -0,0 +1,19 @@
...
+  state!,
+  projected_operator!,
+  outputlevel,
+  which_sweep,
+  sweep_plan,
+  which_region_update,
+  region_kwargs,
+  updater_kwargs,
+)
+  v = ITensor(1.0)
```
mtfishman, line 12:
```suggestion
  v = ITensor(true)
```
to be more agnostic about the element type (`1.0` is `Float64` which would then convert single precision calculations to double precision, for example).

src/solvers/contract.jl
```diff
@@ -0,0 +1,19 @@
...
+  projected_operator = projected_operator![]
+  for j in sites(projected_operator)
+    v *= projected_operator.psi0[j]
+  end
+  Hpsi0 = contract(projected_operator, v)
```
mtfishman, line 17:
```suggestion
  vp = contract(projected_operator, v)
```
or some other name, `Hpsi0` doesn't make much since given the new names throughout the code.

src/solvers/dmrg_x.jl
```diff
@@ -0,0 +1,22 @@
...
+  which_sweep,
+  sweep_plan,
+  which_region_update,
+  region_kwargs,
+  updater_kwargs,
+)
+  # this updater does not seem to accept any kwargs?
+  default_updater_kwargs = (;)
+  updater_kwargs = merge(default_updater_kwargs, updater_kwargs)
+  H = contract(projected_operator![], ITensor(1.0))
```
mtfishman, line 15:
```suggestion
  H = contract(projected_operator![], ITensor(true))
```

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,36 @@
...
+    ishermitian=true,
+    tol=1e-14,
+    krylovdim=3,
+    maxiter=1,
+    verbosity=0,
+    eager=false,
+  )
+  updater_kwargs = merge(default_updater_kwargs, updater_kwargs)  #last collection has precedence
+  howmany = 1
+  which, updater_kwargs = _pop_which_eigenvalue(; updater_kwargs...)
```
mtfishman, line 23:
```suggestion
  (; which) = updater_kwargs
  updater_kwargs = Base.structdiff(updater_kwargs, (; which=nothing))
```
instead of defining `_pop_which_eigenvalue`. I'd be ok with separating out a separate function if it was more general (like handled popping out any specified entry of a `NamedTuple`, not just `which`), but as it is `_pop_which_eigenvalue` is very specialized.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,36 @@
...
+    howmany,
+    which;
+    updater_kwargs..., #this leaves it
+  )
+  return vecs[1], (; info, eigvals=vals)
+end
+
+function _pop_which_eigenvalue(; which_eigenvalue, kwargs...)
+  return which_eigenvalue, NamedTuple(kwargs)
+end
```
mtfishman, line 36:
```suggestion
```

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,36 @@
...
+  projected_operator!,
+  outputlevel,
+  which_sweep,
+  sweep_plan,
+  which_region_update,
+  region_kwargs,
+  updater_kwargs,
+)
+  default_updater_kwargs = (;
+    which_eigenvalue=:SR,
```
mtfishman, line 13: I have a slight preference for calling this `which_eigval` since there is a function `eigvals` in `LinearAlgebra`.

src/solvers/eigsolve.jl
b-kloss: Sure, i wasn't aware of how to do it with `structdiff`.

src/solvers/eigsolve.jl
b-kloss: What about calling it `which` (the name of this keyword argument in `KrylovKit`)? Otherwise `which_eigval` is fine.

test/test_treetensornetworks/test_solvers/test_tdvp.jl
```diff
@@ -24,14 +25,14 @@ using Test
...
     # Time evolve forward:
-    ψ1 = tdvp(H, -0.1im, ψ0; nsteps=1, cutoff, nsite=1)
+    ψ1 = tdvp(H, -0.1im, ψ0; nsteps=1, cutoff, nsites=1)
 
     #
-    #TODO: exponentiate is now the default, so switch this to applyexp
+    #TODO: applyexp is removed, no need to test a different backend here and below for now.
     #
-    #Different backend solvers, default solver_backend = "applyexp"
+    #Different backend updaters, default updater_backend = "applyexp"
```
mtfishman, line 33: Seems like we should just remove this comment.

mtfishman: Or is this a reminder to delete some of these tests, perhaps in a future PR?

b-kloss: It seems to me that it is not crucial to test whether passing a user-defined solver works, since the whole code design relies on this being possible (is indirectly also tested in the time-dependent tests.). So can just remove these tests and comments now?

src/solvers/eigsolve.jl
mtfishman: From what I see, in `KrylovKit` `which` isn't a keyword argument, but a positional argument, so the name `which` isn't part of the `KrylovKit` interface (see https://jutho.github.io/KrylovKit.jl/stable/man/eig/#KrylovKit.eigsolve). My concern with calling it `which` is that it isn't clear what it is referring to.

test/test_treetensornetworks/test_solvers/test_tdvp.jl
mtfishman: Yeah that would be better.

src/solvers/eigsolve.jl
b-kloss: Nevermind, it's not a keyword arg in KrylovKit.

test/test_treetensornetworks/test_solvers/test_tdvp.jl
b-kloss: I removed the related tests.

mtfishman:
Looks good, thanks!

PR #122, opened 2024-01-12
Fermionic `OpSum` to `TTN` constructor

b-kloss:
This PR generalizes the construction of `TTN`s from `OpSum`s to systems with fermionic degrees of freedom when `ITensors.using_auto_fermion()==true`.
The implementation closely follows the logic in the `OpSum` to `MPO` constructor in `ITensors.jl` (in fact replicates quite a bit of code) and does not expose the `auto_fermion` logic.
The functionality passes tests (`test/test_opsum_to_ttno.jl`) for fermions and electrons.

Tests added:
- `test_opsum_ttn.jl`: tests for non-fermionic `QN` `inds` on comb graph, and comb graph with a single internal vertex. These tests contract the Hamiltonian `TTNO` and compare with one generated by `ITensors.MPO`. 
- `test_opsum_ttn.jl`: tests for fermionic `QN` `inds` on comb graph for `NNN` tight-binding model on comb graph. Since comparison `\approxeq` between fermionic ITensors is broken, the Hamiltonian is cast into a dense matrix for both `TTN` and `MPO` format (using Combiners) and compared.
- `test_treetensornetworks/test_solvers/test_dmrg.jl`: tests added for `NNN` `heisenberg` and `hubbard` model on comb graph, comparison of ground state energies with result from `ITensors.dmrg`. Tests for single-site DMRG with `QNs` have been removed due to missing `random_ttn` and lack of `subspace_expansion`.

To Do:
- [X] Implement fermionic permutation sign in `ttn_svd`.
- [x] Cleanup convoluted (fermion/arrow-direction related) logic inside `svd_ttn`.
- [X] Implement tests for fermionic systems.
- [X] Test DMRG for NNN Hubbard Model on comb tree. 
- [ ] Figure out whether comparison of fermionic ITensors is broken.
- [x] Add more tests for solvers including fermionic systems?
- [ ] Add test that verifies that vanishing coefficient terms do not influence final bond dimension. (Probably better in another PR since it's unrelated to the fermion implementation.)


b-kloss:
@mtfishman @emstoudenmire 
It looks like `svd_ttn` is working for fermions now and the logic for the fermion-sign really only comes in for the permutation w.r.t. the direct product in the last step.

src/models.jl
```diff
@@ -1,6 +1,90 @@
 _maybe_fill(x, n) = x
 _maybe_fill(x::Number, n) = fill(x, n)
 
+#= Make the free fermion Hamiltonian for the up spins
```
b-kloss, line 4: Remove.

emstoudenmire:
Nice! So did you find & fix that other bug that was showing up for electrons but was related to QNs?

b-kloss:
That one turned out to be a typo in the Hubbard model OpSum that made it
have nonzero flux.

emstoudenmire:
I'm excited to hear it's already working. Is it passing tests on multiple fermionic cases? If so this is a big development.

b-kloss:
It's working for spinless fermions and the Hubbard model (including next-nearest neighbors in both cases).
DMRG results on a tree agree with MPS calculations.
Comparing two fermionic ITensors `@test A ≈ B` seems to be broken but I checked by casting to a Dense matrix that the elements agree.

emstoudenmire:
Those are some pretty strenuous tests. Great. We can discuss the failing ITensor comparison test, since maybe that's more of a low-level bug with ITensors.jl ?

src/models.jl
```diff
@@ -1,6 +1,74 @@
...
+  h = _maybe_fill(h, nv(g))
+  ℋ = OpSum()
+  if !iszero(t)
+    for e in edges(g)
+      ℋ += -t, "Cdag", maybe_only(src(e)), "C", maybe_only(dst(e))
+      ℋ += -t, "Cdag", maybe_only(dst(e)), "C", maybe_only(src(e))
+    end
+  end
+  if !iszero(t')
+    # TODO, more clever way of looping over next to nearest neighbors?
```
mtfishman, line 14: There should be some functions from `Graphs.jl` we can use for this, for example [neighborhood](https://juliagraphs.org/Graphs.jl/dev/algorithms/connectivity/#Graphs.neighborhood-Union{Tuple{T},%20Tuple{U},%20Tuple{AbstractGraph{T},%20Integer,%20Any},%20Tuple{AbstractGraph{T},%20Integer,%20Any,%20AbstractMatrix{U}}}%20where%20{U%3C:Real,%20T%3C:Integer}), i.e. it could be `setdiff(neighborhood(g, v, 2), neighbors(g, v))`.

test/test_treetensornetworks/test_solvers/test_dmrg.jl
```diff
@@ -108,35 +108,95 @@ end
...
+      d[v] = isodd(i) ? "Up" : "Dn"
+    end
+    states = v -> d[v]
+    psi = TTN(s, states)
+
+    nsweeps = 10
+    maxdim = [10, 20, 40, 100]
+    sweeps = Sweeps(nsweeps) # number of sweeps is 5
+    maxdim!(sweeps, 10, 20, 40, 100) # gradually increase states kept
+    cutoff!(sweeps, cutoff)
```
mtfishman, line 184: Let's avoid using the `Sweeps` object, we really should deprecate it at this point, the keyword argument syntax is preferred.

src/models.jl
```diff
@@ -1,6 +1,74 @@
 _maybe_fill(x, n) = x
 _maybe_fill(x::Number, n) = fill(x, n)
 
+function tight_binding(g::AbstractGraph; t=1.0, tp=0.0, h::Union{<:Real,Vector{<:Real}}=0)
+  h = _maybe_fill(h, nv(g))
+  ℋ = OpSum()
+  if !iszero(t)
```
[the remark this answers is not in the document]
b-kloss: It is not necessary, and we should actually test that having terms with zero coefficients doesn't change the bond dimension. However, as a general rule (this is in `src/models.jl`), I am in favor of having this check because it lessens the computational expense of constructing the TTN (in the current implementation).

b-kloss:
I added the tests we discussed:
a) non-interacting model GS energy from many-body and single-particle calculation
b) comparison of `dmrg` with JW-Strings and `ITensors.MPO` with `autofermion` with `ITensors.MPO` and `TTN`

b-kloss:
Assuming that tests are passing, this is ready to merge from my side.

src/models.jl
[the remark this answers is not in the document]
mtfishman: Sounds good, we can reassess later.

src/models.jl
```diff
@@ -1,10 +1,84 @@
...
 _maybe_fill(x::Number, n) = fill(x, n)
 
+function nth_nearest_neighbors(g, v, n::Int)  #ToDo: Add test for this.
+  if n >= 2
+    return setdiff(
+      neighborhood(g, v, n), union(neighborhood(g, v, n - 1), neighborhood(g, v, n - 2))
+    )
+  else
+    return neighborhood(g, v, 1)
+  end
```
mtfishman, line 11:
```suggestion
  isone(n) && return neighbors(g, v)
  return setdiff(
    neighborhood(g, v, n), union(neighborhood(g, v, n - 1), neighborhood(g, v, n - 2))
  )
```
I prefer that style, where special cases are handled first followed by the general implementation, I think it is easier to read compared to nesting things in if-statements.

Also, why is `union(neighborhood(g, v, n - 1), neighborhood(g, v, n - 2))` needed? Isn't `neighborhood(g, v, n - 2)` already included in `neighborhood(g, v, n - 1)`?

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -59,8 +59,10 @@ function ttn_svd(
   vertextype_sites = vertextype(sites)
   thishasqns = any(v -> hasqns(sites[v]), vertices(sites))
 
+  linkdir_ref = ITensors.using_auto_fermion() ? ITensors.In : ITensors.Out
```
mtfishman, line 62: This seems strange to me, could you explain this? Would at least be good to put an explanation as a comment so we can understand it later on.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -458,7 +447,7 @@ function sorteachterm(os::OpSum, sites::IndsNetwork{V,<:Index}, root_vertex::V)
       prevsite = currsite
 
       if fermionic
-        error("No verified fermion support for automatic TTN constructor!") # no verified support, just throw error
+        #error("No verified fermion support for automatic TTN constructor!") # no verified support, just throw error
```
mtfishman, line 450:
```suggestion
```
assuming this is no longer relevant.

test/test_opsum_to_ttn.jl
```diff
@@ -125,19 +141,78 @@ using Test
...
+    tooth_lengths = fill(2, 3)
+    c = named_comb_tree(tooth_lengths)
+    is = siteinds("Fermion", c; conserve_nf=true)
+
+    # test with next-nearest neighbor tight-binding model
+    t = 1.0
+    tp = 0.4
+    U = 0.0
+    h = 0.5
+    H = ITensorNetworks.tight_binding(c; t=t, tp=tp, h=h)
```
mtfishman, line 159:
```suggestion
    H = ITensorNetworks.tight_binding(c; t, tp, h)
```
and same thing elsewhere.

test/test_opsum_to_ttn.jl
```diff
@@ -125,19 +141,78 @@ using Test
...
+
+    # add combination of longer range interactions
+    Hlr = copy(H)
+
+    @testset "Svd approach" for root_vertex in leaf_vertices(is)
+      # get TTN Hamiltonian directly
+      Hsvd = TTN(H, is; root_vertex=root_vertex, cutoff=1e-10)
+      # get corresponding MPO Hamiltonian
+      sites = only.([is[v] for v in reverse(post_order_dfs_vertices(c, root_vertex))])
```
mtfishman, line 168:
```suggestion
      sites = [only(is[v]) for v in reverse(post_order_dfs_vertices(c, root_vertex))]
```

test/test_opsum_to_ttn.jl
```diff
@@ -125,19 +141,78 @@ using Test
...
+
+      @test_broken Tmpo ≈ Tttno # ToDo fix comparison for fermionic tensors
+      # In the meantime: matricize tensors and convert to dense Matrix to compare element by element
+      Tmm = _fermionic_to_dense_matrix(Tmpo)
+      Ttm = _fermionic_to_dense_matrix(Tttno)
+      Tmm, Ttm = _make_inds_same(Tmm, Ttm)
+
+      dTmm = Matrix(dense(Tmm), inds(Tmm)[1], inds(Tmm)[2])
+      dTtm = Matrix(dense(Ttm), inds(Tmm)[1], inds(Tmm)[2])
+      @test any(dTmm - dTtm .> 1e-14)
```
mtfishman, line 191:
```suggestion
      @test any(>(1e-14), dTmm - dTtm)
```

test/test_treetensornetworks/test_solvers/test_dmrg.jl
```diff
@@ -112,37 +112,101 @@ end
...
-  cutoff!(sweeps, cutoff)
+
+  # setup model
+  tooth_lengths = fill(2, 3)
+  c = named_comb_tree(tooth_lengths)
+  s = siteinds("Electron", c; conserve_qns=use_qns)
+  U = 2.0
+  t = 1.3
+  tp = 0.6
+  os = ITensorNetworks.hubbard(c; U=U, t=t, tp=tp)
```
mtfishman, line 171:
```suggestion
  os = ITensorNetworks.hubbard(c; U, t, tp)
```

test/test_opsum_to_ttn.jl
```diff
@@ -125,19 +141,78 @@ using Test
...
+      @test norm(Tmpo) ≈ norm(Tttno) rtol = 1e-6
+
+      @test_broken Tmpo ≈ Tttno # ToDo fix comparison for fermionic tensors
+      # In the meantime: matricize tensors and convert to dense Matrix to compare element by element
+      Tmm = _fermionic_to_dense_matrix(Tmpo)
+      Ttm = _fermionic_to_dense_matrix(Tttno)
+      Tmm, Ttm = _make_inds_same(Tmm, Ttm)
+
+      dTmm = Matrix(dense(Tmm), inds(Tmm)[1], inds(Tmm)[2])
+      dTtm = Matrix(dense(Ttm), inds(Tmm)[1], inds(Tmm)[2])
```
mtfishman, line 190: Code like this and `_make_inds_same` that relies on indices being ordered in a certain way generally isn't a good idea.

It seems like what you want is a function that takes an ITensor that has pairs of primed and unprimed indices and turns it into a matrix from primed to unprimed indices.

So maybe:
```julia
function to_matrix(t::ITensor)
  c = combiner(inds(t; plev=0))
  tc = t * c * dag(c')
  @assert order(tc) == 2
  cind = combinedind(c)
  return matrix(tc, cind', cind)
end
```
and then
```julia
  dTmm = to_matrix(Tmpo)
  dTtm = to_matrix(Tttno)
```

mtfishman:
Looks good, thanks a lot @b-kloss. This is a big step forward for ITensor!

src/treetensornetworks/opsum_to_ttn.jl
mtfishman: I see, so basically it is just a matter of what convention we will use for the index directions of the link indices of the TTNO, and one of them is more convenient for the autofermion system?

I don't care about that convention either way, so if one is simpler/better for the autofermion system we could just go with that.

src/models.jl
b-kloss: from what I could tell `neighborhood(g,v,n)` gives you all vertices that can be visited with by traversing n (non-unique) edges, i.e. you can go back on a given edge. the vertices that can be visited via an odd number of edges is an orthogonal set to those that can be visited by an even number. `neighborhood(g, v, n - 1)` shouldn't be necessary here actually.

test/test_opsum_to_ttn.jl
b-kloss: Ok. I made a specialized function to the same thing to two tensors and replace the indices of one of them, which doesn't look as nice as what you suggested (but addresses the stability issue).

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: I went with `linkdir_ref = ITensors.using_auto_fermion() ? ITensors.In : ITensors.In` to indicate that this may still be changed, and also added a comment. Happy to change it to `linkdir_ref = ITensors.In` instead.

test/test_opsum_to_ttn.jl
```diff
@@ -1,12 +1,31 @@
 using Dictionaries
 using ITensors
+using ITensorGaussianMPS: hopping_hamiltonian
 using ITensorNetworks
 using Random
+using LinearAlgebra: eigvals
 using Test
 
+function _to_matrices_with_same_inds(t::ITensor,s::ITensor)
```
mtfishman, line 9: I don't see why this function needs to accept two tensors, or why it needs to make the indices match since it just removes the indices when converting to `Matrix` anyway. The version I proposed seems simpler.

test/test_opsum_to_ttn.jl
mtfishman: What do you mean by "stability issue"? I prefer the simplicity of the version I proposed.

test/test_opsum_to_ttn.jl
b-kloss: It doesn't make sure that the indices are in the same order.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -54,13 +54,19 @@ function ttn_svd(
   maxdim::Int=typemax(Int),
   cutoff=eps(real(coefficient_type)) * 10,
 )
-  sites = deepcopy(sites0)
+  # fermionic and non-fermionic QNIndices have opposite default index direction
+  # which causes problems when using the autofermion sign with non-fermionic QNIndex
+  # ToDo: remove conditional when this is addressed
+  linkdir_ref = ITensors.using_auto_fermion() ? ITensors.In : ITensors.In
+
+  sites = deepcopy(sites0)  # deepcopy because of inplace modification to handle internal indices
```
mtfishman, line 62: I think just `copy(sites0)` should be sufficient.

test/test_opsum_to_ttn.jl
b-kloss: Ah, since it's combined there's nothing to order, right.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -54,13 +54,19 @@ function ttn_svd(
   maxdim::Int=typemax(Int),
   cutoff=eps(real(coefficient_type)) * 10,
 )
-  sites = deepcopy(sites0)
+  # fermionic and non-fermionic QNIndices have opposite default index direction
+  # which causes problems when using the autofermion sign with non-fermionic QNIndex
+  # ToDo: remove conditional when this is addressed
+  linkdir_ref = ITensors.using_auto_fermion() ? ITensors.In : ITensors.In
```
mtfishman, line 60:
```suggestion
  linkdir = ITensors.In
```

src/treetensornetworks/opsum_to_ttn.jl
mtfishman: I would prefer the simpler version, and then leave a comment explaining it is set to `ITensors.In` since that works better with the autofermion system but is subject to change.

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: i do a push! to edit the internal sites, and I verified that it breaks because of that. I can make it work with copy if you prefer.

mtfishman: What breaks? We should only use `copy` instead of `deepcopy` if it works, and if it works then I would prefer `copy`.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -54,12 +54,9 @@ function ttn_svd(
...
   cutoff=eps(real(coefficient_type)) * 10,
 )
-  # fermionic and non-fermionic QNIndices have opposite default index direction
-  # which causes problems when using the autofermion sign with non-fermionic QNIndex
-  # ToDo: remove conditional when this is addressed
-  linkdir_ref = ITensors.using_auto_fermion() ? ITensors.In : ITensors.In
+  linkdir_ref = ITensors.In   # safe to always use autofermion default here
 
-  sites = deepcopy(sites0)  # deepcopy because of inplace modification to handle internal indices 
+  sites = copy(sites0)  # deepcopy because of inplace modification to handle internal indices
```
mtfishman, line 59:
```suggestion
  sites = copy(sites0)  # copy because of inplace modification to handle internal indices 
```

test/test_opsum_to_ttn.jl
```diff
@@ -1,12 +1,23 @@
...
+  tc = (t * c) * dag(c')
+  cind = combinedind(c)
+  return matrix(tc, cind', cind)
+end
+
 @testset "OpSum to TTN converter" begin
   @testset "OpSum to TTN" begin
     # small comb tree
+
+    ITensors.disable_auto_fermion()
```
mtfishman, line 20: I guess this is to circumvent the bug you mentioned? Would be good to leave a comment about that, so we know to remove this line when that bug is fixed.

test/test_opsum_to_ttn.jl
```diff
@@ -72,6 +83,7 @@ using Test
   end
 
   @testset "OpSum to TTN QN" begin
+    ITensors.disable_auto_fermion()
```
mtfishman, line 86: Same comment as above.

test/test_opsum_to_ttn.jl
```diff
@@ -125,19 +137,74 @@ using Test
...
+    end
+  end
+
   @testset "OpSum to TTN QN missing" begin
     # small comb tree
     tooth_lengths = fill(2, 3)
     c = named_comb_tree(tooth_lengths)
     c2 = copy(c)
+    ## add an internal vertex into the comb graph c2
     import ITensorNetworks: add_vertex!, rem_vertex!, add_edge!, rem_edge!
```
mtfishman, line 201: I only use `import` if I am overloading a function, not just bringing it into the namespace to call it. Also it would be better to move this to the top of the file, and bring it into the namespace from the `Graphs.jl` package where they are originally defined, i.e. `using Graphs: add_vertex!, rem_vertex!, add_edge!, rem_edge!`.

src/models.jl
```diff
@@ -1,10 +1,79 @@
...
+  h = _maybe_fill(h, nv(g))
+  ℋ = OpSum()
+  if !iszero(t)
+    for e in edges(g)
+      ℋ -= t, "Cdag", maybe_only(src(e)), "C", maybe_only(dst(e))
+      ℋ -= t, "Cdag", maybe_only(dst(e)), "C", maybe_only(src(e))
+    end
+  end
+  if !iszero(t')
+    # TODO, more clever way of looping over next to nearest neighbors?
```
mtfishman, line 21:
```suggestion
```

test/test_treetensornetworks/test_solvers/test_dmrg.jl
```diff
@@ -112,37 +112,111 @@ end
...
   tooth_lengths = fill(2, 3)
   c = named_comb_tree(tooth_lengths)
-  s = siteinds("S=1/2", c)
-
-  os = ITensorNetworks.heisenberg(c)
 
-  H = TTN(os, s)
-
-  psi = random_ttn(s; link_space=20)
+  @testset "Svd approach" for use_qns in [false, true]
```
mtfishman, line 121:
```suggestion
  @testset "SVD approach" for use_qns in [false, true]
```

mtfishman:
Looks good, thanks again. I'll merge once tests pass.

PR #123, opened 2024-01-12
Fix `externalinds`, add tests for `externalinds` and `internalinds`

JoeyT1994:
This gives a quick bug fix for `externalinds`, so it searches over the vertices of the `ITensorNetwork` instead of the `edges`.
Two tests are added to an existing testset for `externalinds` and `internalinds`.

`externalinds` and `internalinds` are now exported so the user can just call `externalinds(tn)` to get a vector of all open indices in `tn`.


mtfishman:
Looks good, thanks!

PR #126, opened 2024-01-22
Partitioned ITensorNetwork

JoeyT1994:
This PR utilizes the new `PartitionedGraph` structure from `NamedGraphs.jl` (see https://github.com/mtfishman/NamedGraphs.jl/pull/44) to re-write the belief propagation code.

Specifcally, we can now define a `PartitionedITensorNetwork` simply via

`pitn = PartitionedGraph(itn, args....)` where `args` are just some arguments on how to partition the network.

The resulting data structure contains `itn` and partitioning information. This simplifies the nested data graph implementation we were using previously for a partitioned ITensorNetwork. 

The belief propagation code, examples and tests are now all re-written to reflect this and the code should be simpler. Message tensors are now just taken to be a map between the edges of `partitioned_graph(pitn)` and `Vector{ITensor}`. Partitioning information is contained entirely in `pitn`.

Lots of the code in `src/partition.jl`has been deleted as it is redundant now. Code has only been kept when it is relevant to `approx_itensornetwork.jl` which has not been re-written with the new formalism yet. This has all been moved to the folder `src/approx_itensornetwork`


src/ITensorNetworks.jl
```diff
@@ -53,7 +53,8 @@ using NamedGraphs:
   parent_graph,
   vertex_to_parent_vertex,
   parent_vertices_to_vertices,
-  not_implemented
+  not_implemented,
+  parent
```
mtfishman, line 57: Is this needed? Isn't it just an overload of `Base.parent`, and therefore is already available in the namespace?

examples/gauging/gauging_itns.jl
```diff
@@ -7,38 +7,38 @@ using SplitApplyCombine
...
+  ψ::ITensorNetwork, pψψ::PartitionedGraph, bond_tensors::DataGraph, mts
+)
   isometries = vidal_itn_isometries(ψ, bond_tensors)
 
   ψ = copy(ψ)
   mts = copy(mts)
   for e in edges(ψ)
-    s1, s2 = find_subgraph((src(e), 1), mts), find_subgraph((dst(e), 1), mts)
+    vsrc, vdst = src(e), dst(e)
+    pe = which_partitionedge(pψψ, NamedEdge((vsrc, 1) => (vdst, 1)))
```
mtfishman, line 34:
```suggestion
    pe = partitionedge(pψψ, edgetype(pψψ)((vsrc, 1) => (vdst, 1)))
```
I think, not checked.

mtfishman: Also, ideally `partitionedge` would be able to accept either an `AbstractEdge` or a `Pair` as the second argument.

examples/apply/apply_bp/apply_bp.jl
```diff
@@ -52,26 +49,26 @@ function simple_update_bp(
...
         @assert order(only(mts[reverse(e)])) == 2
       end
 
       @assert length(v⃗) == 2
       v1, v2 = v⃗
 
-      s1 = find_subgraph((v1, 1), mts)
-      s2 = find_subgraph((v2, 1), mts)
-      envs = get_environment(ψψ, mts, [(v1, 1), (v1, 2), (v2, 1), (v2, 2)])
+      pe = partitionedge(pψψ, NamedEdge((v1, 1) => (v2, 1)))
```
mtfishman, line 70:
```suggestion
      pe = partitionedge(pψψ, edgetype(pψψ)((v1, 1) => (v2, 1)))
```
Same comment as below, ideally `partitionedge` could accept either an `AbstractEdge` or a `Pair` as the second input.

examples/gauging/gauging_itns.jl
mtfishman: It looks like this file isn't being included in the tests, since the tests pass but it seems incorrect to use `which_partitionedge`.

examples/apply/apply_bp/apply_bp.jl
```diff
@@ -52,26 +49,26 @@ function simple_update_bp(
...
+      envs = get_environment(pψψ, mts, [(v1, 1), (v1, 2), (v2, 1), (v2, 2)])
```
mtfishman, line 71: In this kind of code, can we use:
```julia
envs = get_environment(pψψ, mts, PartitionVertex.([v1, v2]))
```
to avoid having to specify the bra and ket vertices separately?

This is related to a more general question about how we should think about inner products/quadratic forms as tensor networks, which is relevant for the `alternating_udpate` code generalization, since we want a way to naturally associate vertices `(v1, 1)` and `(v1, 2)` with each other. One way is that they are part of the same partition.

examples/apply/apply_bp/apply_bp.jl
```diff
@@ -92,29 +89,28 @@ function simple_update_bp(
...
     if regauge
       println("regauge")
       mts = belief_propagation(
-        ψψ, mts; contract_kwargs=(; alg="exact"), niters=50, target_precision=1e-5
+        pψψ, mts; contract_kwargs=(; alg="exact"), niters=50, target_precision=1e-5
       )
     end
   end
-  return ψ, mts
+  return ψ, pψψ, mts
```
mtfishman, line 103: It feels a bit strange to me to output `ψ`, `pψψ`, and `mts`, since there is a lot of overlapping information across those objects. For example, `ψ` is already stored in `pψψ`, both `pψψ` and `mts` store some information about the graph partitioning.

I think this is worth thinking about more carefully, since it has implications for how to design the BP-based `alternating_update` code.

examples/apply/apply_bp/apply_bp.jl
```diff
@@ -92,29 +89,28 @@ function simple_update_bp(
...
   println("Simple update, Vidal gauge")
   ψψ = norm_network(ψ)
-  mts = message_tensors(partition(ψψ, group(v -> v[1], vertices(ψψ))))
+  pψψ = PartitionedGraph(ψψ, group(v -> v[1], vertices(ψψ)))
   mts = belief_propagation(
-    ψψ, mts; contract_kwargs=(; alg="exact"), niters=50, target_precision=1e-5
+    pψψ; contract_kwargs=(; alg="exact"), niters=50, target_precision=1e-5
   )
-  ψ, bond_tensors = vidal_gauge(ψ, mts)
+  ψ, bond_tensors = vidal_gauge(ψ, pψψ, mts)
```
mtfishman, line 113: Again, it feel strange to me to now have to pass `ψ`, `pψψ`, and `mts`. Let's discuss this, my guess is there is a better design.

examples/belief_propagation/bpexample.jl
```diff
@@ -31,41 +27,33 @@ function main()
...
-  mts = belief_propagation(ψψ, mts; contract_kwargs=(; alg="exact"), niters=20)
-
-  numerator_network = approx_network_region(
-    ψψ, mts, [(v, 1)]; verts_tn=ITensorNetwork([apply(op("Sz", s[v]), ψ[v])])
+  numerator_tensors = approx_network_region(
+    pψψ, mts, [(v, 1)]; verts_tensors=[apply(op("Sz", s[v]), ψ[v])]
   )
-  denominator_network = approx_network_region(ψψ, mts, [(v, 1)])
-  sz_bp = contract(numerator_network)[] / contract(denominator_network)[]
+  denominator_tensors = approx_network_region(pψψ, mts, [(v, 1)])
```
mtfishman, line 37:
```suggestion
  denominator_tensors = approx_network_region(pψψ, mts, PartitionVertex(v))
```
Would that work?

examples/belief_propagation/bpexample.jl
```diff
@@ -31,41 +27,33 @@ function main()
...
+  mts = belief_propagation(
+    pψψ; contract_kwargs=(; alg="exact"), verbose=true, niters=10, target_precision=1e-3
   )
-
-  mts = belief_propagation(ψψ, mts; contract_kwargs=(; alg="exact"), niters=20)
-
-  numerator_network = approx_network_region(
-    ψψ, mts, [(v, 1)]; verts_tn=ITensorNetwork([apply(op("Sz", s[v]), ψ[v])])
+  numerator_tensors = approx_network_region(
+    pψψ, mts, [(v, 1)]; verts_tensors=[apply(op("Sz", s[v]), ψ[v])]
```
mtfishman, line 35:
```suggestion
    pψψ, mts, [PartitionVertex(v)]; verts_tensors=[apply(op("Sz", s[v]), ψ[v])]
```
Same kind of comment as elsewhere, it seems cleaner and clearer to me to start using the `PartitionVertex` type for this kind of thing, unless I am misunderstanding the code.

examples/belief_propagation/bpexample.jl
```diff
@@ -78,30 +66,26 @@ function main()
   ψψ = combine_linkinds(ψψ, combiners)
   ψOψ = combine_linkinds(ψOψ, combiners)
 
-  Z = partition(ψψ, group(v -> v[1], vertices(ψψ)))
-  maxdim = 8
-  mts = message_tensors(Z)
-
+  pψψ = PartitionedGraph(ψψ, collect(values(group(v -> v[1], vertices(ψψ)))))
```
mtfishman, line 69: Are `collect` and `values` needed here? Seems like it got more complicated.

examples/belief_propagation/bpsequences.jl
```diff
@@ -39,36 +34,40 @@ function main()
     ψψ = ψ ⊗ prime(dag(ψ); sites=[])
 
     #Initial message tensors for BP
-    mts_init = message_tensors(
-      ψψ; subgraph_vertices=collect(values(group(v -> v[1], vertices(ψψ))))
-    )
+    pψψ = PartitionedGraph(ψψ, collect(values(group(v -> v[1], vertices(ψψ)))))
```
mtfishman, line 37: Same comment, are `collect` and `values` necessary?

examples/boundary.jl
```diff
@@ -8,8 +8,10 @@ tn = ITensorNetwork(named_grid((6, 3)); link_space=4)
 
 @visualize tn
 
-g = subgraph_vertices(tn; nvertices_per_partition=2)
-sub_vs_1, sub_vs_2 = g[1], g[2]
+ptn = PartitionedGraph(
+  tn, partitioned_vertices(underlying_graph(tn); nvertices_per_partition=2)
+)
+sub_vs_1, sub_vs_2 = vertices(ptn, PartitionVertex(1)), vertices(ptn, PartitionVertex(2))
```
mtfishman, line 14: I think this would be easier to read if it was split onto two lines.

examples/gauging/gauging_itns.jl
```diff
@@ -7,38 +7,38 @@ using SplitApplyCombine
...
+  belief_propagation_iteration
 
 using NamedGraphs
 using NamedGraphs: add_edges!, rem_vertex!, hexagonal_lattice_graph
 using Graphs
 
 """Eager Gauging"""
-function eager_gauging(ψ::ITensorNetwork, bond_tensors::DataGraph, mts::DataGraph)
+function eager_gauging(
+  ψ::ITensorNetwork, pψψ::PartitionedGraph, bond_tensors::DataGraph, mts
```
mtfishman, line 26: I think it's unfortunate that with this new design we know have even more inputs to this function, let's discuss that.

test/test_indsnetwork.jl
```diff
@@ -1,6 +1,7 @@
 using Dictionaries
 using ITensors
 using ITensorNetworks
+using ITensorNetworks: dim
```
mtfishman, line 4: Could you explain this addition?

src/exports.jl
```diff
@@ -73,6 +73,7 @@ export AbstractITensorNetwork,
   TreeTensorNetwork,
   TTN,
   random_ttn,
+  PartitionedITensorNetwork,
```
mtfishman, line 76: Is this defined anywhere?

src/beliefpropagation/sqrt_beliefpropagation.jl
```diff
@@ -74,48 +67,49 @@ function sqrt_belief_propagation(
...
+  regularization=10 * eps(real(scalartype(unpartitioned_graph(pψψ)))),
 )
   sqrt_mts = copy(mts)
-  for e in edges(ψ)
-    vsrc, vdst = src(e), dst(e)
-    ψvsrc, ψvdst = ψ[vsrc], ψ[vdst]
+  for e in partitionedges(pψψ)
+    vsrc, vdst = filter(v -> v[2] == 1, vertices(pψψ, src(e))),
+    filter(v -> v[2] == 1, vertices(pψψ, dst(e)))
+    ψvsrc, ψvdst = unpartitioned_graph(pψψ)[only(vsrc)],
```
mtfishman, line 104: Best to split into separate lines.

src/beliefpropagation/sqrt_beliefpropagation.jl
```diff
@@ -74,48 +67,49 @@ function sqrt_belief_propagation(
...
+    unpartitioned_graph(pψψ)[only(vdst)]
```
mtfishman, line 105: Best to split into separate lines.

src/beliefpropagation/sqrt_beliefpropagation.jl
```diff
@@ -74,48 +67,49 @@ function sqrt_belief_propagation(
...
 
 function sqrt_message_tensors(
-  ψ::ITensorNetwork,
-  mts::DataGraph;
-  eigen_message_tensor_cutoff=10 * eps(real(scalartype(ψ))),
-  regularization=10 * eps(real(scalartype(ψ))),
+  pψψ::PartitionedGraph,
+  mts;
+  eigen_message_tensor_cutoff=10 * eps(real(scalartype(unpartitioned_graph(pψψ)))),
+  regularization=10 * eps(real(scalartype(unpartitioned_graph(pψψ)))),
```
mtfishman, line 98: I guess this kind of thing is unfortunate.

Ideally `scalartype` would be able to reach into any iterable object, and then it could "just work" for partitioned tensor networks. I started an issue about that here:  #127.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -169,48 +135,46 @@ function belief_propagation(
...
+  return belief_propagation(ptn, mts; kwargs...)
 end
-
 """
 Given a subet of vertices of a given Tensor Network and the Message Tensors for that network, return a Dictionary with the involved subgraphs as keys and the vector of tensors associated with that subgraph as values
 Specifically, the contraction of the environment tensors and tn[vertices] will be a scalar.
 """
-function get_environment(tn::ITensorNetwork, mts::DataGraph, verts::Vector; dir=:in)
-  subgraphs = unique([find_subgraph(v, mts) for v in verts])
+function get_environment(ptn::PartitionedGraph, mts, verts::Vector; dir=:in)
```
mtfishman, line 149: It seems like there should be two versions of this function, one which takes partition vertices and one that takes underlying graph vertices, which could just call the one that takes partition vertices.

mtfishman: Also, could you explain the `dir` keyword argument?

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -169,48 +135,46 @@ function belief_propagation(
...
+  env_tensors = reduce(vcat, env_tensors; init=ITensor[])
+  central_tensors = ITensor[
+    (unpartitioned_graph(ptn))[v] for v in setdiff(central_verts, verts)
+  ]
+
+  return vcat(env_tensors, central_tensors)
 end
 
 """
 Calculate the contraction of a tensor network centred on the vertices verts. Using message tensors.
```
mtfishman, line 168: This makes it seems like it actually performs the contraction, while it appears to output a list of uncontracted tensors.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -169,48 +135,46 @@ function belief_propagation(
...
 Defaults to using tn[verts] as the local network but can be overriden
 """
 function approx_network_region(
```
mtfishman, line 171: Like with `get_environment` it seems like this should have versions that accept partition vertices or underlying graph vertices.

test/test_indsnetwork.jl
JoeyT1994: I was founding this test errored without it because it was complaining it was not defined. Similar to the need to import `parent` above. I think this might be a Julia / compile/ Revise issue or something though because you are right you shouldn't need to explicitly import something that is already imported by Base

mtfishman: `dim` is defined in `NDTensors` but reexported by `ITensors`, so it should get brought into the namespace by `using ITensors`.

src/beliefpropagation/beliefpropagation.jl
JoeyT1994: Should we name it `get_environment_tensors` ?

src/beliefpropagation/beliefpropagation.jl
JoeyT1994: I agree that there should now be two versions of this function. Let me work that out. The `dir` argument is whether you want the environment tensors coming on to `verts` or whether you want the environment tensors from `verts` to the rest of the network. This could be dropped though as it is confusing and could be handled by just directly passing `setdiff(vertives(ptn), verts)` by the user.

src/beliefpropagation/beliefpropagation.jl
JoeyT1994: Agreed. Let me write those out.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,70 @@
...
-)
-  mts = message_tensors_skeleton(subgraphs)
-  for e in edges(subgraphs)
-    inds_e = commoninds(subgraphs[src(e)], subgraphs[dst(e)])
-    itensors = itensor_constructor(inds_e)
-    mts[e] = ITensorNetwork(itensors)
-    mts[reverse(e)] = dag(mts[e])
+  mts = Dict()
+  for e in partitionedges(ptn)
+    src_e_itn = unpartitioned_graph(subgraph(ptn, [src(e)]))
```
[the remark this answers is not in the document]
JoeyT1994: Hmmm, and then `subgraph(ptn, v::PartitionVertex)` outputs a partitioned graph? I could see the logic in it.

src/beliefpropagation/beliefpropagation.jl
mtfishman: Yeah, best to drop that.

src/beliefpropagation/beliefpropagation.jl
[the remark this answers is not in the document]
mtfishman: I was suggesting that `subgraph(ptn, v::PartitionVertex)` would be a special case that outputs an unpartitioned graph (i.e. just the subgraph of the unpartitioned graph on the vertices contained in the partition).

src/beliefpropagation/beliefpropagation.jl
mtfishman: Sure, that makes sense.

src/beliefpropagation/beliefpropagation.jl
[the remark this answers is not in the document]
JoeyT1994: I see, yeah I could make that change to `NamedGraphs` !

src/beliefpropagation/beliefpropagation.jl
mtfishman: We could just call it `environment_tensors`.

src/beliefpropagation/beliefpropagation.jl
mtfishman: Also, `approx_network_region` seems like a strange function to me. Maybe it isn't really needed at this point given the added convenience of `PartitionedGraph`, since it really is just grabbing the environment and then tensors in a partition?

JoeyT1994: Yeah it is just grabbing the environment and then concatenating the `ITensors` with some local tensors on the partition. I think removing it would be fine actually.

examples/boundary.jl
```diff
@@ -8,8 +8,10 @@ tn = ITensorNetwork(named_grid((6, 3)); link_space=4)
 
 @visualize tn
 
-g = subgraph_vertices(tn; nvertices_per_partition=2)
-sub_vs_1, sub_vs_2 = g[1], g[2]
+pvertices = partitioned_vertices(underlying_graph(tn); nvertices_per_partition=2)
```
mtfishman, line 11: Is `underlying_graph` needed here?

examples/distances.jl
```diff
@@ -14,4 +14,4 @@ t = dijkstra_tree(ψ, only(center(ψ)))
 @show a_star(ψ, (2, 1), (2, 5))
 @show mincut_partitions(ψ)
 @show mincut_partitions(ψ, (1, 1), (3, 5))
-@show subgraph_vertices(ψ; npartitions=2)
+@show partitioned_vertices(underlying_graph(ψ); npartitions=2)
```
mtfishman, line 17: Is `underlying_graph` needed here?

src/ITensorNetworks.jl
mtfishman: You resolved but I'm not sure how it is resolved.

src/apply.jl
```diff
@@ -314,13 +314,13 @@ function ITensors.apply(
 
     for vn in neighbors(ψ, src(e))
       if (vn != dst(e))
-        ψv1 = noprime(ψv1 * bond_tensors[vn => src(e)])
+        ψv1 = noprime(ψv1 * bond_tensors[NamedEdge(vn => src(e))])
```
mtfishman, line 317: It's kind of unfortunate that we need to add the `NamedEdge` wrapper now.

Maybe we could use a `DataGraph` for the cache but just keep it simple and only have edge data?

src/apply.jl
```diff
@@ -339,17 +339,17 @@ function ITensors.apply(
     replaceind!(S, ind_to_replace, ind_to_replace_with')
     replaceind!(V, ind_to_replace, ind_to_replace_with)
```
mtfishman, line 340:
```suggestion
    S = replaceind(S, ind_to_replace => ind_to_replace_with')
    V = replaceind(V, ind_to_replace => ind_to_replace_with)
```

src/apply.jl
```diff
@@ -339,17 +339,17 @@ function ITensors.apply(
...
 
-    ψv1, bond_tensors[e], ψv2 = U * Qᵥ₁, S, V * Qᵥ₂
+    ψv1, bond_tensors[e], bond_tensors[reverse(e)], ψv2 = U * Qᵥ₁, S, S, V * Qᵥ₂
```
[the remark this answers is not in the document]
mtfishman: Also a bit unfortunate it got more complicated, this kind of code indicates it may be better to use a more sophisticated data structure like `DataGraph`.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,67 @@
 function message_tensors(
-  tn::ITensorNetwork;
-  nvertices_per_partition=nothing,
-  npartitions=nothing,
-  subgraph_vertices=nothing,
-  kwargs...,
+  ptn::PartitionedGraph; itensor_constructor=inds_e -> ITensor[dense(delta(inds_e))]
```
mtfishman, line 2: I would separate this out into a `default_message_tensor_constructor(inds_e) = ...` function.

Also, probably should use `denseblocks` instead of `dense` to handle QN tensors.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,67 @@
...
 """
-DO a single update of a message tensor using the current subgraph and the incoming mts
+Do a single update of a message tensor using the current subgraph and the incoming mts
 """
 function update_message_tensor(
-  tn::ITensorNetwork,
-  subgraph_vertices::Vector,
-  mts::Vector{ITensorNetwork};
-  contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
+  ptn::PartitionedGraph, edge::PartitionEdge, mts; contract_kwargs=(; alg="exact")
```
mtfishman, line 19: I would define `default_contract_kwargs() = (; alg="exact")` and then set `contract_kwargs=default_contract_kwargs()`, then you can use that throughout the code in other parts and we can easily change defaults if we want to from one location.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,67 @@
...
 
-  contract_list = ITensor[mts_itensors; ITensor[tn[v] for v in subgraph_vertices]]
-  tn = if isone(length(contract_list))
-    copy(only(contract_list))
-  else
-    ITensorNetwork(contract_list)
-  end
+  contract_list = ITensor[
+    incoming_messages
+    ITensor(subgraph(ptn, src(edge)))
```
mtfishman, line 30: What is the `ITensor(...)` constructor doing here?

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,67 @@
...
-
-  itn = if typeof(contract_output) == ITensor
-    ITensorNetwork(contract_output)
+    mt = first(contract(ITensorNetwork(contract_list); contract_kwargs...))
   else
-    first(contract_output)
+    mt = contract(
+      contract_list; sequence=contraction_sequence(contract_list; alg="optimal")
+    )
   end
```
mtfishman, line 39: Use an if-statement here is a design red flag for me.

Given that there are so many potential ways we might want to update message tensors (including updates that may also gauge the state, many potential contraction backends, etc.), I think it would be better to base this around a callback function that can be passed to the user to perform a message tensor update.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,67 @@
...
-  normalize!.(vertex_data(itn))
 
-  return itn
-end
+  mt = isa(mt, ITensor) ? ITensor[mt] : ITensor(mt)
+  normalize!.(mt)
```
mtfishman, line 42: This code doesn't feel right to me. Can't was always use `Vector{ITensor}`?

Also, I think it would be better to do out-of-place normalization, for example `mt .= normalize.(mt)`.

examples/distances.jl
JoeyT1994: No it isn't good spot!

src/ITensorNetworks.jl
JoeyT1994: Good spot, I will remove from imports. Sorry about that

src/beliefpropagation/beliefpropagation.jl
JoeyT1994: `subgraph(ptn, src(edge))` returns an `ITensorNetwork` over the corresponding `PartitionVertex`: `ITensor(itn::ITensorNetwork)` returns a `Vector{ITensor}` of the `ITensors` that make up the `itn`?

src/beliefpropagation/beliefpropagation.jl
JoeyT1994: Yeah I agree there is some way to unify this. Currently `contract` could return either an `ITensor` or a tuple corresponding to an `ITensorNetwork` and a `Float64` (from the return of `approx_ITensorNetwork`). So I was trying to account for both of these but maybe it should be some default. I could also imagine a world where `contract` returns `Vector{ITensor}` too

src/ITensorNetworks.jl
mtfishman: No problem, thanks.

src/beliefpropagation/beliefpropagation.jl
mtfishman: I think one way to think about this is to not decide on a particular output of the message tensor update function, but rather require it to output the correct format/data for the message tensor cache.

src/beliefpropagation/beliefpropagation.jl
JoeyT1994: Okay I have renamed that function to `Vector{ITensor}(::ITensorNetwork)`!

JoeyT1994:
Okay I have made changes which avoid the if statements that were in the `belief_propagation` code. Now the user can specify things like the `cache` for belief propagation, how to construct the message tensors and how to contract them together.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,82 @@
...
-  mts_itensors = reduce(vcat, ITensor.(mts); init=ITensor[])
-
-  contract_list = ITensor[mts_itensors; ITensor[tn[v] for v in subgraph_vertices]]
-  tn = if isone(length(contract_list))
-    copy(only(contract_list))
-  else
-    ITensorNetwork(contract_list)
-  end
+  pedges = setdiff(
+    partitionedges(ptn, boundary_edges(ptn, vertices(ptn, src(edge)); dir=:in)),
```
mtfishman, line 43: I think this would be good to define in a variable up above, like `partition_boundary_edges = ...` since this is a long line to read.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -134,30 +112,30 @@ function belief_propagation_iteration(
...
 function belief_propagation_iteration(
-  tn::ITensorNetwork,
-  mts::DataGraph;
-  contract_kwargs=(; alg="density_matrix", output_structure=path_graph_structure, maxdim=1),
+  ptn::PartitionedGraph,
+  mts;
+  contractor=default_contractor,
   compute_norm=false,
-  edges=edge_sequence(mts),
+  edges=PartitionEdge.(edge_sequence(partitioned_graph(ptn))),
```
mtfishman, line 119: Should separate off into a `default_edge_sequence(::PartitionedGraph)` function.

src/gauging.jl
```diff
@@ -26,15 +27,13 @@ function vidal_gauge(
     vsrc, vdst = src(e), dst(e)
     ψvsrc, ψvdst = ψ_vidal[vsrc], ψ_vidal[vdst]
 
-    s1, s2 = find_subgraph((vsrc, 1), mts), find_subgraph((vdst, 1), mts)
-    edge_ind = commoninds(ψ_vidal[vsrc], ψ_vidal[vdst])
+    pe = partitionedge(pψψ, NamedEdge((vsrc, 1) => (vdst, 1)))
```
mtfishman, line 30: Can this work if we drop of `NamedEdge` constructor? Maybe would require a generalization of `partitionedge` in `NamedGraphs`, but ideally anywhere an `AbstractEdge` is accepted, we can also use a `Pair`.

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,82 @@
...
+    contraction_sequence_alg="optimal",
+    svd_kwargs...,
   )
-  for v in vertices(mts)
-    mts[v] = subgraphs[v]
-  end
 
+  tn, _ = contract(ITensorNetwork(contract_list); contract_kwargs...)
+  mts = Vector{ITensor}(tn)
+  mts = normalize!.(mts)
```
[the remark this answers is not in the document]
JoeyT1994: I don't think `Normalize(it::ITensor)` exists? Or at least it is not exported by `ITensors`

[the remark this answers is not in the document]
mtfishman: Ok, we could do:
```julia
  mts .= normalize!.(copy.(mts))
```

src/gauging.jl
mtfishman: Yeah, we should add a definition like that.

src/beliefpropagation/beliefpropagation.jl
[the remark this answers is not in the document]
mtfishman: The main goal here is to not modify the tensor data in-place, and instead modify the collection storing the tensors.

I generally try to avoid in-place modification of tensor data unless it is really necessary for performance, since it makes code easier to reason about and can make the code less generic (for example if the tensor data can't be modified for some reason).

src/beliefpropagation/beliefpropagation.jl
```diff
@@ -1,104 +1,82 @@
...
+function default_contractor(contract_list::Vector{ITensor})
+  return ITensor[normalize!(
+    contract(contract_list; sequence=contraction_sequence(contract_list; alg="optimal"))
+  )]
 end
 
-function message_tensors_skeleton(subgraphs::DataGraph)
-  mts = DataGraph{vertextype(subgraphs),vertex_data_type(subgraphs),ITensorNetwork}(
-    directed_graph(underlying_graph(subgraphs))
+function contract_to_MPS(contract_list::Vector{ITensor}; svd_kwargs...)
```
mtfishman, line 9:
```suggestion
function contract_to_mps(contract_list::Vector{ITensor}; svd_kwargs...)
```

JoeyT1994:
Okay I re-wrote it based on function callback like we said. Currently the `contract` functions have been put in `contract.jl` but there is clearly some overlap there with existing `contract` functions. Up to you whether we re-write that a bit now or we leave as is until a future PR

mtfishman:
Let's leave that for a future PR.

This is definitely a big improvement to the BP code, and will make it a lot easier to implement more ambitious plans, like using BP in other functions like `inner`, `expect`, `alternating_update`, etc.

Issue #128, opened 2024-01-24
Efficient implementation of `expect` for `TTN`?


b-kloss:
`expect(op,state)` with `state` a `TTN` does not appear to be efficient: 

```julia
using ITensors
using ITensorNetworks
s=siteinds("S=1/2",10)
a=randomMPS(s;linkdims=100)
b=random_mps(s;internal_inds_space=100)
expect(a,"Sz") #ITensors implementation
expect("Sz",b) #ITensorNetworks implementation
```

where the last line raises an `OutOfMemory` error.

PR #129, opened 2024-01-24
Efficient implementation of `expect` for `AbstractTTN`

b-kloss:
This PR addresses #128 and adds an implementation of expect that should allow evaluation of local observables at all sites in a `TTN` at linear cost in the number of vertices in the `TTN`.

Basic tests have been added:
- comparing with `ITensors.expect` evaluated on the output of `random_mps` converted to `ITensors.MPS`, as well as verifying the `Sz`
- expectation value for a product-TTN in the computational basis for a `comb_tree`.

Open issues:
- performance testing should be added at some point / optimality of implementation is not verified
- Currently does not support multiple site-indices per vertex. This could probably easily be addressed similar to the way this is handled in the `svd_ttn` for converting `OpSum` to `TTN`.

### DIFF
--- a/src/treetensornetworks/abstracttreetensornetwork.jl
+++ b/src/treetensornetworks/abstracttreetensornetwork.jl
@@ -440,7 +440,7 @@ function expect(
     @assert isone(length(sites[v]))
     #ToDo: Add compatibility with more than a single index per vertex
     op_v = op(operator, only(sites[v]))
-    res[v] = scalar(dag(state[v]) * apply(op_v, state[v]))
+    res[v] = (dag(state[v]) * apply(op_v, state[v]))[]
   end
   return mapreduce(typeof, promote_type, res).(res)
 end
--- a/src/treetensornetworks/abstracttreetensornetwork.jl
+++ b/src/treetensornetworks/abstracttreetensornetwork.jl
@@ -424,15 +424,17 @@ end
 function expect(
   operator::String,
   state::AbstractTTN;
-  sites=siteinds(state),
-  root_vertex=default_root_vertex(sites), #ToDo: verify that this is a sane default
+  vertices=vertices(state),
+  root_vertex=default_root_vertex(siteinds(state)), #ToDo: verify that this is a sane default
 )
   # ToDo: for performance it may be beneficial to also implement expect! and change the orthogonality center in place
   # assuming that the next algorithmic step can make use of the orthogonality center being moved to a different vertex
-  vertices = reverse(post_order_dfs_vertices(sites, root_vertex)) #ToDo: Verify that this is indeed the correct order for performance
+  sites=siteinds(state)
+  ordered_vertices = reverse(post_order_dfs_vertices(sites, root_vertex)) #ToDo: Verify that this is indeed the correct order for performance
   ElT = promote_itensor_eltype(state)
   res = Dictionary(vertices, Vector{ElT}(undef, length(vertices)))
-  for v in vertices
+  for v in ordered_vertices
+    !(v in vertices) && continue  #only compute expectation values for required vertices
     state = orthogonalize(state, v)
     @assert isone(length(sites[v]))
     Op = ITensors.op(operator, only(sites[v])) #ToDo: Add compatibility with more than a single index per vertex
--- a/src/treetensornetworks/abstracttreetensornetwork.jl
+++ b/src/treetensornetworks/abstracttreetensornetwork.jl
@@ -429,7 +429,7 @@ function expect(
 )
   # ToDo: for performance it may be beneficial to also implement expect! and change the orthogonality center in place
   # assuming that the next algorithmic step can make use of the orthogonality center being moved to a different vertex
-  sites=siteinds(state)
+  sites = siteinds(state)
   ordered_vertices = reverse(post_order_dfs_vertices(sites, root_vertex)) #ToDo: Verify that this is indeed the correct order for performance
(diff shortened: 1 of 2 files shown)


b-kloss:
I'll fix keyword argument signature that makes the solver_tests fail.

b-kloss:
If tests pass this is ready for review.

src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -420,3 +420,25 @@ function inner(
...
+  operator::String,
+  state::AbstractTTN;
+  vertices=vertices(state),
+  root_vertex=default_root_vertex(siteinds(state)), #ToDo: verify that this is a sane default
+)
+  # ToDo: for performance it may be beneficial to also implement expect! and change the orthogonality center in place
+  # assuming that the next algorithmic step can make use of the orthogonality center being moved to a different vertex
+  sites = siteinds(state)
+  ordered_vertices = reverse(post_order_dfs_vertices(sites, root_vertex)) #ToDo: Verify that this is indeed the correct order for performance
+  ElT = promote_itensor_eltype(state)
```
mtfishman, line 434: I think you can use `scalartype(state)` here (same thing with a shorter name, and more generic).

src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -420,3 +420,25 @@ function inner(
...
+  res = Dictionary(vertices, Vector{ElT}(undef, length(vertices)))
+  for v in ordered_vertices
+    !(v in vertices) && continue  #only compute expectation values for required vertices
+    state = orthogonalize(state, v)
+    @assert isone(length(sites[v]))
+    Op = ITensors.op(operator, only(sites[v])) #ToDo: Add compatibility with more than a single index per vertex
```
mtfishman, line 440:
```suggestion
    Op = op(operator, only(sites[v])) #ToDo: Add compatibility with more than a single index per vertex
```
Does that work? Also I would change `Op` to `op_v`, since we generally favor lower case variable names.

mtfishman: Also I think it is better to put the comment above the line of code, not appended to the end.

src/treetensornetworks/abstracttreetensornetwork.jl
mtfishman: Another thing is that the operator could have a different element type than the state, so this code will fail in that case. We could build the operator on each site in `vertices` ahead of time, store that in a dictionary, and then extract the element type of the operator, and determine the output element type by promoting the operator element type and the state element type to a common element type.

b-kloss: I was going to store the expectation values in an untyped (for the value) dictionary and determine the element type after the loop and create a new typed dictionary from the values and keys of the untyped one.
Should I pay attention to replacing usage of `psi` `operator` etc throughout the whole file (and similarly reflect style changes in src/expect.jl)?

mtfishman: That seems like a good approach, you can use this pattern for that:
```julia
julia> d = Dictionary(["a", "b", "c"], undef)
3-element Dictionary{String, Any}
 "a" │ #undef
 "b" │ #undef
 "c" │ #undef

julia> d["a"] = 1.0
1.0

julia> d["b"] = 2.0
2.0

julia> d["c"] = 3.0
3.0

julia> mapreduce(typeof, promote_type, d).(d)
3-element Dictionary{String, Float64}
 "a" │ 1.0
 "b" │ 2.0
 "c" │ 3.0
```
I think in this PR we shouldn't change the style of the entire file, its generally better to keep PRs more narrowly focused in scope (either add a new feature and use the best style within that new code, or make a PR that improves the style across a lot of code).

mtfishman:
Looks good, thanks!

test/test_treetensornetworks/test_expect.jl
```diff
@@ -0,0 +1,32 @@
...
+  s = siteinds("S=1/2", c)
+  d = Dict()
+  magnetization = Dict()
+  for (i, v) in enumerate(vertices(s))
+    d[v] = isodd(i) ? "Up" : "Dn"
+    magnetization[v] = isodd(i) ? 0.5 : -0.5
+  end
+  states = v -> d[v]
+  state = TTN(s, states)
+  res = expect("Sz", state)   # just testing that this doesn't go via expect for arbitrary tensor networks.
```
mtfishman, line 28:
```suggestion
  res = expect("Sz", state)
```

src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -420,3 +420,27 @@ function inner(
...
+  sites = siteinds(state)
+  ordered_vertices = reverse(post_order_dfs_vertices(sites, root_vertex))
+  res = Dictionary(vertices, undef)
+  for v in ordered_vertices
+    !(v in vertices) && continue
+    state = orthogonalize(state, v)
+    @assert isone(length(sites[v]))
+    #ToDo: Add compatibility with more than a single index per vertex
+    op_v = op(operator, only(sites[v]))
+    res[v] = scalar(dag(state[v]) * apply(op_v, state[v]))
```
mtfishman, line 443:
```suggestion
    res[v] = (dag(state[v]) * apply(op_v, state[v]))[]
```

PR #130, opened 2024-01-25
Fill `ITensor` with zeros if empty after contracting with `ProjTTN`.

b-kloss:
This PR ensures that the result of a `ProjTTN` to `ITensor` contraction always results in an ITensor with the same `flux` and the allowed blocks densely filled.
Lack of this edit leads to instability of `KrylovKit.exponentiate` when used on product states of fermionic systems with regions where the action of the projected Hamiltonian on the state vanishes. There may be more appropriate fixes to this at the ITensors level, if this PR is not compatible with planned changes to the ITensors storage.

### DIFF
--- a/src/treetensornetworks/projttns/abstractprojttn.jl
+++ b/src/treetensornetworks/projttns/abstractprojttn.jl
@@ -80,6 +80,10 @@ function contract(P::AbstractProjTTN, v::ITensor)::ITensor
   for it in itensor_map
     Hv *= it
   end
+  if isempty(Hv)
+    Hv = similar(v)
+    Hv .= 0
+  end
   return Hv
 end
 


b-kloss:
This also only works for ProjTTN of an operator with vanishing flux. The cleaner way would be to compute the flux of the environment in order to compute the flux of the tensor resulting from the contraction, allocate the resulting tensor filled with zeros, and in-place add the actual result of the contraction into it (all inside the if clause).

mtfishman:
@b-kloss could you explain this a bit more? Under what situation does `isempty(Hv)` return `true`? I think that would help to determine a fix to use.

b-kloss:
This happens in the very first call to `PH(phi)` in from `KrylovKit.exponentiate` when the action of the projected Hamiltonian on the state vanishes. E.g. in a region of a product state that's inert under the local Hamiltonian, let's say a single domain wall initial condition for a tight-binding model in 1D on an MPS, more than two sites away from the domain-wall.

```
Hv = ITensor ord=2
Dim 1: (dim=1|id=867|"19,20")' <Out>
 1: QN("Nf",10,-1) => 1
Dim 2: (dim=2|id=441|"Fermion,Site,n=20")' <Out>
 1: QN("Nf",0,-1) => 1
 2: QN("Nf",1,-1) => 1
NDTensors.BlockSparse{Float64, Vector{Float64}, 2}
 1×2

storage(Hv) = Float64[]```

mtfishman:
Superseded by #1318, so I'll close this.

PR #131, opened 2024-01-26
Fix mutation behavior of `position`

b-kloss:
This PR makes `position(pos, PH)` not mutate `PH::AbstractProjTTN`. Non-mutating versions of `make_environment` are also implemented for the concrete subtypes. Additionally, `copy` is implemented for the concrete types using `copy_values_keys` on `PH.enviroments` to provide copy safety (not entirely necessary if only in-place version is kept) and the original implementation renamed to `unsafe_copy`.
 A test example verifying copy safety is provided (the original implementation was not copy-safe due to 
andyferris/Dictionaries.jl#98) .

ToDo:
- [x] bring `copy_keys_values` into namespace
- [ ] test whether it makes sense to keep out-of-place version for performance?
- [x] add test for out-place version if kept
- [ ] also modify `ProjTTNSum` ?


src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -133,10 +133,10 @@ function position(
...
-  for e in internal_edges(P)
-    unset!(P.environments, e)
-  end
+  # remove internal edges (out of place)
+  ie = internal_edges(P)
+  newenvskeys = filter(!in(ie), keys(P.environments))
+  P = ProjTTN(pos, P.H, getindices(P.environments, newenvskeys))
   # make all environments surrounding new position
   for e in incident_edges(P)
     make_environment!(P, psi, e)
```
mtfishman, line 142: Seems like we should also make this out-of-place.

b-kloss: l. 142?
I also implemented that, but it's not necessary here, since `getindices` returns a new `Dictionary`. If you prefer, I can push the implementation with `make_environment` (out of place).

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -133,10 +133,10 @@ function position(
...
   # shift position
   P = shift_position(P, pos)
-  # invalidate environments corresponding to internal edges
-  for e in internal_edges(P)
-    unset!(P.environments, e)
-  end
+  # remove internal edges (out of place)
+  ie = internal_edges(P)
+  newenvskeys = filter(!in(ie), keys(P.environments))
+  P = ProjTTN(pos, P.H, getindices(P.environments, newenvskeys))
```
mtfishman, line 139: Could we wrap this into a function, like `invalidate_environments`?

src/treetensornetworks/projttns/abstractprojttn.jl
mtfishman: Also we could split this loop off into a separate function `make_environments`.

mtfishman: Yeah, I think relying on that copying behavior is a bit low-level, so best to rewrite in terms of an out-of-place function.

b-kloss:
Any idea why the test was canceled?
image

mtfishman:
> Any idea why the test was canceled?

No idea, could just be an issue on Github's side. We can try restarting them.

src/treetensornetworks/projttns/abstractprojttn.jl
b-kloss: I implemented both in place and out of place versions. We may want to get rid of the in-place versions or simplify them to make them fully in-place, right now it's a mix.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -131,13 +131,43 @@ end
...
+end
+
+function position!(
+  P::AbstractProjTTN{V}, psi::TTN{V}, pos::Union{Vector{<:V},NamedEdge{V}}
+) where {V}
+  P = shift_position(P, pos)
+  P = invalidate_environments(P)
+  make_environments!(P, psi, e)
+  return P
+end
```
mtfishman, line 147: I think we can remove this, it doesn't seem to be adding much and we could add an in-place version later if needed.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -131,13 +131,43 @@ end
...
+
+function make_environments(P::AbstractProjTTN{V}, psi::TTN{V}) where {V}
+  P = copy(P)
+  for e in incident_edges(P)
+    set!(P.environments, e, make_environment(P, psi, e))
+  end
+  return P
+end
+
+function make_environments!(P::AbstractProjTTN{V}, psi::TTN{V}) where {V}
```
mtfishman, line 170: I would prefer to have just the out-of-place version for now.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -131,13 +131,43 @@ end
...
+  make_environments!(P, psi, e)
+  return P
+end
+
+function invalidate_environments(P::AbstractProjTTN)
   ie = internal_edges(P)
   newenvskeys = filter(!in(ie), keys(P.environments))
-  P = ProjTTN(pos, P.H, getindices(P.environments, newenvskeys))
-  # make all environments surrounding new position
+  P = ProjTTN(P.pos, P.H, getindices(P.environments, newenvskeys))
```
mtfishman, line 152: Something we can do to clean this up a bit and make it more generic to other `AbstractProjTTN` types is define a generic function `set_environments`, for example:
```julia
set_environments(p::AbstractProjTTN, environments) = error("Not implemented")

set_environments(p::ProjTTN, environments) = ProjTTN(p.pos, p.H, environments)
```
I like to define those kind of field setter functions for all of my types, it makes it easier to modify immutable objects. I also like using the package [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) to make it easier to define these kind of field setter functions.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -131,13 +131,43 @@ end
...
+) where {V}
+  P = shift_position(P, pos)
+  P = invalidate_environments(P)
+  make_environments!(P, psi, e)
+  return P
+end
+
+function invalidate_environments(P::AbstractProjTTN)
   ie = internal_edges(P)
   newenvskeys = filter(!in(ie), keys(P.environments))
```
mtfishman, line 151: I think we should define a function `environments(p::AbstractProjTTN)` which accesses `p.environments`, generally I try to define those kinds of field accessor functions instead of accessing fields with the `.` syntax, which I think looks cleaner and helps with writing generic code.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -131,13 +131,43 @@ end
...
+  return P
+end
+
+function invalidate_environment(P::AbstractProjTTN{V}, e::NamedEdge{V}) where {V}
+  newenvskeys = filter(isequal(e), keys(P.environments))
+  P = ProjTTN(P.pos, P.H, getindices(P.environments, newenvskeys))
+  return P
+end
+
+function make_environments(P::AbstractProjTTN{V}, psi::TTN{V}) where {V}
```
mtfishman, line 162:
```suggestion
function make_environments(P::AbstractProjTTN, psi::TTN)
```
I think it is best to be as minimal as possible with type constraints.

src/treetensornetworks/projttns/abstractprojttn.jl
mtfishman: Similarly, I think it would be good to define functions `operator(p::ProjTTN) = p.H` and `position(p::ProjTTN) = p.pos` and use those in the code.

src/treetensornetworks/projttns/projttn.jl
```diff
@@ -55,3 +39,42 @@ function make_environment!(P::ProjTTN{V}, psi::TTN{V}, e::NamedEdge{V})::ITensor
   )
   return env
 end
+
+function make_environment(P::ProjTTN{V}, psi::TTN{V}, e::NamedEdge{V})::ITensor where {V}
```
mtfishman, line 43:
```suggestion
function make_environment(P::ProjTTN, psi::TTN, e::NamedEdge)
```

mtfishman:
Seems like we could change all uses of `::TTN` to `::AbstractTTN` to make it more general.

mtfishman:
I think overall the new code is definitely an improvement, but needs a cleanup and style improvements based on my comments.

src/treetensornetworks/projttns/abstractprojttn.jl
b-kloss: I ended up keeping separate in-place and out-of-place versions. We can remove the in-place ones if you prefer, but I thought they may be handy to keep around just in case there is a performance difference due to the recursive calls of `make_environment(!)` (and to investigate what actually breaks due to the issue in `Dictionaries.jl`).

src/treetensornetworks/projttns/abstractprojttn.jl
b-kloss: I defined setter and getter functions. Since [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) isn't used in ITensorNetworks I think it's better not to introduce it here?

src/treetensornetworks/projttns/projttn.jl
b-kloss: I removed the type restriction in the newly introduced functions, but didn't touch what already existed in that style.

src/ITensorNetworks.jl
```diff
@@ -20,6 +20,7 @@ using IterTools
 using KrylovKit: KrylovKit
 using LinearAlgebra
 using NamedGraphs
+
```
mtfishman, line 23:
```suggestion
```

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -121,25 +126,55 @@ function Base.size(P::AbstractProjTTN)::Tuple{Int,Int}
...
-  P::AbstractProjTTN{V}, psi::TTN{V}, pos::Union{Vector{<:V},NamedEdge{V}}
+  P::AbstractProjTTN{V}, psi::AbstractTTN{V}, pos::Union{Vector{<:V},NamedEdge{V}}
+) where {V}
+  P = shift_position(P, pos)
+  P = invalidate_environments(P)
+  P = make_environments(P, psi)
+  return P
+end
+
+function position!(
```
mtfishman, line 145: I would prefer to remove this version.

src/treetensornetworks/projttns/abstractprojttn.jl
mtfishman: Still seems to be using `P.environments` instead of `environments(P)`.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -121,25 +126,55 @@ function Base.size(P::AbstractProjTTN)::Tuple{Int,Int}
...
+
+function make_environments(P::AbstractProjTTN, psi::AbstractTTN)
   for e in incident_edges(P)
-    make_environment!(P, psi, e)
+    P = make_environment(P, psi, e)
   end
   return P
 end
+
+function make_environments!(P::AbstractProjTTN, psi::AbstractTTN)
```
mtfishman, line 176: I would prefer to remove this in-place version and add it back if needed for performance.

src/treetensornetworks/projttns/projttn.jl
```diff
@@ -20,38 +21,83 @@ function set_nsite(P::ProjTTN, nsite)
...
-    hascommoninds(environment(P, e), psi[src(e)]),
+    hascommoninds(environment(P, e), state[src(e)]),
     "Something went wrong, probably re-orthogonalized this edge in the same direction twice!"
   )
   return env
 end
+
+function make_environment(
+  P::ProjTTN{V}, state::AbstractTTN{V}, e::NamedEdge{V}
+)::ProjTTN{V} where {V}
```
mtfishman, line 72:
```suggestion
function make_environment(
  P::ProjTTN, state::AbstractTTN, e::NamedEdge
)
```
Since `make_environment` and `make_environment!` are both pretty complicated, we should keep just one of them. I would prefer to just keep `make_environment` for now and then make an in-place version if needed for performance.

src/treetensornetworks/projttns/projttn.jl
```diff
@@ -20,38 +21,83 @@ function set_nsite(P::ProjTTN, nsite)
...
-  return ProjTTN(pos, P.H, P.environments)
+  return ProjTTN(pos, operator(P), environments(P))
 end
 
-function make_environment!(P::ProjTTN{V}, psi::TTN{V}, e::NamedEdge{V})::ITensor where {V}
+set_environments(p::ProjTTN, environments) = ProjTTN(pos(p), operator(p), environments)
+function set_environment(p::ProjTTN, edge, env)
+  newenvs = merge(environments(p), Dictionary((edge,), (env,)))
+  return ProjTTN(pos(p), operator(p), newenvs)
+end
```
mtfishman, line 31: I we keep `set_environment!` (which is ok with me) it is better to implement this as:
```julia
set_environment(p::AbstractProjTTN, edge, env) = set_environment!(copy(p), edge, env)
```
to not have two implementations, i.e. it is a good practice to define out-of-place versions of a function int terms of in-place versions by copying the input.

src/treetensornetworks/projttns/projttn.jl
```diff
@@ -20,38 +21,83 @@ function set_nsite(P::ProjTTN, nsite)
...
+set_environment!(p::ProjTTN, edge, env) = set!(environments(p), edge, env)
```
mtfishman, line 32:
```suggestion
function set_environment!(p::ProjTTN, edge, env)
  set!(environments(p), edge, env)
  return p
end
```
I like to use a convention of returning the object being modified from in-place functions.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -1,17 +1,22 @@
 abstract type AbstractProjTTN{V} end
 
+#accessors for fields common to all concrete subtypes
+environments(p::AbstractProjTTN) = p.environments
+operator(p::AbstractProjTTN) = p.operator
+underlying_graph(P::AbstractProjTTN) = underlying_graph(operator(P))
+pos(P::AbstractProjTTN) = P.pos
```
mtfishman, line 7: These should only be defined for the concrete `ProjTTN` type, I don't think they will make sense for general `AbstractProjTTN` which could have very different data structures.

src/ITensorNetworks.jl
```diff
@@ -33,6 +34,7 @@ using TimerOutputs
 
 using DataGraphs: IsUnderlyingGraph, edge_data_type, vertex_data_type
 using Graphs: AbstractEdge, AbstractGraph, Graph, add_edge!
+using NamedGraphs: copy_keys_values
```
mtfishman, line 37: By the way, it looks like the `Dictionaries.jl` copying issue was fixed: https://github.com/andyferris/Dictionaries.jl/pull/136. We'll have to bump support for `Dictionaries` to `0.4` by changing the compat entry here: https://github.com/mtfishman/ITensorNetworks.jl/blob/v0.4.0/Project.toml#L40 in order to use it.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -121,25 +126,55 @@ function Base.size(P::AbstractProjTTN)::Tuple{Int,Int}
...
       plev(i) > 0 && (d *= dim(i))
     end
   end
   return (d, d)
 end
 
 function position(
-  P::AbstractProjTTN{V}, psi::TTN{V}, pos::Union{Vector{<:V},NamedEdge{V}}
+  P::AbstractProjTTN{V}, psi::AbstractTTN{V}, pos::Union{Vector{<:V},NamedEdge{V}}
+) where {V}
```
mtfishman, line 138:
```suggestion
)
```

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -121,25 +126,55 @@ function Base.size(P::AbstractProjTTN)::Tuple{Int,Int}
...
+end
+
+function invalidate_environments(P::AbstractProjTTN)
+  ie = internal_edges(P)
+  newenvskeys = filter(!in(ie), keys(P.environments))
+  P = set_environments(P, getindices(P.environments, newenvskeys))
+  return P
+end
+
+function invalidate_environment(P::AbstractProjTTN, e::NamedEdge)
```
mtfishman, line 163:
```suggestion
function invalidate_environment(P::AbstractProjTTN, e::AbstractEdge)
```
to be more agnostic about what kind of graph it may be.

src/treetensornetworks/projttns/projttn.jl
```diff
@@ -3,15 +3,16 @@ ProjTTN
 """
 struct ProjTTN{V} <: AbstractProjTTN{V}
   pos::Union{Vector{<:V},NamedEdge{V}} # TODO: cleanest way to specify effective Hamiltonian position?
-  H::TTN{V}
+  operator::AbstractTTN{V}
```
mtfishman, line 6: In general it isn't a good idea for performance to put an abstract type as a field. Is `AbstractTTN` needed here? If not I would just put it back to `TTN` and we can generalize it if needed (by making `ProjTTN` parameterized by the type of the `operator` field).

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -44,7 +49,7 @@ end
 
 environment(P::AbstractProjTTN, edge::Pair) = environment(P, edgetype(P)(edge))
 function environment(P::AbstractProjTTN{V}, edge::NamedEdge{V})::ITensor where {V}
```
mtfishman, line 51:
```suggestion
function environment(P::AbstractProjTTN, edge::AbstractEdge)
```

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -1,57 +1,114 @@
 struct ProjTTNApply{V} <: AbstractProjTTN{V}
   pos::Union{Vector{<:V},NamedEdge{V}}
-  psi0::TTN{V}
-  H::TTN{V}
+  init_state::AbstractTTN{V}
+  operator::AbstractTTN{V}
```
mtfishman, line 4: We shouldn't put abstract types as fields, I would revert this and we can generalize if needed in another way.

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -1,57 +1,114 @@
...
 end
 
 function copy(P::ProjTTNApply)
-  return ProjTTNApply(P.pos, copy(P.psi0), copy(P.H), copy(P.environments))
+  return ProjTTNApply(
+    pos(P), copy(init_state(P)), copy(operator(P)), copy_keys_values(environments(P))
+  )
 end
 
+function unsafe_copy(P::ProjTTNApply)
```
mtfishman, line 22: Could you explain this function?

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -1,57 +1,114 @@
...
-  return ProjTTNApply(pos, P.psi0, P.H, P.environments)
+  return ProjTTNApply(pos, init_state(P), operator(P), environments(P))
+end
+
+function set_environments(p::ProjTTNApply, environments)
+  return ProjTTNApply(pos(p), init_state(p), operator(p), environments)
+end
+
+function set_environment(p::ProjTTNApply, edge, env)
+  newenv = merge(p.environments, Dictionary((edge,), (env,)))
```
mtfishman, line 38:
```suggestion
  newenv = merge(environments(p), Dictionary((edge,), (env,)))
```

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -1,57 +1,114 @@
...
   end
   @assert(
-    hascommoninds(environment(P, e), psi[src(e)]),
+    hascommoninds(environment(P, e), state[src(e)]),
     "Something went wrong, probably re-orthogonalized this edge in the same direction twice!"
   )
   return env
 end
+
+function make_environment(
```
mtfishman, line 81: Again, I would prefer to not have both `make_environment` and `make_environment!`.

test/test_treetensornetworks/test_position.jl
```diff
@@ -0,0 +1,73 @@
...
+  PHc = ITensorNetworks.unsafe_copy(PH)
+  PHc = position(PHc, psi, [vs[2], vs[5]])
+  @test keys(environments(PH)) == original_keys
+  @test keys(environments(PHc)) != original_keys
+  # test that position! is itself not copysafe
+  # but that copy-safety is due to use of work around in implementation of copy
+  PHc = ITensorNetworks.unsafe_copy(PH)
+  ITensorNetworks.position!(PHc, psi, [vs[2], vs[5]])
+  @test_broken keys(environments(PH)) == original_keys  # make this a proper test once andyferris/Dictionaries.jl#98 is resolved
+  @test keys(environments(PHc)) != original_keys
```
mtfishman, line 67: I think these tests are pretty complicated. Maybe once we move to using the new `copy` defined in https://github.com/andyferris/Dictionaries.jl/pull/136 we can remove a lot of this? Or maybe just make a more succinct test of some specific behavior?

src/treetensornetworks/projttns/projttn_apply.jl
b-kloss: This is basically only there for testing purposes, and will become the normal copy function with the fix in Dictionaries.jl

test/test_treetensornetworks/test_position.jl
b-kloss: We can reduce it to a single test with the copying behaviour fixed.

src/treetensornetworks/projttns/projttn_apply.jl
b-kloss: Ok. Do you have an opinion of whether the mutating or non mutating version is prefered? With the copying behaviour fixed it's reasonable to have position do a copy of its input projected_operator and call mutating functions from there. But we can also do full non-mutating calls to the end.

b-kloss: Didn't see your comment above that answers this.

mtfishman: It's a good question, and I don't think it's obvious one way or another without benchmarking. But let's lean towards non-mutating versions for now and add back mutation as needed for performance.

src/treetensornetworks/projttns/abstractprojttn.jl
mtfishman: Yeah, that's fine, we can start using it in a future PR. Just pointing it out for future reference since I think it is a good way of thinking about designing types.

src/ITensorNetworks.jl
b-kloss: Could you please bump it in ITensors?

mtfishman: Oh right, I forgot about that. Sure thing.

mtfishman: I updated `NamedGraphs`, `DataGraphs`, `NDTensors`, and `ITensors` to use `Dictionaries` v0.4, and removed `copy_keys_values`.

mtfishman: The new version of `ITensors` is still being registered but should be done soon: https://github.com/JuliaRegistries/General/pull/99896

b-kloss: seems like there's still a conflict with IsApprox in ITensorNetworks (my dev'd ITensors is updated and works)
ERROR: Unsatisfiable requirements detected for package Dictionaries [85a47980]:
 Dictionaries [85a47980] log:
 ├─possible versions are: 0.1.0-0.4.0 or uninstalled
 ├─restricted to versions 0.4 by ITensors [9136182c], leaving only versions: 0.4.0
 │ └─ITensors [9136182c] log:
 │   ├─possible versions are: 0.3.55 or uninstalled
 │   └─ITensors [9136182c] is fixed to version 0.3.55
 └─restricted by compatibility requirements with IsApprox [28f27b66] to versions: 0.2.0-0.3.29 — no versions left
   └─IsApprox [28f27b66] log:
     ├─possible versions are: 0.1.0-0.1.7 or uninstalled
     ├─restricted to versions 0.1 by ITensors [9136182c], leaving only versions: 0.1.0-0.1.7
     │ └─ITensors [9136182c] log: see above
     └─restricted to versions 0.1.7-0.1 by ITensorNetworks [2919e153], leaving only versions: 0.1.7
       └─ITensorNetworks [2919e153] log:
         ├─possible versions are: 0.4.0 or uninstalled
         └─ITensorNetworks [2919e153] is fixed to version 0.4.0

mtfishman: I see, I guess `IsApprox` needs to update its compat entry for `Dictionaries.jl` to support v0.4.

b-kloss: Yes, but ITensors also has IsApprox v0.1 as a dependency --- why is the conflict not showing up for ITensors?

b-kloss: I see, ITensors sets it to v0.1 while ITensorNetworks sets it v0.1.7, so more restrictive. Switching to v0.1 for ITensorNetworks seems to work (assuming it's not breaking something else).

mtfishman: That sounds fine for now, while we wait for a fix for `IsApprox.jl` (https://github.com/jlapeyre/IsApprox.jl/pull/8).

src/treetensornetworks/projttns/projttn_apply.jl
b-kloss: Replaced this by call to inplace with copy.

src/treetensornetworks/projttns/abstractprojttn.jl
b-kloss: Ok. Should `environments`, `operator`, and `pos` exist for `AbstractProjTTN` as error("Not Implemented"), so the developer knows that these accessors should be defined for any concrete type (regardless of what the field names actually are in the concrete types?).
Also in that case, I'd say
`underlying_graph(P::AbstractProjTTN) = underlying_graph(operator(P))`
should be defined only for AbstractProjTTN, unless a concrete type has reason to overwrite it?

mtfishman: Yeah, they should error as you say.

I'm not sure if `underlying_graph(P::AbstractProjTTN) = underlying_graph(operator(P))` makes sense as a generic definition. For example, if it is a sum of operators, what would `operator` return?

b-kloss: Ok, I'll treat it like the others.

mtfishman:
Looks good, thanks @b-kloss. This kind of cleanup/refactor will help when generalizing to loopy graphs with BP (@JoeyT1994).

PR #132, opened 2024-02-01
Generalize `ProjTTNSum`, replace `ProjTTNApply` with `ProjOuterProdTTN`

b-kloss:
This PR intends to generalize the design of `ProjTTNSum` to be `Vectors` of an arbitrary concrete subtype of AbstractProjTTN.
Changes:
- Made `ProjTTNSum` subtype of `AbstractProjTTN` and and represent a sum arbitrary `AbstractProjTTN`
- Removed type restrictions on `position`, in line with recent changes to `position` for AbstractProjTTN.
- Renamed `ProjTTNApply` to `ProjOuterProdTTN` and redefine functionality to be in line with that of `ProjTTN` for an operator `O` of the form `O=A|v><v|A^dagger`. The prior functionality is accessible via `contract_ket`, which is used e.g. in `contract_updater`, the backend for variational compression of `|x> = \sum_i A_i |b_i>`.

ToDo:
- [x] implement tests for sum of `ProjTTNApply`
- [x] generalize contract solver
- [x] discuss whether the parametrization is sensible
- [x] discuss problematic constructors
- [x] define field accessor functions instead of accessing fields directly 
- [x] remove outdated comments


src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,17 +1,25 @@
 """
 ProjTTNSum
 """
-struct ProjTTNSum{V}
-  pm::Vector{ProjTTN{V}}
-  function ProjTTNSum(pm::Vector{ProjTTN{V}}) where {V}
-    return new{V}(pm)
+struct ProjTTNSum{T<:AbstractProjTTN}
+  pm::Vector{T} where {T}
```
mtfishman, line 5:
```suggestion
  pm::Vector{T}
```
You shouldn't need the `where {T}` inside the definition (in fact I don't even know what that does). The way I wrote it, Julia infers that the `T` in the field definition and top level type definition are the same.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,17 +1,25 @@
...
+  function ProjTTNSum(pm::Vector{T}) where {T}
+    return new{T}(pm)
```
mtfishman, line 7:
```suggestion
  function ProjTTNSum(pm::Vector)
    return new{eltype(pm)}(pm)
```
I generally prefer that style, using functions to access type parameters instead of `where` when possible.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,17 +1,25 @@
...
   end
 end
+#ToDo: define accessor functions
 
 copy(P::ProjTTNSum) = ProjTTNSum(copy.(P.pm))
 
+# The following constructors don't generalize well, maybe require to pass AbstractProjTTN instead of TTN and remove these?
```
mtfishman, line 14: It seems reasonable to have a simple constructor that assumes it will be `ProjTTN`.

src/treetensornetworks/projttns/projttnsum.jl
mtfishman: Also maybe this should be constrained to accept `AbstractProjTTN`, i.e.:
```julia
  function ProjTTNSum(pm::Vector{<:AbstractProjTTN})
    # ...
  end
```

src/treetensornetworks/projttns/projttnsum.jl
mtfishman: Another suggestions would be to call this field `terms` instead of `pm`, and then have an accessor function `terms(projected_operator::ProjTTNSum) = projected_operator.terms`. That's a convention I've used for types that represent lazy sums of objects.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,17 +1,25 @@
...
+    return new{T}(pm)
   end
 end
+#ToDo: define accessor functions
 
 copy(P::ProjTTNSum) = ProjTTNSum(copy.(P.pm))
 
+# The following constructors don't generalize well, maybe require to pass AbstractProjTTN instead of TTN and remove these? 
 ProjTTNSum(ttnos::Vector{<:TTN}) = ProjTTNSum([ProjTTN(M) for M in ttnos])
+function ProjTTNSum(init_states::Vector{<:TTN}, ttnos::Vector{<:TTN})
```
mtfishman, line 16: It makes more sense to me to just use the constructor `ProjTTNSum(terms::Vector{<:AbstractProjTTN})` and pass it `Vector{ProjTTNApply}` instead of defining a specialized constructor.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,17 +1,25 @@
...
+  return ProjTTNSum([
+    ProjTTNApply(state, operator) for (state, operator) in zip(init_states, ttnos)
+  ])
+end
 
+# This constructor can't differentiate between different concrete AbstractProjTTN, remove?
 ProjTTNSum(Ms::TTN{V}...) where {V} = ProjTTNSum([Ms...])
```
mtfishman, line 23:
```suggestion
ProjTTNSum(operators::AbstractTTN...) = ProjTTNSum([operators...])
```

mtfishman:
@b-kloss this looks like a promising approach to supporting more general sums of projected operators.

src/treetensornetworks/projttns/projttnsum.jl
mtfishman: Specifically, two sets of `Vector{<:TTN}` could really mean anything, _a priori_ I wouldn't expect this to construct a sum of `ProjTTNApply` as opposed to a sum of some other type of `AbstractProjTTN`.

src/treetensornetworks/projttns/projttnsum.jl
mtfishman: Though I would actually prefer to get rid of this constructor, and only have constructors that expect `Vector{<:AbstractTTN}` or `Vector{<:AbstractProjTTN}`.

b-kloss: I agree. Do you think it's enough to just have a constructor for `Vector{<:AbstractProjTTN}`? The other one would just default to making `ProjTTN`s, which is also ok.

mtfishman: I think keeping a definition `ProjTTNSum(operators::Vector{<:AbstractTTN}) = ProjTTNSum(ProjTTN.(operators))` makes sense to me, since (at least to me) it is pretty obvious what it should return.

b-kloss:
If tests pass this is ready for review from my side.

src/treetensornetworks/projttns/projttn.jl
```diff
@@ -68,3 +68,11 @@ function make_environment(P::ProjTTN, state::AbstractTTN, e::AbstractEdge)
   )
   return P
 end
+
+function contract(P::ProjTTN)::ITensor
```
mtfishman, line 72:
```suggestion
function contract(P::ProjTTN)
```
Let's remove the other type assertions as well.

Also, why isn't this implemented?

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,67 @@
...
 
-on_edge(P::ProjTTNSum) = on_edge(P.pm[1])
+on_edge(P::ProjTTNSum) = on_edge(terms(P)[1])
 
-nsite(P::ProjTTNSum) = nsite(P.pm[1])
+nsite(P::ProjTTNSum) = nsite(terms(P)[1])
 
 function set_nsite(Ps::ProjTTNSum, nsite)
-  return ProjTTNSum(map(M -> set_nsite(M, nsite), Ps.pm))
+  return ProjTTNSum(map(M -> set_nsite(M, nsite), Ps.terms))
```
mtfishman, line 23:
```suggestion
  return ProjTTNSum(map(p -> set_nsite(p, nsite), terms(Ps)))
```

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -105,6 +109,10 @@ function product(P::AbstractProjTTN, v::ITensor)::ITensor
   return noprime(Pv)
 end
 
+function product(P::AbstractProjTTN)
+  return error("Not implemented")
+end
```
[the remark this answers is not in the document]
b-kloss: `product(P)` is self-contained for `P::ProjTTNApply` but not for `P::ProjTTN`. Ideally, the implementation of `product(P,v)` should be invalidated for `P::ProjTTNApply`. Not sure if that's a good way to define an interface though.

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -71,3 +71,34 @@ function make_environment(P::ProjTTNApply, state::AbstractTTN, e::AbstractEdge)
...
+  for it in itensor_map
+    Hv *= it
+  end
+  return Hv
+end
+
+#ToDo: Is this a good idea?
+function product(P::ProjTTNApply)::ITensor
+  return noprime(contract(P::ProjTTNApply))
+end
```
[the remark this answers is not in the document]
b-kloss: The point of this is to make `contract_updater` generic for `ProjTTNApply` and ProjTTNSum (of `ProjTTNApply`'s).

test/test_treetensornetworks/test_solvers/test_contract.jl
```diff
@@ -25,6 +25,14 @@ using Test
   Hpsi = apply(H, psi; alg="fit", init=psi')
   @test inner(psi, Hpsi) ≈ inner(psi', H, psi) atol = 1E-5
 
+  # Test basic usage for use with multiple ProjTTNApply with default parameters
+  # BLAS.axpy-like test
+  os_id = OpSum()
+  os_id += -1, "Id", 1, "Id", 2
+  minus_identity = mpo(os_id, s)
+  Hpsi = apply([H, minus_identity], [psi, copy(psi)]; alg="fit", init=psi')
```
mtfishman, line 33: I don't think `apply` is the right thing to call this, I was pretty confused about what this meant.

For an API for summing a product of networks, I think it makes more sense to call this:
```julia
sum([(H, psi), (minus_identity, psi)]; kwargs...)
```
i.e. group by which networks are being contracted.

For the `apply` version, I don't think an API is obvious. An experimental thing I was trying out in `ITensors.jl` was using upper case function names to represent lazy function applications, so:
```julia
sum([Apply(H, psi), Apply(minus_identity, psi)]; kwargs...)
```
Otherwise we could go with:
```julia
sum_apply([(H, psi), (minus_identity, psi)]; kwargs...)
```
i.e. denoting it is a sum of applications of operators onto states.

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -71,3 +71,34 @@ function make_environment(P::ProjTTNApply, state::AbstractTTN, e::AbstractEdge)
   )
   return P
 end
+
+function contract(P::ProjTTNApply)::ITensor
```
mtfishman, line 75: I'm confused, shouldn't something like this have already been defined if `ProjTTNApply` was working before?

b-kloss: The only place `ProjTTNApply` was used was in `contract_updater`. There, the argument `v` in `contract(P,v)` was constructed by accessing init_state(P). Now, this is done inside contract(P::ProjTTNApply).

mtfishman: But why not keep it as it was? What does this new design improve on?

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -65,27 +65,15 @@ function _separate_first_two(V::Vector)
   return frst, scnd, rst
 end
 
+projected_operator_tensors(P::AbstractProjTTN) = error("Not implemented.")
+
+#contract(P::AbstractProjTTN, v::ITensor)::ITensor = foldl(*,projected_operator_tensors(P);init=v)
 function contract(P::AbstractProjTTN, v::ITensor)::ITensor
```
mtfishman, line 71:
```suggestion
function contract(P::AbstractProjTTN, v::ITensor)
```
and same thing elsewhere.

src/treetensornetworks/projttns/projttn.jl
```diff
@@ -68,3 +68,21 @@ function make_environment(P::ProjTTN, state::AbstractTTN, e::AbstractEdge)
...
+  # two environments, then TTN tensor, then other environments
+  if on_edge(P)
+    itensor_map = environments
+  else
+    itensor_map = Union{ITensor,OneITensor}[] # TODO: will a Hamiltonian TTN tensor ever be a OneITensor?
+    for s in sites(P)
+      site_envs = filter(hascommoninds(operator(P)[s]), environments)
+      frst, scnd, rst = _separate_first_two(site_envs)
+      site_tensors = vcat(frst, scnd, operator(P)[s], rst)
+      append!(itensor_map, site_tensors)
```
mtfishman, line 84: Do you know what all of this is doing? It is pretty hard to understand. I thought this code would just need to loop through `sites(P)` and grab the operator on that site, i.e. `operator(P)[s]`.

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -71,3 +71,28 @@ function make_environment(P::ProjTTNApply, state::AbstractTTN, e::AbstractEdge)
...
+  if on_edge(P)
+    append!(itensor_map, environments)
+  else
+    for s in sites(P)
+      site_envs = filter(hascommoninds(operator(P)[s]), environments)
+      frst, scnd, rst = _separate_first_two(site_envs)
+      site_tensors = vcat(frst, scnd, operator(P)[s], rst)
+      append!(itensor_map, site_tensors)
+    end
+  end
```
mtfishman, line 92: Similar comment as above, I'm confused why this code is so complicated.

test/test_treetensornetworks/test_solvers/test_contract.jl
```diff
@@ -65,6 +73,14 @@ end
   Hpsi = apply(H, psi; alg="fit")
   @test inner(psi, Hpsi) ≈ inner(psi', H, psi) atol = 1E-5
 
+  # Test basic usage for multiple ProjTTNApply with default parameters
+  # BLAS.axpy-like test
+  os_id = OpSum()
+  os_id += -1, "Id", vertices(s)[1], "Id", vertices(s)[1]
+  minus_identity = TTN(os_id, s)
+  Hpsi = apply([H, minus_identity], [psi, copy(psi)]; alg="fit", init=psi')
```
mtfishman, line 81: It makes sense to me to require `init` to have primed site indices (`init=psi'`) for `contract`, but for `apply` I would think they should be unprimed.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,67 @@
...
 
-Base.size(P::ProjTTNSum) = size(P.pm[1])
+Base.size(P::ProjTTNSum) = size(terms(P)[1])
 
-function position(
-  P::ProjTTNSum{V}, psi::TTN{V}, pos::Union{Vector{<:V},NamedEdge{V}}
-) where {V}
-  return ProjTTNSum(map(M -> position(M, psi, pos), P.pm))
+#ToDo remove parametrization? 
+function position(P::ProjTTNSum, psi::TTN, pos)
```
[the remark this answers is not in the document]
mtfishman: Also we should use `operator` and `state` for these kinds of variable names.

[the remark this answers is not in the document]
mtfishman: Or I guess `P` is ok since it is a projected operator and that seems to be what is used throughout the code (though I wish we used a convention `proj_operator` or something like that).

src/treetensornetworks/projttns/projttn.jl
b-kloss: Yes, I think it's a heuristic hardcoded contraction order. If the optimal contraction order was determined afterwards, the code could look simpler.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,67 @@
...
 function product(P::ProjTTNSum, v::ITensor)::ITensor
-  Pv = product(P.pm[1], v)
-  for n in 2:length(P.pm)
-    Pv += product(P.pm[n], v)
+  Pv = product(terms(P)[1], v)
+  for n in 2:length(terms(P))
+    Pv += product(terms(P)[n], v)
+  end
+  return Pv
+end
```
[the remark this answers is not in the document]
b-kloss: Yes, this one is superfluous.

src/treetensornetworks/projttns/projttn.jl
mtfishman: I see. Since it is a tree tensor network, it seems like a simpler way to write it would be to collect the tensors as a TTN like `region_operator_ttn = TTN([operator(P)[s] for s in sites(P)])` and then determine the order with `post_order_dfs_vertices(region_operator_ttn)`.

test/test_treetensornetworks/test_solvers/test_contract.jl
b-kloss: I agree with this API-wise, but the `noprime` is only executed after the variational compression. We can retrieve `init` and `noprime` it in the overload of `apply`.

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -71,3 +71,28 @@ function make_environment(P::ProjTTNApply, state::AbstractTTN, e::AbstractEdge)
...
+  return itensor_map
+end
+
+function product(P::ProjTTNApply, v::ITensor)::ITensor
```
mtfishman, line 96:
```suggestion
function product(P::ProjTTNApply, v::ITensor)
```

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,67 @@
...
+
+function contract(P::ProjTTNSum, v::ITensor)::ITensor
+  Pv = contract(terms(P)[1], v)
+  for n in 2:length(terms(P))
+    Pv += contract(terms(P)[n], v)
   end
```
mtfishman, line 48: We should be able to write this more simply as something like:
```julia
Pv = sum(p -> contract(p, v), terms(P))
```

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,67 @@
...
 
 function Base.eltype(P::ProjTTNSum)
-  elT = eltype(P.pm[1])
-  for n in 2:length(P.pm)
-    elT = promote_type(elT, eltype(P.pm[n]))
+  elT = eltype(terms(P)[1])
+  for n in 2:length(terms(P))
+    elT = promote_type(elT, eltype(terms(P)[n]))
   end
   return elT
```
mtfishman, line 57: I think this should work:
```suggestion
  return mapreduce(eltype, promote_type, terms(P))
```

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -33,35 +33,18 @@ incident_edges(P::ProjTTNSum) = incident_edges(terms(P)[1])
...
 internal_edges(P::ProjTTNSum) = internal_edges(terms(P)[1])
 
-function product(P::ProjTTNSum, v::ITensor)::ITensor
-  Pv = product(terms(P)[1], v)
-  for n in 2:length(terms(P))
-    Pv += product(terms(P)[n], v)
-  end
-  return Pv
-end
+product(P::ProjTTNSum, v::ITensor) = noprime(contract(P, v))
```
mtfishman, line 36: Can't this just be removed entirely in favor of the `AbstractProjTTN` version?

src/treetensornetworks/solvers/contract.jl
```diff
@@ -69,23 +70,37 @@ function contract(tn1::AbstractTTN, tn2::AbstractTTN; alg="fit", kwargs...)
...
 )
-  tn12 = contract(tn1s, tn2s; alg="fit", kwargs...)
+  #plin=plev(first(externalinds(init)))
+  #outputindsnet = flatten_external_indsnetwork(first(tns), first(tn2s))
+  #plout=plev(outputindsnet[first(vertices(outputindsnet))])
+  plev_inc = plev_diff(
+    flatten_external_indsnetwork(first(first(tns)), last(first(tns))),
+    external_indsnetwork(init),
+  )
+  init = prime(init, plev_inc)
```
mtfishman, line 96: I think we shouldn't allow arbitrary differences, and just require pairs of 0 and 1.

src/treetensornetworks/solvers/contract.jl
```diff
@@ -69,23 +70,37 @@ function contract(tn1::AbstractTTN, tn2::AbstractTTN; alg="fit", kwargs...)
...
+  alg != "fit" && error("sum_apply not implemented for other algorithms than fit.")
```
mtfishman, line 97: I think we can remove this check, and allow it to fail based on the function not being defined for the given `Algorithm` type.

src/treetensornetworks/solvers/contract.jl
```diff
@@ -1,30 +1,48 @@
 function contract(
   ::Algorithm"fit",
-  tn1::AbstractTTN,
-  tn2::AbstractTTN;
-  init=random_ttn(flatten_external_indsnetwork(tn1, tn2); link_space=trivial_space(tn1)),
+  tns::Vector{<:Tuple{<:AbstractTTN,<:AbstractTTN}};
```
mtfishman, line 3: I think we should call this `sum_contract`.

test/test_treetensornetworks/test_solvers/test_contract.jl
b-kloss: I went with the last option
```
sum_apply([(H, psi), (minus_identity, psi)]; kwargs...)
```

src/treetensornetworks/projttns/projttn.jl
b-kloss: I suggest we do this in another PR since it is not exactly related to ProjTTNSum?

src/treetensornetworks/projttns/projttn_apply.jl
```diff
@@ -71,3 +71,28 @@ function make_environment(P::ProjTTNApply, state::AbstractTTN, e::AbstractEdge)
...
+  return noprime(contract(P, v))
+end
```
mtfishman, line 98: Can we delete this in favor of the generic `AbstractProjTTN` version?

src/treetensornetworks/solvers/contract.jl
```diff
@@ -1,30 +1,48 @@
...
+  init,
+  #init=random_ttn(
+  #  flatten_external_indsnetwork(first(tn1s), first(tn2s));
+  #  link_space=trivial_space(first(tn1s)),
+  #),
```
mtfishman, line 8: Reminder to delete this.

src/treetensornetworks/projttns/projttnsum.jl
b-kloss: `ProjTTNSum` is not a subtype of `AbstractProjTTN` if I am not mistaken.

src/treetensornetworks/solvers/contract.jl
```diff
@@ -1,30 +1,48 @@
...
   nsweeps=1,
```
mtfishman, line 9: I think we shouldn't be setting a default for this.

src/treetensornetworks/solvers/contract.jl
```diff
@@ -1,30 +1,48 @@
...
   if n == 1
-    v = only(vertices(tn2))
-    return typeof(tn2)([tn1[v] * tn2[v]])
+    res = 0
+    for (tn1, tn2) in zip(tn1s, tn2s)
+      v = only(vertices(tn2))
+      res += tn1[v] * tn2[v]
+    end
+    typeof([res])
+    return nothing
```
mtfishman, line 31:
```suggestion
    return typeof(tn2)([res])
```
I assume.

src/treetensornetworks/projttns/projttnsum.jl
b-kloss:
```
julia> supertypes(ITensorNetworks.ProjTTNSum)
(ProjTTNSum, Any)

julia> supertypes(ITensorNetworks.ProjTTN)
(ProjTTN, ITensorNetworks.AbstractProjTTN, Any)
```

src/treetensornetworks/solvers/contract.jl
b-kloss: I restricted the difference to be 1 and raise an error otherwise.

src/treetensornetworks/projttns/projttn_apply.jl
b-kloss: yes

src/treetensornetworks/solvers/contract.jl
```diff
@@ -43,6 +59,10 @@ function contract(
   return psi
 end
 
+function contract(a::Algorithm"fit", tn1::AbstractTTN, tn2::AbstractTTN; kwargs...)
+  return contract(a, [(tn1, tn2)]; kwargs...)
```
mtfishman, line 63:
```suggestion
  return sum_contract(a, [(tn1, tn2)]; kwargs...)
```

src/treetensornetworks/solvers/contract.jl
b-kloss: Yes, thanks for catching this.

src/treetensornetworks/projttns/projttn.jl
mtfishman: Sure, makes sense.

src/treetensornetworks/projttns/projttnsum.jl
mtfishman: Oh that is strange, seems like it should be.

src/treetensornetworks/projttns/projttn.jl
mtfishman: I raised an issue about it here: #133.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -65,27 +65,15 @@ function _separate_first_two(V::Vector)
   return frst, scnd, rst
 end
 
+projected_operator_tensors(P::AbstractProjTTN) = error("Not implemented.")
+
+#contract(P::AbstractProjTTN, v::ITensor)::ITensor = foldl(*,projected_operator_tensors(P);init=v)
```
mtfishman, line 70:
```suggestion
# contract(P::AbstractProjTTN, v::ITensor)= foldl(*, projected_operator_tensors(P); init=v)
```
I would be ok with using this as the implementation.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,50 @@
...
-function product(P::ProjTTNSum, v::ITensor)::ITensor
-  Pv = product(P.pm[1], v)
-  for n in 2:length(P.pm)
-    Pv += product(P.pm[n], v)
-  end
-  return Pv
-end
+product(P::ProjTTNSum, v::ITensor) = noprime(contract(P, v))
+
+contract(P::ProjTTNSum, v::ITensor)::ITensor = sum(p -> contract(p, v), terms(P))
```
mtfishman, line 38:
```suggestion
contract(P::ProjTTNSum, v::ITensor) = sum(p -> contract(p, v), terms(P))
```

src/treetensornetworks/projttns/projouterprodttn.jl
```diff
@@ -0,0 +1,115 @@
...
+    vertextype(operator)[],
+    internal_state,
+    operator,
+    Dictionary{edgetype(operator),ITensor}(),
+  )
+end
+
+function copy(P::ProjOuterProdTTN)
+  return ProjOuterProdTTN(
+    P.pos, copy(internal_state(P)), copy(operator(P)), copy(environments(P))
```
mtfishman, line 25:
```suggestion
    copy(pos(P)), copy(internal_state(P)), copy(operator(P)), copy(environments(P))
```

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,51 @@
 """
 ProjTTNSum
 """
-struct ProjTTNSum{V}
-  pm::Vector{ProjTTN{V}}
-  function ProjTTNSum(pm::Vector{ProjTTN{V}}) where {V}
-    return new{V}(pm)
+struct ProjTTNSum{V} <: AbstractProjTTN{V}
+  terms::Vector{AbstractProjTTN{V}}
```
mtfishman, line 5:
```suggestion
struct ProjTTNSum{T<:AbstractProjTTN{V}} <: AbstractProjTTN{V}
  terms::Vector{T}
```

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,51 @@
...
+  function ProjTTNSum(terms::Vector{<:AbstractProjTTN{V}}) where {V}
+    return new{V}(terms)
```
mtfishman, line 7:
```suggestion
  function ProjTTNSum(terms::Vector{<:AbstractProjTTN})
    return new{eltype(terms)}(terms)
```

src/treetensornetworks/projttns/projttnsum.jl
mtfishman: This doesn't look to be resolve as I suggested, we shouldn't use an abstract type as a constraint on the field, instead it should be parameterized as I showed.

src/treetensornetworks/projttns/projouterprodttn.jl
```diff
@@ -0,0 +1,113 @@
...
+  for t in itensor_map
+    v *= t
+  end
+  return v
+end
+
+# ToDo: debug this, not yet working
+# probably 
+function contract(P::ProjOuterProdTTN, x::ITensor)
+  return conj(contract_ket(P, dag(x))) * contract_ket(P, ITensor(true))
```
mtfishman, line 112: I realized that it may be more efficient to do:
```julia
  ket = contract_ket(P, ITensor(true)
return (dag(ket) * x) * ket
```
(not sure if I got the conjugation/arrows correct).

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,59 +1,51 @@
...
-struct ProjTTNSum{V}
-  pm::Vector{ProjTTN{V}}
-  function ProjTTNSum(pm::Vector{ProjTTN{V}}) where {V}
-    return new{V}(pm)
+struct ProjTTNSum{V} <: AbstractProjTTN{V}
+  terms::Vector{T} where {T<:AbstractProjTTN{V}}
+  function ProjTTNSum(terms::Vector{T}) where {V,T<:AbstractProjTTN{V}}
+    return new{V}(terms)
   end
 end
```
mtfishman, line 9:
```suggestion
vertextype(::Type{<:AbstractProjTTN{V}}) where {V} = V
vertextype(p::AbstractProjTTN) = typeof(p)

struct ProjTTNSum{V,T<:AbstractProjTTN{V}} <: AbstractProjTTN{V}
  terms::Vector{T}
  function ProjTTNSum(terms::Vector{<:AbstractProjTTN})
    return new{vertextype(eltype(terms)),eltype(terms)}(terms)
  end
end
```
What you wrote would still be seen as a field with an abstract type by the Julia compiler. The way I wrote it, if `T` is a concrete type, then `ProjTTNSum` will be a concrete type. See https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type.

I see part of the problem you were having was the details of how to handle the type parameter `V` of the `AbstractProjTTN`, dealing with nested type parameters like that can be tricky. I think what I wrote above is correct, or at least close to what we should do.

test/test_treetensornetworks/test_solvers/test_contract.jl
```diff
@@ -22,31 +22,44 @@ using Test
...
+  Hpsi_via_dmrg = dmrg(Hfit, psi; updater_kwargs=(; which_eigval=:LR,), nsweeps=1)
+  @test abs(inner(Hpsi_via_dmrg, Hpsi / norm(Hpsi))) ≈ 1 atol = 1E-4
+
+  # Test basic usage for use with multiple ProjOuterProdTTN with default parameters
+  # BLAS.axpy-like test
+  os_id = OpSum()
+  os_id += -1, "Id", 1, "Id", 2
+  minus_identity = mpo(os_id, s)
+  Hpsi = ITensorNetworks.sum_apply(
+    [(H, psi), (minus_identity, copy(psi))]; alg="fit", init=psi, nsweeps=1
```
mtfishman, line 38: Does the second `psi` need to be copied here?

src/treetensornetworks/projttns/projttnsum.jl
mtfishman: Note that `vertextype` is a function defined in `NamedGraphs.jl` so we should overload it from there.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,18 +1,25 @@
...
 
 terms(P::ProjTTNSum) = P.terms
+factors(P::ProjTTNSum) = P.factors
 
 copy(P::ProjTTNSum) = ProjTTNSum(copy.(terms(P)))
 
-ProjTTNSum(operators::Vector{<:AbstractTTN}) = ProjTTNSum(ProjTTN.(operators))
+function ProjTTNSum(operators::Vector{<:AbstractProjTTN})
+  return ProjTTNSum(operators, fill(1, length(operators)))
```
mtfishman, line 18:
```suggestion
  return ProjTTNSum(operators, fill(one(Bool), length(operators)))
```
to be more agnostic about the element type.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -1,18 +1,25 @@
...
+end
+function ProjTTNSum(operators::Vector{<:AbstractTTN})
+  return ProjTTNSum(ProjTTN.(operators), fill(1, length(operators)))
```
mtfishman, line 21:
```suggestion
  return ProjTTNSum(ProjTTN.(operators), fill(one(Bool), length(operators)))
```

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -34,9 +41,15 @@ internal_edges(P::ProjTTNSum) = internal_edges(terms(P)[1])
 
 product(P::ProjTTNSum, v::ITensor) = noprime(contract(P, v))
 
-contract(P::ProjTTNSum, v::ITensor) = sum(p -> contract(p, v), terms(P))
+contract(P::ProjTTNSum, v::ITensor) =
+  mapreduce(+, zip(factors(P), terms(P))) do (f, p)
+    f * contract(p, v)
+  end
```
mtfishman, line 47: Seems like this should get split into a multiline function.

src/treetensornetworks/projttns/projttnsum.jl
```diff
@@ -34,9 +41,15 @@ internal_edges(P::ProjTTNSum) = internal_edges(terms(P)[1])
...
 
-contract_ket(P::ProjTTNSum, v::ITensor) = sum(p -> contract_ket(p, v), terms(P))
+contract_ket(P::ProjTTNSum, v::ITensor) =
+  mapreduce(+, zip(factors(P), terms(P))) do (f, p)
+    f * contract_ket(p, v)
+  end
```
mtfishman, line 52: Same thing, should get split into a multiline function.

test/test_treetensornetworks/test_solvers/test_contract.jl
```diff
@@ -28,17 +28,26 @@ using Test
...
   Hpsi = ITensorNetworks.sum_apply(
-    [(H, psi), (minus_identity, copy(psi))]; alg="fit", init=psi, nsweeps=1
+    [(H, psi), (minus_identity, psi)]; alg="fit", init=psi, nsweeps=1
   )
   @test inner(psi, Hpsi) ≈ (inner(psi', H, psi) - norm(psi)^2) atol = 1E-5
 
+  #Hpsi = ITensorNetworks.dmrg(
+  #  [(H, psi), (minus_identity, psi)]; alg="fit", init=psi, nsweeps=1
+  #)
+  #@test inner(psi, Hpsi) ≈ (inner(psi', H, psi) - norm(psi)^2) atol = 1E-5
```
mtfishman, line 49: Is this not working? If so let's put a comment reminding us to investigate it.

src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -101,6 +101,9 @@ function Base.eltype(P::AbstractProjTTN)::Type
   return ElType
 end
 
+vertextype(::Type{<:AbstractProjTTN{V}}) where {V} = V
+vertextype(p::AbstractProjTTN) = typeof(p)
```
mtfishman, line 105:
```suggestion
vertextype(p::AbstractProjTTN) = vertextype(typeof(p))
```

mtfishman:
Looks good, thanks! Over all a nice improvement which extends the functionality a lot (but also cleans up the code).

PR #136, opened 2024-02-08
Add support for tensor network forms (bilinear and quadratic)

JoeyT1994:
This PR adds the `BilinearForm` and `QuadraticForm` types along with the `AbstractForm` 

We define an `AbstractForm` to be of the structure `<x|A|y>` where `x`, `y` are Tensor Network States and `A` is a Tensor Network Operator. The `AbstractForm` contains the mapping from the vertices of the original states to their names in the `Form`, along with the mapping from the vector space to the dual space.

The `BilinearForm` is the generic case where `x`, `y` and `A` are all distinct and constructors are defined for this case.
The `QuadraticForm` is the case when `x` and `y` are identical and the constructor definitions reflect that. In this case functions like `gradient(qf::QuadraticForm, state_vertices::Vector)` are defined for getting the derivative of the `qf` with respect to the vertices of the original state `x`. 

A short series of tests are included.


src/Forms/quadraticform.jl
```diff
@@ -0,0 +1,70 @@
...
+  (cache!)=nothing,
+  contractor_kwargs...,
+)
+  qf_bra_vertices, qf_ket_vertices = bra_ket_vertices(qf, state_vertices)
+  tn_vertices = setdiff(vertices(qf), vcat(qf_bra_vertices, qf_ket_vertices))
+  reduced_tn, _ = induced_subgraph(tensornetwork(qf), tn_vertices)
+  return contract(reduced_tn, contractor_kwargs...)
+end
+
+function update_qf!(qf::QuadraticForm, state_vertex, state::ITensor)
```
mtfishman, line 64: Maybe just `update!`?

src/Forms/quadraticform.jl
```diff
@@ -0,0 +1,70 @@
...
+  setindex_preserve_graph!(tensornetwork(qf)[ket_map(state_vertex)], state, state_vertex)
+  #Change to be consistent with dual space map and link space map. How can we do this elegantly??
+  return setindex_preserve_graph!(
+    tensornetwork(qf)[bra_map(state_vertex)], prime(dag(state)), state_vertex
```
mtfishman, line 68: I thought the space map was supposed to handle mapping the indices, so we wouldn't hardcode any particular mapping like `prime`.

src/Forms/quadraticform.jl
```diff
@@ -0,0 +1,70 @@
...
+  return QuadraticForm(ket, operator; kwargs...)
+end
+
+function bra_ket_vertices(qf::QuadraticForm, state_vertices::Vector)
+  ket_vertices = [ket_map(qf)[sv] for sv in state_vertices]
+  bra_vertices = [bra_map(qf)[sv] for sv in state_vertices]
+  return bra_vertices, ket_vertices
+end
+
+function gradient(
```
mtfishman, line 45: I think it should be called `derivative`, since it is just taking the derivative with respect to one group of vertices. `gradient` would be computing the derivative for multiple sets of inputs (say, multiple derivatives, one for each input). See for example https://docs.juliahub.com/General/AbstractDifferentiation/stable/user_guide/#AbstractDifferentiation-backends.

src/Forms/quadraticform.jl
```diff
@@ -0,0 +1,70 @@
+struct QuadraticForm{V,KetMap,BraMap,OperatorMap,VectorSpaceMap,LinkSpaceMap} <:
+       AbstractForm{V,KetMap,BraMap,OperatorMap,VectorSpaceMap}
+  tn::ITensorNetwork{V}
+  bra_vs_map::BraMap
+  ket_vs_map::KetMap
+  operator_vs_map::OperatorMap
+  dual_space_map::VectorSpaceMap
+  link_space_map::LinkSpaceMap
+end
```
mtfishman, line 9: One idea would be to have `QuadraticForm` wrap a `BilinearForm`, i.e.:
```julia
struct QuadraticForm{Form<:BilinearForm} <: AbstractForm
  form::Form
end

bilinear_form(qf::QuadraticForm) = qf.form
```
Then constructors and function calls could just forward to those of `BilinearForm`.

mtfishman: I think that makes sense mathematically as well since `QuadraticForm` is a special case of `BilinearForm` (kind of like how in Julia's LinearAlgebra library a `Hermitian` matrix is a wrapper around a `Matrix`).

src/Forms/abstractbilinearform.jl
```diff
@@ -0,0 +1,14 @@
+abstract type AbstractForm{V,BraMap,KetMap,OperatorMap,SpaceMap} <:
```
mtfishman, line 1: I don't think there is a need for `AbstractForm` to have so many type parameters. Abstract types generally have fewer parameters than the subtypes, and just try to have ones which are expected to be common across all or most subtypes. We could even just start with it just having `V` to match `AbstractITensorNetwork{V}`.

src/Forms/bilinearform.jl
```diff
@@ -0,0 +1,42 @@
...
+default_operator_vertex_map(v) = (v, "operator")
+default_tno_constructor(s::IndsNetwork) = delta_network(s)
+
+struct BilinearForm{V,KetMap,BraMap,OperatorMap,SpaceMap} <:
+       AbstractForm{V,KetMap,BraMap,OperatorMap,SpaceMap}
+  tn::ITensorNetwork{V}
+  bra_vs_map::BraMap
+  ket_vs_map::KetMap
+  operator_vs_map::OperatorMap
+  dual_space_map::SpaceMap
```
mtfishman, line 13:
```suggestion
  bra_vertex_map::BraMap
  ket_vertex_map::KetMap
  operator_vertex_map::OperatorMap
  dual_index_map::IndexMap
```
In ITensors.jl we use `space` to refer to the part of an Index without the identifiers (the ID, prime levels, and tags).

src/Forms/bilinearform.jl
```diff
@@ -0,0 +1,42 @@
...
+end
+
+#Needed for implementation
+bra_map(blf::BilinearForm) = blf.bra_vs_map
+ket_map(blf::BilinearForm) = blf.ket_vs_map
+operator_map(blf::BilinearForm) = blf.operator_vs_map
+space_map(blf::BilinearForm) = blf.dual_space_map
```
mtfishman, line 20:
```suggestion
bra_vertex_map(blf::BilinearForm) = blf.bra_vs_map
ket_vertex_map(blf::BilinearForm) = blf.ket_vs_map
operator_vertex_map(blf::BilinearForm) = blf.operator_vs_map
dual_index_map(blf::BilinearForm) = blf.dual_index_map
```

src/Forms/bilinearform.jl
```diff
@@ -0,0 +1,42 @@
...
+  bra::ITensorNetwork, operator::ITensorNetwork, ket::ITensorNetwork; kwargs...
+)
+  return Form(bra, operator, ket; form_type="Bilinear", kwargs...)
+end
+
+function BilinearForm(
+  bra::ITensorNetwork,
+  ket::ITensorNetwork;
+  dual_space_map=default_index_map,
+  tno_constructor=default_tno_constructor,
```
mtfishman, line 35:
```suggestion
  operator_constructor=default_operator_constructor,
```

src/Forms/bilinearform.jl
```diff
@@ -0,0 +1,42 @@
...
+
+function BilinearForm(
+  bra::ITensorNetwork, operator::ITensorNetwork, ket::ITensorNetwork; kwargs...
+)
+  return Form(bra, operator, ket; form_type="Bilinear", kwargs...)
+end
+
+function BilinearForm(
+  bra::ITensorNetwork,
+  ket::ITensorNetwork;
```
mtfishman, line 33:
```suggestion
  bra::AbstractITensorNetwork,
  ket::AbstractITensorNetwork;
```
For example to allow `TTN` inputs.

src/Forms/quadraticform.jl
mtfishman: Actually that link isn't so helpful, I can't really make sense of how `AbstractDifferentiation.jl` distinguishes between `derivative` and `gradient`.

For our case, the way I see it using `derivative` could help distinguish the potential ambiguity that:
```julia
gradient(f, [v1, v2, v3])
```
could refer to either:
```julia
[∂ᵥ₁f, ∂ᵥ₂f, ∂ᵥ₃f]
```
i.e. gradient with respect to the three specified vertices, or:
```julia
∂ᵥ₁∂ᵥ₂∂ᵥ₃f
```
i.e. a single 3rd order derivative, which is closer to what we are looking for with this operation. So I was thinking using `derivative(f, [v1, v2, v3])` would make it clearer that we mean the latter, and we could leave `gradient` to represent the former.

src/Forms/quadraticform.jl
mtfishman: Also I would lean towards making this out-of-place at first, i.e. `update(qf::QuadraticForm, state_vertex, state::ITensor)`, with the same implementation with a `copy` in front. Then we can see if an in-place version is needed for performance.

src/Forms/quadraticform.jl
```diff
@@ -0,0 +1,70 @@
...
+  qf::QuadraticForm,
+  state_vertices::Vector;
+  (cache!)=nothing,
+  contractor_kwargs...,
+)
+  qf_bra_vertices, qf_ket_vertices = bra_ket_vertices(qf, state_vertices)
+  tn_vertices = setdiff(vertices(qf), vcat(qf_bra_vertices, qf_ket_vertices))
+  reduced_tn, _ = induced_subgraph(tensornetwork(qf), tn_vertices)
+  return contract(reduced_tn, contractor_kwargs...)
+end
```
mtfishman, line 62: A small refactor which could help with generic code and makes the logic a bit simpler:
```julia
function derivative(f::AbstractForm, state_vertices::Vector; kwargs...)
  tn_vertices = derivative_vertices(f, state_vertices)
  return derivative(tensornetwork(f), tn_vertices; kwargs...)
end

function derivative(f::AbstractForm, state_vertices::Vector; kwargs...)
  bra_vertices, ket_vertices = bra_ket_vertices(f, state_vertices)
  return setdiff(vertices(f), vcat(bra_vertices, ket_vertices))
end

function derivative(tn::AbstractITensorNetwork, vertices::Vector; kwargs...)
  # Implementation of `derivative` for a non-form tensor network, specialized to the algorithm, cache, etc.
end
```
Then `derivative(tn::AbstractITensorNetwork, vertices::Vector; kwargs...)` could be implemented somewhere else in the code, say with `contract`.

src/Forms/quadraticform.jl
```diff
@@ -0,0 +1,70 @@
...
+  operator_space = union_all_inds(s, dual_space_map(s; links=[]))
+  operator = tno_constructor(operator_space)
+  return QuadraticForm(ket, operator; kwargs...)
+end
+
+function bra_ket_vertices(qf::QuadraticForm, state_vertices::Vector)
+  ket_vertices = [ket_map(qf)[sv] for sv in state_vertices]
+  bra_vertices = [bra_map(qf)[sv] for sv in state_vertices]
+  return bra_vertices, ket_vertices
+end
```
mtfishman, line 43: I would split this into two functions, `bra_vertices` and `ket_vertices`. Then maybe define `bra_ket_vertices(f::AbstractForm, state_vertices::Vector) = [bra_vertices(f, state_vertices), ket_vertices(f, state_vertices)]`. Having that functionality in one function that returns a tuple doesn't seem so useful, since it's about as easy to call `bra_vertices` and `ket_vertices` separately if you need those groups of vertices as separate groups.

test/test_forms.jl
```diff
@@ -0,0 +1,31 @@
...
+  g = named_grid((1, 4))
+  s = siteinds("S=1/2", g)
+  s_operator = union_all_inds(s, prime(s; links=[]))
+  χ = 2
+  Random.seed!(1234)
+  ψket = randomITensorNetwork(s; link_space=χ)
+  ψbra = randomITensorNetwork(s; link_space=χ)
+  A = delta_network(s_operator)
+
+  blf = BilinearForm(ψbra, A, ψket)
```
mtfishman, line 19: Maybe the constructor should be `BilinearForm(ψbra, ψket, A)`? Say following a convention that the state (active vertices) go first, then the operator (passive vertices).

src/Forms/quadraticform.jl
JoeyT1994: Yeah it should, I was just writing a quick template. But refactoring it now!

test/test_forms.jl
mtfishman: Another option is `BilinearForm(A, bra, ket)` and `QuadraticForm(A, ket)`, i.e. the operator is passed first. That might make more sense since the operator is more baked into the object, while the states are likely to be modified.

A slightly more exotice interface could be `BilinearForm(A)(bra, ket)` and `QuadraticForm(A)(ket)`, i.e. `BilinearForm(A)` and `QuadraticForm(A)` create functions that takes states, which is closer to the math definitions. Then `BilinearForm()(bra, ket)` and `QuadraticForm()(ket)` could make forms with identity operators.

mtfishman: This leads me to think that we should actually call these types `AbstractFormNetwork`, `BilinearFormNetwork`, and `QuadraticFormNetwork`, since I think they are more specialized than simple bilinear forms and quadratic forms (i.e. they actually define a whole set of bilinear forms and quadratic forms, by removing certain subsets of bra and ket vertices). That opens up the possibility of defining `AbstractForm`, `BilinearForm`, and `QuadraticForm` types which are closer to the original definitions and not particular to wrapping a tensor network.

src/Forms/bilinearform.jl
JoeyT1994: Sounds sensible!

src/Forms/quadraticform.jl
JoeyT1994: Yeah I can see how this makes things cleaner and avoids the need for `construct_form.jl`

src/Forms/quadraticform.jl
JoeyT1994: Okay that makes sense, I have moved to using `derivative` instead

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
+default_bra_vertex_map(v) = (v, "bra")
+default_ket_vertex_map(v) = (v, "ket")
+default_operator_vertex_map(v) = (v, "operator")
+default_operator_constructor(s::IndsNetwork) = delta_network(s)
+
+struct BilinearFormNetwork{V,KetMap,BraMap,OperatorMap} <: AbstractFormNetwork{V}
+  tn::AbstractITensorNetwork{V}
```
mtfishman, line 7:
```suggestion
struct BilinearFormNetwork{V,TensorNetwork<:AbstractITensorNetwork{V}
,BraMap,KetMap,OperatorMap} <: AbstractFormNetwork{V}
  tensornetwork::TensorNetwork
```
We should avoid having fields with abtract types, see https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type.

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
...
+  ket::AbstractITensorNetwork;
+  bra_vertex_map=default_bra_vertex_map,
+  ket_vertex_map=default_ket_vertex_map,
+  operator_vertex_map=default_operator_vertex_map,
+)
+  @assert Set(externalinds(operator)) ==
+    union(Set(externalinds(ket)), Set(externalinds(bra)))
+  @assert isempty(findall(in(internalinds(bra)), internalinds(ket)))
+  @assert isempty(findall(in(internalinds(bra)), internalinds(operator)))
+  @assert isempty(findall(in(internalinds(ket)), internalinds(operator)))
```
mtfishman, line 25: I think we can just remove these checks for now to keep the code simpler, and see in practice what kinds of mistakes are made. My guess is that if there are index issues they can be caught somewhere else, lower down (say in the `ITensorNetwork` constructor, or somewhere in the Krylov solver).

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
...
+
+  bra_renamed = rename_vertices_itn(bra, bra_vertex_map)
+  ket_renamed = rename_vertices_itn(ket, ket_vertex_map)
+  operator_renamed = rename_vertices_itn(operator, operator_vertex_map)
+
+  tn = union(union(bra_renamed, operator_renamed), ket_renamed)
```
mtfishman, line 31:
```suggestion
  # TODO: Reminder to fix `union` so that `union(bra_renamed, operator_renamed, ket_renamed)` works.
  tn = union(union(bra_renamed, operator_renamed), ket_renamed)
```

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
...
+  ket_vertex_map=default_ket_vertex_map,
+  operator_vertex_map=default_operator_vertex_map,
+)
+  @assert Set(externalinds(operator)) ==
+    union(Set(externalinds(ket)), Set(externalinds(bra)))
+  @assert isempty(findall(in(internalinds(bra)), internalinds(ket)))
+  @assert isempty(findall(in(internalinds(bra)), internalinds(operator)))
+  @assert isempty(findall(in(internalinds(ket)), internalinds(operator)))
+
+  bra_renamed = rename_vertices_itn(bra, bra_vertex_map)
```
mtfishman, line 27:
```suggestion
  # TODO: Reminder to fix `rename_vertices(::AbstractITensorNetwork)`.
  bra_renamed = rename_vertices_itn(bra, bra_vertex_map)
```

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
+default_bra_vertex_map(v) = (v, "bra")
+default_ket_vertex_map(v) = (v, "ket")
+default_operator_vertex_map(v) = (v, "operator")
+default_operator_constructor(s::IndsNetwork) = delta_network(s)
+
+struct BilinearFormNetwork{V,KetMap,BraMap,OperatorMap} <: AbstractFormNetwork{V}
+  tn::AbstractITensorNetwork{V}
+  bra_vertex_map::BraMap
+  ket_vertex_map::KetMap
+  operator_vertex_map::OperatorMap
```
mtfishman, line 10:
```suggestion
  operator_vertex_map::OperatorMap
  bra_vertex_map::BraMap
  ket_vertex_map::KetMap
```
Let's order the fields in the ordering of the standard constructor call.

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
...
+    bra_vertex_map(blf),
+    ket_vertex_map(blf),
+    operator_vertex_map(blf),
+  )
+end
+
+function BilinearFormNetwork(
+  bra::AbstractITensorNetwork,
+  ket::AbstractITensorNetwork;
+  operator_constructor=default_operator_constructor,
```
mtfishman, line 56: I don't really like making this a keyword argument, it makes it kind of strange that there are two different ways of specifying the operator. I think we can just hard-code a call to `delta_network` inside the function.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,86 @@
+default_index_map = prime
+
+struct QuadraticFormNetwork{V,FormNetwork<:BilinearFormNetwork{V},IndexMap} <:
+       AbstractFormNetwork{V}
+  formnetwork::FormNetwork
+  dual_index_map::IndexMap
```
mtfishman, line 6: I think there should also be an inverse map so you can map back and forth between bra and ket indices. Maybe we can call them `index_map` and `inv_index_map`?

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,86 @@
...
+end
+
+bilinear_formnetwork(qf::QuadraticFormNetwork) = qf.formnetwork
+function QuadraticFormNetwork(
+  operator::AbstractITensorNetwork,
+  bra::AbstractITensorNetwork,
+  ket::AbstractITensorNetwork;
+  dual_index_map=default_index_map,
+  kwargs...,
+)
```
mtfishman, line 16: Do we need this constructor? I was thinking we would just have `QuadraticFormNetwork(operator, ket)` and `QuadraticFormNetwork(ket)`.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,86 @@
...
+  bra = map_inds(dual_index_map, dag(ket))
+  return QuadraticFormNetwork(operator, bra, ket; dual_index_map, kwargs...)
+end
+
+function QuadraticFormNetwork(
+  ket::AbstractITensorNetwork;
+  dual_index_map=default_index_map,
+  operator_constructor=default_operator_constructor,
```
mtfishman, line 56: Same comment as for the `BilinearFormNetwork` constructor.

Though maybe we can

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,86 @@
...
+  return QuadraticFormNetwork(operator, ket; kwargs...)
+end
+
+function bra_vertices(qf::QuadraticFormNetwork, state_vertices::Vector)
+  return [bra_vertex_map(qf)(sv) for sv in state_vertices]
+end
+
+function ket_vertices(qf::QuadraticFormNetwork, state_vertices::Vector)
+  return [ket_vertex_map(qf)(sv) for sv in state_vertices]
+end
```
mtfishman, line 71: Can these be `AbstractFormNetwork` definitions? Don't these also apply to `BilinearFormNetwork`?

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
...
+  kwargs...,
+)
+  operator_space = union_all_inds(siteinds(bra), siteinds(ket))
```
mtfishman, line 59: Let's use `inds` instead of `space`.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,86 @@
...
+
+function QuadraticFormNetwork(
+  ket::AbstractITensorNetwork;
+  dual_index_map=default_index_map,
+  operator_constructor=default_operator_constructor,
+  kwargs...,
+)
+  s = siteinds(ket)
+  operator_space = union_all_inds(s, dual_index_map(s; links=[]))
+  operator = operator_constructor(operator_space)
```
mtfishman, line 61: Let's use `inds` instead of `space`.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,86 @@
...
+    vertices(f), vcat(bra_vertices(qf, state_vertices), ket_vertices(qf, state_vertices))
+  )
+end
+
+function update(qf::QuadraticFormNetwork, state_vertex, state::ITensor)
+  qf = copy(qf)
+  state_inds = inds(state)
+  state_dag = replaceinds(dag(state), state_inds, dual_index_map(qf).(state_inds))
+  setindex_preserve_graph!(tensornetwork(qf), state, ket_vertex_map(qf)(state_vertex))
+  setindex_preserve_graph!(tensornetwork(qf), state_dag, bra_vertex_map(qf)(state_vertex))
```
mtfishman, line 84:
```suggestion
  # TODO: Maybe add a check that it really does preserve the graph.
  setindex_preserve_graph!(tensornetwork(qf), state, ket_vertex_map(qf)(state_vertex))
  setindex_preserve_graph!(tensornetwork(qf), state_dag, bra_vertex_map(qf)(state_vertex))
```

src/ITensorNetworks.jl
```diff
@@ -98,6 +98,9 @@ include("renameitensornetwork.jl")
 include("boundarymps.jl")
 include(joinpath("beliefpropagation", "beliefpropagation.jl"))
 include(joinpath("beliefpropagation", "beliefpropagation_schedule.jl"))
+include(joinpath("FormNetworks", "abstractformnetwork.jl"))
+include(joinpath("FormNetworks", "bilinearformnetwork.jl"))
+include(joinpath("FormNetworks", "quadraticformnetwork.jl"))
```
mtfishman, line 103:
```suggestion
include(joinpath("formnetworks", "abstractformnetwork.jl"))
include(joinpath("formnetworks", "bilinearformnetwork.jl"))
include(joinpath("formnetworks", "quadraticformnetwork.jl"))
```

src/Forms/quadraticform.jl
mtfishman: The way I wrote it above, it could be that `bra_ket_vertices` is specialized to the `AbstractFormNetwork` subtype.

src/FormNetworks/quadraticformnetwork.jl
JoeyT1994: Good point it is redundant and a bad idea anyway

src/FormNetworks/quadraticformnetwork.jl
JoeyT1994: Good point, yes they do apply to `BilinearFormNetwork`

src/FormNetworks/bilinearformnetwork.jl
mtfishman: Looks like the field ordering is still bra-ket-operator instead of operator-bra-ket.

JoeyT1994: Huh, convinced I fixed this. Anyway, should be fixed now.

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -0,0 +1,61 @@
...
+  ket_vertex_map=default_ket_vertex_map,
+)
+
+  # TODO: Reminder to fix `rename_vertices(::AbstractITensorNetwork)`.
+  bra_renamed = rename_vertices_itn(bra, bra_vertex_map)
+  ket_renamed = rename_vertices_itn(ket, ket_vertex_map)
+  operator_renamed = rename_vertices_itn(operator, operator_vertex_map)
+
+  # TODO: Reminder to fix `union` so that `union(bra_renamed, operator_renamed, ket_renamed)` works.
+  tn = union(union(bra_renamed, operator_renamed), ket_renamed)
```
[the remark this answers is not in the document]
mtfishman: It doesn't support a general function map of the names, just a vertex name that you will append to the current vertices to define the new vertex names.

I think it could be nice to use that. It would basically just require changing the vertex maps to just strings, say:
```julia
default_bra_vertex_suffix(v) = "bra"
default_ket_vertex_suffix(v) = "ket"
default_operator_vertex_suffix(v) = "operator"
```
then store the suffixes in `BilinearForm` instead of the maps.

src/FormNetworks/abstractformnetwork.jl
```diff
@@ -0,0 +1,36 @@
...
+tensornetwork(f::AbstractFormNetwork) = not_implemented()
+copy(f::AbstractFormNetwork) = not_implemented()
+derivative_vertices(f::AbstractFormNetwork) = not_implemented()
+bra_ket_vertices(f::AbstractFormNetwork, state_vertices::Vector) = not_implemented()
+
+bra(f::AbstractFormNetwork) = induced_subgraph(f, collect(values(bra_vertex_map(f))))
+ket(f::AbstractFormNetwork) = induced_subgraph(f, collect(values(ket_vertex_map(f))))
+function operator(f::AbstractFormNetwork)
+  return induced_subgraph(f, collect(values(operator_vertex_map(f))))
+end
```
mtfishman, line 17: I think these don't make sense any more with the new map design?

If we use the new vertex suffix design I suggest below, these could search the vertices for the vertex suffix.

mtfishman: Also it may be good to design these around functions `bra_vertices(f::AbstractFormNetwork) = filter(v -> last(v) == bra_vertex_suffix(f), vertices(f))` and then define `bra_network(f::AbstractFormNetwork) = induced_subgraph(f, bra_vertices(f))`.

mtfishman: Though I would have thought maybe this should return networks with the vertices mapped back to the original ones (i.e. with the suffix removed).

src/FormNetworks/bilinearformnetwork.jl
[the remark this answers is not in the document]
mtfishman: Unfortunately `disjoint_union` seems to be broken for more than two inputs, i.e.:
```julia
⊔("x" => g1, "y" => g2, "z" => g3)
```
throws an error inside `union`, but that is easily fixed with:
```julia
Base.union(graph1::AbstractNamedGraph, graph2::AbstractNamedGraph, graph3::AbstractNamedGraph, graph_rest::AbstractNamedGraph...) = union(union(graph1, graph2), graph3, graph_rest...)
```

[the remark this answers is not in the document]
mtfishman: In the future we could also support general maps as part of `disjoint_union`, i.e. `⊔((v -> (v, "x")) => g1, (v -> (v, "y")) => g2, (v -> (v, "z")) => g3)` and we could then go back to storing vertex maps instead of vertex suffixes, but suffixes seem sufficient for now.

src/FormNetworks/bilinearformnetwork.jl
mtfishman: Not resolved yet.

src/FormNetworks/bilinearformnetwork.jl
mtfishman: Not resolved yet.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,86 @@
+default_index_map = prime
```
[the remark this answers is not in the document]
mtfishman: Not resolved yet.

src/FormNetworks/bilinearformnetwork.jl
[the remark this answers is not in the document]
mtfishman: Also to preserve some of the current functionality, we could define `bra_vertex_map(f::AbstractForm) = (v -> (v, bra_vertex_suffix(f))`, which will probably be nice to have anyway.

mtfishman:
@JoeyT1994 you marked a number of comments as resolved but then I don't see the corresponding changes made.

JoeyT1994:
@mtfishman I think something has gone out of sync with Github because they are resolved on my local machine. I think it keeps using .jl files from `FormNetworks` before it was renamed to `formnetworks`. Let me try to resolve it

src/FormNetworks/bilinearformnetwork.jl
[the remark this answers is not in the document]
JoeyT1994: Okay nice yeah this looks neater. Funnily the `disjoint_union(suffix => tn::ITensorNetwork)` seems to automatically work. So it is renaming it correctly. But `rename_vertices(tn::ITensorNetwork, map)` doesn't work because `rename_vertices` doesn't map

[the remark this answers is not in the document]
JoeyT1994: Actually `rename_vertices` does wrap because it is defined in `AbstractITensorNetwork.jl` which is why `disjoint_union` automatically works. Hence, I don't know why `rename_vertices_itn` is defined at all. I will delete it.

src/FormNetworks/abstractformnetwork.jl
JoeyT1994: Yeah you're right they don't make sense. The latest commit should have fixed that and reflect the new `suffix` logic.

src/FormNetworks/quadraticformnetwork.jl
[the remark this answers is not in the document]
JoeyT1994: Currently sticking to `prime` and `nonprime` because `mapprime` and `replaceprime` appear to only be defined on `Vector{Index}` types?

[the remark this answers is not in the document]
mtfishman: Yes, right, I forgot about that. `prime` and `noprime` are good for this PR.

src/FormNetworks/bilinearformnetwork.jl
```diff
@@ -1,61 +1,52 @@
...
-  ket::AbstractITensorNetwork;
-  operator_constructor=default_operator_constructor,
-  kwargs...,
+  bra::AbstractITensorNetwork, ket::AbstractITensorNetwork; kwargs...
 )
-  operator_space = union_all_inds(siteinds(bra), siteinds(ket))
-  O = tno_constructor(operator_space)
+  operator_inds = union_all_inds(siteinds(bra), siteinds(ket))
+  O = delta_network(operator_inds)
   return BilinearFormNetwork(bra, O, ket; kwargs...)
```
mtfishman, line 51: I think this should be:
```julia
return BilinearFormNetwork(O, bra, ket; kwargs...)
```

src/utils.jl
```diff
@@ -25,3 +25,12 @@ function line_to_tree(line::Vector)
...
 end
+
+function Base.union(
+  graph1::AbstractNamedGraph,
+  graph2::AbstractNamedGraph,
+  graph3::AbstractNamedGraph,
+  graph_rest::AbstractNamedGraph...,
+)
+  return union(union(graph1, graph2), graph3, graph_rest...)
+end
```
mtfishman, line 36: Could you move this to a PR in `NamedGraphs.jl`?

src/FormNetworks/abstractformnetwork.jl
```diff
@@ -0,0 +1,63 @@
...
+  return rename_vertices(inv_vertex_map(f), first(induced_subgraph(f, bra_vertices(f))))
+end
+function ket(f::AbstractFormNetwork)
+  return rename_vertices(inv_vertex_map(f), first(induced_subgraph(f, ket_vertices(f))))
+end
+function operator(f::AbstractFormNetwork)
+  return rename_vertices(
+    inv_vertex_map(f), first(induced_subgraph(f, operator_vertices(f)))
+  )
+end
```
mtfishman, line 47: Let's call these `bra_network`, `ket_network`, `operator_network`.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,75 @@
...
+end
+
+function QuadraticFormNetwork(
+  ket::AbstractITensorNetwork; dual_index_map=default_index_map, kwargs...
+)
+  s = siteinds(ket)
+  operator_inds = union_all_inds(s, dual_index_map(s; links=[]))
+  operator = delta_network(operator_inds)
+  return QuadraticFormNetwork(operator, ket; kwargs...)
+end
```
mtfishman, line 61: I think it would be simpler if both of these just directly call the corresponding `BilinearFormNetwork` constructor and then wrap that in a `QuadraticFormNetwork`.

src/utils.jl
mtfishman: Thanks, should be available in NamedGraphs v0.1.20 now (we should make that the lower bound of the `[compat]` entry in `Project.toml`).

src/exports.jl
```diff
@@ -71,6 +71,8 @@ export AbstractITensorNetwork,
   mps,
   ortho_center,
   set_ortho_center,
+  QuadraticFormNetwork,
+  BilinearFormNetwork,
```
mtfishman, line 75:
```suggestion
  BilinearFormNetwork,
  QuadraticFormNetwork,
```
to put them in alphabetical order.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,87 @@
...
+  operator_inds = union_all_inds(s, dual_index_map(s; links=[]))
+  operator = delta_network(operator_inds)
+  bra = map_inds(dual_index_map, dag(ket))
+  blf = BilinearFormNetwork(operator, bra, ket; kwargs...)
+  return QuadraticFormNetwork(blf, dual_index_map, dual_inv_index_map)
+end
+
+function bra_ket_vertices(qf::QuadraticFormNetwork, state_vertices::Vector)
+  return vcat(bra_vertices(qf, state_vertices), ket_vertices(qf, state_vertices))
+end
```
mtfishman, line 77: Can't this be a generic `AbstractForm` definition?

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,87 @@
...
+
+function update(qf::QuadraticFormNetwork, state_vertex, state::ITensor)
+  qf = copy(qf)
+  state_inds = inds(state)
+  state_dag = replaceinds(dag(state), state_inds, dual_index_map(qf).(state_inds))
+  # TODO: Maybe add a check that it really does preserve the graph.
+  setindex_preserve_graph!(tensornetwork(qf), state, ket_vertex_map(qf)(state_vertex))
+  setindex_preserve_graph!(tensornetwork(qf), state_dag, bra_vertex_map(qf)(state_vertex))
+  return qf
+end
```
mtfishman, line 87: Can this be designed around an `update(blf::BilinearFormNetwork, ...)`, where `update(qf::QuadraticFormNetwork, ...)` forwards to that?

mtfishman: Seems like the only logic particular to `QuadraticFormNetwork` is mapping the state it it's dual.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,87 @@
...
+  return vcat(bra_vertices(qf, state_vertices), ket_vertices(qf, state_vertices))
+end
+
+function update(qf::QuadraticFormNetwork, state_vertex, state::ITensor)
+  qf = copy(qf)
+  state_inds = inds(state)
+  state_dag = replaceinds(dag(state), state_inds, dual_index_map(qf).(state_inds))
+  # TODO: Maybe add a check that it really does preserve the graph.
+  setindex_preserve_graph!(tensornetwork(qf), state, ket_vertex_map(qf)(state_vertex))
+  setindex_preserve_graph!(tensornetwork(qf), state_dag, bra_vertex_map(qf)(state_vertex))
```
mtfishman, line 85: Seems like one of these should be `state_dag`.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,87 @@
...
+  dual_index_map=default_index_map,
+  dual_inv_index_map=default_inv_index_map,
+  kwargs...,
+)
+  s = siteinds(ket)
+  operator_inds = union_all_inds(s, dual_index_map(s; links=[]))
+  operator = delta_network(operator_inds)
+  bra = map_inds(dual_index_map, dag(ket))
+  blf = BilinearFormNetwork(operator, bra, ket; kwargs...)
+  return QuadraticFormNetwork(blf, dual_index_map, dual_inv_index_map)
```
mtfishman, line 72: I was picturing this could just involve calling out to `blf = BilinearFormNetwork(bra, ket; kwargs...)` and `BilinearFormNetwork` would handle the logic of making the operator.

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,87 @@
...
+  bra::AbstractITensorNetwork,
+  ket::AbstractITensorNetwork;
+  dual_index_map=default_index_map,
+  dual_inv_index_map=default_inv_index_map,
+  kwargs...,
+)
+  return QuadraticFormNetwork(
+    BilinearFormNetwork(operator, bra, ket; kwargs...), dual_index_map, dual_inv_index_map
+  )
+end
```
mtfishman, line 23: Reminder to remove this function.

src/FormNetworks/quadraticformnetwork.jl
JoeyT1994: Line 85 updates with `state_dag` ?

src/FormNetworks/quadraticformnetwork.jl
JoeyT1994: Good point, yeah that can be done.

src/FormNetworks/quadraticformnetwork.jl
JoeyT1994: Yeah it can be

src/FormNetworks/quadraticformnetwork.jl
mtfishman: Huh I guess I just read that wrong, too much code to review haha.

src/FormNetworks/quadraticformnetwork.jl
JoeyT1994: Sure, I made an `update` for `BilinearFormNetwork` now

src/FormNetworks/quadraticformnetwork.jl
```diff
@@ -0,0 +1,66 @@
...
+  return QuadraticFormNetwork(blf, dual_index_map, dual_inv_index_map)
+end
+
+function QuadraticFormNetwork(
+  ket::AbstractITensorNetwork;
+  dual_index_map=default_index_map,
+  dual_inv_index_map=default_inv_index_map,
+  kwargs...,
+)
+  s = siteinds(ket)
```
mtfishman, line 55:
```suggestion
```
Looks like this isn't needed anymore.

JoeyT1994:
I don't think so, at least for now. Just trying to change the folder name, git is very strange about renaming case sensitive things so I think I need to use some extra commands

JoeyT1994:
Okay renaming dealt with

mtfishman:
Looks good. I think this already takes us pretty close to being able to use general tensor networks in `alternating_update`!

PR #138, opened 2024-02-13
Remove `lattices.jl`

JoeyT1994:
Quick PR to remove `src/lattices.jl` as most of the code there is not being used and should be part of `NamedGraphs.jl` anyway.

Calls to `chain_lattice_graph(n::Int64)` have been replaced with `named_path_graph(n)` which already exists in `NamedGraphs.jl`


mtfishman:
Thanks!

PR #139, opened 2024-02-15
Belief propagation cache

JoeyT1994:
This PR combines the `partitionedITensorNetwork` and `message_tensors` being passed around and used in belief propagation into one `bpc::BeliefPropgationCache` containing both. Relevant functionality is added for `bpc` such as getting message tensors, getting environments, updating message tensors (via belief propagation) and updating the tensors in the `partitionedITensorNetwork`.

Thus `src/beliefpropgation/beliefpropagation.jl` has essentially moved to `src/caches/beliefpropagationcache.jl` 

`test/test_beliefpropagation.jl` has been refactored to reflect the new interface.


src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
+default_mt_constructor(inds_e) = ITensor[denseblocks(delta(inds_e))]
```
mtfishman, line 1: I think it is better to avoid using a shorthand like `mt` since I think it will be easy to forget what that means.

Maybe just `default_message_tensor`?

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+default_mt_storage(ptn::PartitionedGraph) = Dict()
```
mtfishman, line 2: Again, best to avoid `mt`.

Maybe `default_cache_storage`?

Also, I would prefer to use `Dictionary()` here instead of `Dict()`, since generally I like that type better (it is faster, has a nice interface for functionality like `map`, etc.).

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+function default_contractor(contract_list::Vector{ITensor}; kwargs...)
+  return contract_exact(contract_list; kwargs...)
+end
+default_contractor_kwargs() = (; normalize=true, contraction_sequence_alg="optimal")
+@traitfn default_bp_niters(g::::(!IsDirected)) = is_tree(g) ? 1 : nothing
+@traitfn function default_bp_niters(g::::IsDirected)
+  return default_bp_niters(undirected_graph(underlying_graph(g)))
+end
+function message_tensor_diff(mt_a::Vector{ITensor}, mt_b::Vector{ITensor})
+  LHS, RHS = contract(mt_a), contract(mt_b)
```
mtfishman, line 12:
```suggestion
  lhs, rhs = contract(mt_a), contract(mt_b)
```
Should use lower case for names of objects that aren't types.

src/caches/beliefpropagationcache.jl
mtfishman: Or maybe `default_message_tensors`? Since I see that inside `BeliefPropagationCache` the field storing the message tensors is called `message_tensors` which I think is a good name.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
+default_mt_constructor(inds_e) = ITensor[denseblocks(delta(inds_e))]
+default_mt_storage(ptn::PartitionedGraph) = Dict()
+function default_contractor(contract_list::Vector{ITensor}; kwargs...)
```
mtfishman, line 3: Maybe `default_bp_contract` or `default_bp_update`?

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+         norm((denseblocks(LHS) / sum(diag(LHS))) - (denseblocks(RHS) / sum(diag(RHS))))
+end
+
+struct BeliefPropagationCache{PTN,MTS}
+  partitioned_itensornetwork::PTN
+  message_tensors::MTS
+end
+
+#Constructors...
+function BeliefPropagationCache(ptn::PartitionedGraph; mt_storage=default_mt_storage)
```
mtfishman, line 23:
```suggestion
function BeliefPropagationCache(ptn::PartitionedGraph; message_tensors=default_message_tensors(ptn))
```
I think there isn't really a need to have that keyword argument accept a function, users can just pass that actual message tensors.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+function message_tensor_diff(mt_a::Vector{ITensor}, mt_b::Vector{ITensor})
+  LHS, RHS = contract(mt_a), contract(mt_b)
+  return 0.5 *
+         norm((denseblocks(LHS) / sum(diag(LHS))) - (denseblocks(RHS) / sum(diag(RHS))))
+end
+
+struct BeliefPropagationCache{PTN,MTS}
+  partitioned_itensornetwork::PTN
+  message_tensors::MTS
+end
```
mtfishman, line 20: After looking at the code below that uses what's currently called the `default_mt_constructor` (would be good to discuss that name), I think it could make sense to include that in the `BeliefPropagationCache`, since I think it is pretty fundamental to how the cache works (i.e. you need to know how to construct the message tensors if one doesn't exist, so in many ways it is tied to the `message_tensors`).

In terms of the name for that field, I don't like using the shorthand `mt` in `default_mt_constructor` because I think it will be easy to forget what that means (mostly I try to avoid shorthands, unless the name becomes really long without a shorthand or the shorthand is very standard, like "MPS" or "BP"). An alternative is `default_message_tensor`, one issue I see with that is it doesn't indicate it is a function constructing the message tensor, but maybe that is ok and it should be obvious it must be a function.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+
+function initial_message_tensor(
+  bpc::BeliefPropagationCache, edge::PartitionEdge; mt_constructor=default_mt_constructor
+)
+  ptn = partitioned_itensornetwork(bpc)
+  src_e_itn = subgraph(ptn, src(edge))
+  dst_e_itn = subgraph(ptn, dst(edge))
+  inds_e = commoninds(src_e_itn, dst_e_itn)
+  return mt_constructor(inds_e)
+end
```
mtfishman, line 50: I think we could just call this `default_message_tensor`.

Maybe we could also shorten this function by defining a function:
```julia
function linkinds(bpc::BeliefPropagationCache, edge::PartitionEdge)
  ptn = partitioned_itensornetwork(bpc)
  src_e_itn = subgraph(ptn, src(edge))
  dst_e_itn = subgraph(ptn, dst(edge))
  return commoninds(src_e_itn, dst_e_itn)
end

function default_message_tensor(
  bpc::BeliefPropagationCache, edge::PartitionEdge
)
  return default_message_tensor(bpc)(linkinds(bpc, edge))
end
```
also making use of my suggestion above to store the `default_message_tensor` constructor inside the `BeliefPropagationCache`.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+
+function message_tensor(
+  bpc::BeliefPropagationCache, edge::PartitionEdge; mt_constructor=default_mt_constructor
+)
+  mts = message_tensors(bpc)
+  return haskey(mts, edge) ? mts[edge] : initial_message_tensor(bpc, edge; mt_constructor)
```
mtfishman, line 56: You can use the function `get` for this kind of code pattern: https://docs.julialang.org/en/v1/base/collections/#Base.get

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+end
+function message_tensors(
+  bpc::BeliefPropagationCache, edges::Vector{PartitionEdge}; kwargs...
```
mtfishman, line 59:
```suggestion
  bpc::BeliefPropagationCache, edges::Vector{<:PartitionEdge}; kwargs...
```
I would be surprised if it was working the way you wrote it, maybe needs a test.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+end
+
+"""
+Compute message tensor as product of incoming mts and local state
+"""
+function updated_message_tensor(
+  bpc::BeliefPropagationCache,
+  edge::PartitionEdge;
+  contractor=default_contractor,
+  contractor_kwargs=default_contractor_kwargs(),
```
mtfishman, line 84: I'm going back and forth about the name for this. I may just prefer the name `contract` and `contract_kwargs`. In general, I think it is better to name functions that are performing an action as a verb (unless they are named after what they are outputting, in which case they would be a noun with the name of the thing they are outputting).

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+  mt_constructor=default_mt_constructor,
+)
+  ptn = partitioned_itensornetwork(bpc)
+  pb_edges = partitionedges(ptn, boundary_edges(ptn, vertices(ptn, src(edge)); dir=:in))
```
mtfishman, line 88: What is `bp` referring to in the name `pb_edges`?

src/caches/beliefpropagationcache.jl
mtfishman: Also, I realized until farther down that I had assumed `default_mt_constructor` would output a single message tensor. Maybe we could call this `default_message` to be more agnostic about the output type, i.e. it could in principle output a single `ITensor`, `Vector{ITensor}`, `ITensorNetwork`, etc. and those could all be considered messages.

src/caches/beliefpropagationcache.jl
mtfishman: Also, maybe just call it `message_diff`. I was confused by the name `message_tensor_diff`, I wasn't sure if `mt_a` and `mt_b` were each messages (where the messages are a set of tensors, for partitioned BP) or a set of messages.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+  return mt_constructor(inds_e)
+end
+
+function message_tensor(
+  bpc::BeliefPropagationCache, edge::PartitionEdge; mt_constructor=default_mt_constructor
+)
+  mts = message_tensors(bpc)
+  return haskey(mts, edge) ? mts[edge] : initial_message_tensor(bpc, edge; mt_constructor)
+end
+function message_tensors(
```
mtfishman, line 58:
```suggestion
function messages(
```

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+  contractor_kwargs=default_contractor_kwargs(),
+  mt_constructor=default_mt_constructor,
+)
+  ptn = partitioned_itensornetwork(bpc)
+  pb_edges = partitionedges(ptn, boundary_edges(ptn, vertices(ptn, src(edge)); dir=:in))
+
+  incoming_messages = [
+    message_tensor(bpc, e_in; mt_constructor) for e_in in setdiff(pb_edges, [reverse(edge)])
+  ]
+  incoming_messages = reduce(vcat, incoming_messages; init=ITensor[])
```
mtfishman, line 93: Probably best to factorize this out into a function `incoming_messages(bpc::BeliefPropagationCache,
  edge::PartitionEdge)`.

Probably also helpful to have a function `incoming_messages(bpc::BeliefPropagationCache, vertex)` which `incoming_messages(bpc::BeliefPropagationCache,
  edge::PartitionEdge)` could make use of.

src/caches/beliefpropagationcache.jl
mtfishman: Another alternative is to be more agnostic about the action it is doing, and just call it `update`. I'm worried that the code is a bit too "zoomed in" on the particular application we have in mind (simple BP and partitioned BP), while missing the bigger picture of the BP algorithm, i.e. that the messages and updates can be pretty arbitrary but the BP algorithm has a common structure. I think that considering that broader context can help us think of better names, code structures, etc.

So for example, I could imagine a function designed like this:
```julia
function update_message(
  bp_cache::BeliefPropagationCache,
  edge::PartitionEdge;
  update=default_update(bp_cache),
  update_kwargs=default_update_kwargs(bp_cache),
)
  vertex = src(edge)
  factor_vertex = factor(bp_cache, vertex) # Not sure if this is the best name, but basically it just the subgraph
  incoming_messages_edge = incoming_messages(bp_cache, edge)
  return update(factor_vertex, incoming_messages_edge; update_kwargs...)
end
```
`factor`, `incoming_message`, and `update` as an interface are agnostic about what BP is being run on, what the message are, and how the updates might be done, but we can make specialized versions of `update` and `factor` for the case of running partitioned BP on a tensor network with messages that are `Vector{ITensor}`.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+)
+  bpc_updated = copy(bpc)
+  mts = message_tensors(bpc_updated)
+  c = 0
+  for e in edges
+    mts[e] = updated_message_tensor(bpc_updated, e; mt_constructor, kwargs...)
+    if compute_norm
+      c += message_tensor_diff(message_tensor(bpc, e; mt_constructor), mts[e])
+    end
+  end
```
mtfishman, line 121: Is this computing the norm or the error? `compute_norm` makes me think it would be computing the norm of the network.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+  return bpc_updated, c / (length(edges))
```
mtfishman, line 122: I think instead of returning the difference like this, we can instead pass it as a keyword argument rapped in a reference, i.e.:
```julia
function update(
  bp_cache::BeliefPropagationCache,
  edges::Vector{<:PartitionEdge};
  (update_diff!)=nothing,
  kwargs...,
)
  # ...
  if !isnothing(update_diff)
    update_diff![] += message_diff(message(bp_cache, edge), messages[edge])
  end
  return bp_cache_updated
end
```
and then it can be called like this:
```julia
diff = Ref(0)
update(bp_cache, edges; (update_diff!)=diff)
```
and then `diff` will store the difference between the updated and original message tensors.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+end
+
+"""
+Do parallel updates between groups of edges of all message tensors
+Currently we send the full message tensor data struct to update for each edge_group. But really we only need the
+mts relevant to that group.
+"""
+function update(
+  bpc::BeliefPropagationCache, edge_groups::Vector{<:Vector{<:PartitionEdge}}; kwargs...
+)
```
mtfishman, line 139: Same suggestion about the syntax for outputting the message update difference using a keyword argument.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+
+  for (state, vertex) in zip(states, vertices)
+    setindex_preserve_graph!(tn, state, vertex)
+  end
+  return bpc
+end
+
+function update(bpc, state, vertex)
+  return update(bpc, ITensor[state], [vertex])
+end
```
mtfishman, line 195: I think we should come up with a different name than `update`, since that is used for updating messages throughout the rest of the code and I think this function could easily be confused for that functionality.

Maybe we could use `update_factors` to use BP terminology. Also, I think it is better to order the arguments as `update_factors(bp_cache::BeliefPropagationCache, vertices, factors)` instead of `update_factors(bp_cache::BeliefPropagationCache, factors, vertices)`, i.e. pass the vertices and then the factors.

mtfishman: Also, I think we could get rid of the version taking one vertex, since it's not clear how to distinguish between a vector of vertices vs. a vertex that uses a vector as the name.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
...
+
+"""
+Get the relevant message tensors coming on to the vertices (of the underlying itensornetwork in the cache)
+"""
+function environment_tensors(bpc::BeliefPropagationCache, verts::Vector)
```
mtfishman, line 200: Maybe `incoming_messages`, as suggested above as part of the message update function.

src/exports.jl
```diff
@@ -71,6 +71,7 @@ export AbstractITensorNetwork,
   mps,
   ortho_center,
   set_ortho_center,
+  BeliefPropagationCache,
```
mtfishman, line 74: Let's not export it for now, if anything I would prefer to not export anything new and reconsider all of the exports we have right now.

src/gauging.jl
```diff
@@ -27,13 +26,17 @@ function vidal_gauge(
     vsrc, vdst = src(e), dst(e)
     ψvsrc, ψvdst = ψ_vidal[vsrc], ψ_vidal[vdst]
 
-    pe = partitionedge(pψψ, (vsrc, 1) => (vdst, 1))
+    pe = partitionedge(partitioned_itensornetwork(bpc), (vsrc, 1) => (vdst, 1))
```
mtfishman, line 29: This is a bit unfortunate. Can we make `BeliefPropagationCache` an `AbstractPartitionedGraph` and make these functions work automatically? Or at least forward these kinds of definitions?

src/gauging.jl
```diff
@@ -173,8 +167,9 @@ function symmetric_to_vidal_gauge(
 
   for e in edges(ψ)
     vsrc, vdst = src(e), dst(e)
-    pe = partitionedge(pψψ, NamedEdge((vsrc, 1) => (vdst, 1)))
-    bond_tensors[e], bond_tensors[reverse(e)] = only(mts[pe]), only(mts[pe])
+    pe = partitionedge(partitioned_itensornetwork(bpc), NamedEdge((vsrc, 1) => (vdst, 1)))
```
mtfishman, line 170: Same comment as above about making `partitionedge` work for `BeliefPropagationCache`.

Also, is `NamedEdge` needed here?

mtfishman:
@JoeyT1994 over all I think this is a big improvement, I like basing the BP code around the interface `update(::BeliefPropagationCache)` and I can see that this same kind of interface could extend to other cache types. For example, I could imagine extending this to `update(::ContractionCache)` corresponding to a cache for exact tensor network contraction, where the cache stores the tensor network, a contraction sequence, and a map from a subset of vertices to the result of contracting the subregion of the tensor network.

Then we can define `contract(::ITensorNetwork; (cache!)::AbstractContractionCache)` as a general interface for performing a contraction, where it uses the cache and updates the cache where appropriate (not saying we should do that in this PR, I think the scope of this PR is good as it is).

I've suggested a number of stylistic changes. One broad comment is that I think it would be good to change `bpc` to `bp_cache`, I kept forgetting what `bpc` stands for.

src/gauging.jl
```diff
@@ -73,30 +76,22 @@ function vidal_gauge(
...
+"""Use an ITensorNetwork ψ in the symmetric gauge and its belief propagation cache to put ψ into the vidal gauge. Return the bond tensors and ψ_vidal."""
 function vidal_gauge(
   ψ::ITensorNetwork,
-  pψψ::PartitionedGraph,
-  mts;
+  bpc::BeliefPropagationCache;
   eigen_message_tensor_cutoff=10 * eps(real(scalartype(ψ))),
   regularization=10 * eps(real(scalartype(ψ))),
   edges=NamedGraphs.edges(ψ),
   svd_kwargs...,
```
mtfishman, line 86: It seems like you could just define all of these keyword
```suggestion
  kwargs...
```
and then forward all of them to the next call to `vidal_gauge`.

mtfishman:
I think this definitely moves in the right direction with the gauging code.

An issue I still have with that code is that it's not always clear when a function is just using a BP cache as it is input or if it will run BP to update the cache.

Here is a suggestion for that:
```julia
function vidal_gauge(
  ψ::ITensorNetwork,
  bond_tensors;
  (cache!)=nothing,
  cache_update_kwargs=default_cache_update_kwargs(cache![]),
  message_cutoff=10 * eps(real(scalartype(ψ))),
  regularization=10 * eps(real(scalartype(ψ))),
  edges=NamedGraphs.edges(ψ),
  svd_kwargs...,
)
  if isnothing(cache!)
    cache! = Ref(default_cache(ψ))
  end
  cache![] = update(cache![]; cache_update_kwargs...)
  # Use messages from `cache!` to update to the vidal gauge
end

function vidal_gauge(ψ::ITensorNetwork; kwargs...)
  bond_tensors = initialize_bond_tensors(ψ)
  return vidal_gauge(ψ, bond_tensors; kwargs...)
end
```
(I think we can just change `eigen_message_tensor_cutoff` to `message_cutoff`.) Then the external interface can be:
```julia
# Builds BP cache internally and runs BP with defaults
Γ, Λ = vidal_gauge(ψ)

# Builds BP cache internally and runs BP with 20 iterations
Γ, Λ = vidal_gauge(ψ; cache_update_kwargs=(; niters=20))

# Builds BP cache internally and runs BP with 20 iterations, capture the BP cache for future use
bp_cache_ref = Ref()
Γ, Λ = vidal_gauge(ψ; (cache!)=bp_cache_ref, cache_update_kwargs=(; niters=20))
bp_cache = bp_cache_ref[]

# Build BP cache externally, run BP externally, and pass to `vidal_gauge`. Don't run BP inside `vidal_gauge`
ψψ = norm_network(ψ)
bp_cache = BeliefPropagationCache(ψψ, group(v -> v[1], vertices(ψψ)))
bp_cache = update(bp_cache; niters=20)
# Maybe also have a special keyword argument `update_cache=false`.
Γ, Λ = vidal_gauge(ψ; (cache!)=Ref(bp_cache), cache_update_kwargs=(; niters=0))

# Build BP cache externally and pass to `vidal_gauge`, which runs BP to update the cache.
ψψ = norm_network(ψ)
bp_cache = BeliefPropagationCache(ψψ, group(v -> v[1], vertices(ψψ)))
bp_cache_ref = Ref(bp_cache)
Γ, Λ = vidal_gauge(ψ; (cache!)=bp_cache_ref, cache_update_kwargs=(; niters=20))
bp_cache_updated = bp_cache_ref[]
```

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,221 @@
+default_mt_constructor(inds_e) = ITensor[denseblocks(delta(inds_e))]
+default_mt_storage(ptn::PartitionedGraph) = Dict()
+function default_contractor(contract_list::Vector{ITensor}; kwargs...)
+  return contract_exact(contract_list; kwargs...)
+end
+default_contractor_kwargs() = (; normalize=true, contraction_sequence_alg="optimal")
+@traitfn default_bp_niters(g::::(!IsDirected)) = is_tree(g) ? 1 : nothing
```
mtfishman, line 7: I would suggest changing all uses of `niters` to `maxiter`, since it is really a maximum number of iterations, since it can be stopped short by the `target_precision`.

`maxiter` is used by both `KrylovKit.jl` (https://jutho.github.io/KrylovKit.jl/stable/man/eig/#KrylovKit.eigsolve) and by `IterativeSolvers.jl` (https://iterativesolvers.julialinearalgebra.org/dev/linear_systems/cg/#CG).

Additionally, I would vote for changing `target_precision` to `tol`, which is also a common name for that kind of keyword argument (used in `KrylovKit.jl`, `IterativeSolvers.jl`, and elsewhere).

src/caches/beliefpropagationcache.jl
JoeyT1994: Thinking about this it's probably best to have `boundary_partition_edges(ptn::PartitionedGraph, vertices::Vector{<:PartitionVertex}; dir)` defined in `NamedGraphs`? Then those functions will be fairly simple in the BP code here?

JoeyT1994: Only issue with `incoming_messages(bpc::BeliefPropagationCache, vertex)` is it wouldn't filter out the message tensor incoming from `dst` to `src` and we wouldn't want to have to make a comparison between `messages` to do that

src/caches/beliefpropagationcache.jl
JoeyT1994: `partition_boundary_edges` I think, but best to just spell out in full

src/caches/beliefpropagationcache.jl
mtfishman: Good point about using `incoming_messages(bpc::BeliefPropagationCache, vertex)` directly in `incoming_messages(bpc::BeliefPropagationCache, edge::PartitionEdge)`.

A general `boundary_partition_edges(ptn::PartitionedGraph, vertices::Vector{<:PartitionVertex}; dir)` in `NamedGraphs.jl` makes sense.

src/caches/beliefpropagationcache.jl
JoeyT1994: I guess it is more of an `error` as when it goes to `0` BP is converged. So can rename it like that

src/caches/beliefpropagationcache.jl
JoeyT1994: True although there isn't actually a `partitionedITensorNetwork` type?

src/gauging.jl
JoeyT1994: My only concern with making a `BeliefPropagationCache` and `AbstractPartitionedGraph` was that if, down the line we introduced an `AbstractCache` type and wanted to unify the different cache types under that umbrella

JoeyT1994: At the moment I am not sure what the `AbstractCache` type would look like so maybe that is not a worry?

src/caches/beliefpropagationcache.jl
mtfishman: `setindex!` in Julia has a conventional ordering of `setindex!(object, value, index)`, so I try to follow that argument ordering for `setindex!` and related functions, but with other functions I prefer the ordering `f(object, index, value)` since that is more natural to me, and is also followed by other functions in the ecosystem ([setfield!](https://docs.julialang.org/en/v1/base/base/#Core.setfield!),  [Dictionaries.set!](https://github.com/andyferris/Dictionaries.jl), [Accessors.set](https://juliaobjects.github.io/Accessors.jl/stable/docstrings/#Accessors.set), etc.).

src/caches/beliefpropagationcache.jl
mtfishman: Could be defined for `PartitionedGraph`.

src/caches/beliefpropagationcache.jl
JoeyT1994: We could write something like `incoming_messages(bpc::BeliefPropagationCache, vertices::Vector{<:PartitionVertex}; ignore_edges = PartitionEdge[])` which ignores some subset of edges? This would help unify things I think?

src/gauging.jl
mtfishman: That's a good point. We may not actually need an `AbstractCache` type, and just allow any object to be a cache if you define the right overloads for it. But for now, let's at least define `partitionedge` for `BeliefPropagationCache` by forwarding to `partitioned_itensornetwork` to make this code simpler, and think about the type hierarchy in future PRs.

Related to that, my guess is if we just loosen type constraints in many of the partitioned graph functions to allow them to accept anything as inputs, then a lot of partitioned graph functionality could "just work" as long as some minimal number of functions were defined, even on a non-graph type like the current `BeliefPropagationCache`. So we could think of many of the partitioned graph functionality as more of an interface rather than written specifically for the `PartitionedGraph` type or `AbstractPartitionedGraph` subtypes.

Specifically, say we took all of this `PartitionedGraph` code: https://github.com/mtfishman/NamedGraphs.jl/blob/v0.1.21/src/Graphs/partitionedgraphs/partitionedgraph.jl#L46-L95 and the `AbstractPartitionedGraph` code: https://github.com/mtfishman/NamedGraphs.jl/blob/v0.1.21/src/Graphs/partitionedgraphs/abstractpartitionedgraph.jl (besides the definitions for pure `Graphs.jl`/`AbstractGraph` functionality) and removed the `::PartitionedGraph` and `::AbstractPartitionedGraph` type constraints. Then they could "just work" for `BeliefPropagationCache` without making it an `AbstractPartitionedGraph` subtype, as long as it overloads some minimal interface which makes it a partitioned graph.

src/caches/beliefpropagationcache.jl
mtfishman: That sounds like a good approach.

src/caches/beliefpropagationcache.jl
JoeyT1994: But isn't that strange to have a function `f(a)` which relies on some field in `a` being of a certain type (which the actual struct `a` is not restricted to)?

mtfishman: I think that is more a philosophical question about coding style and generic code design, but in practice I don't see a problem with doing that in this case. The function will just fail farther down if the partitioned graph isn't wrapping an underlying graph which is an `AbstractITensorNetwork`, or something that has the interface of an `AbstractITensorNetwork`. In general I have been leaning towards being less restrictive about typing, and focusing more on function interfaces (i.e. which functions work on an object, rather than which type it is or what it is a subtype of, which is called [duck typing](https://en.wikipedia.org/wiki/Duck_typing)).

mtfishman: We can always add some more restrictions later if needed, for example by defining a `PartitionedITensorNetwork` type or making a trait, but I think that's only needed if there is some ambiguity that arises, for example if there is some other reasonable definition of `linkinds` for a `PartitionedGraph` that is wrapping something else besides an `AbstractITensorNetwork`, which require a specialized implementation not covered by the current implementation.

In general, I've found that it is better to start out being as unrestrictive on type constraints of functions (and fields of types) as possible, which forces you to think about interfaces rather than focusing too much on the particular objects you have in mind, which ultimately makes the code more generic and concise in the long run.

mtfishman: If anything, I would lean towards making the definition even more generic, and define it as:
```julia
function linkinds(tn, edge::PartitionEdge)
  tn_src = subgraph(tn, src(edge))
  tn_dst = subgraph(tn, dst(edge))
  return commoninds(tn_src, tn_dst)
end
```
Then, this function works for any type that defines `subgraph(tn, vertex::PartionVertex)`, which could be `BeliefPropagationCache`, `AbstractITensorNetwork`, or something else.

src/caches/beliefpropagationcache.jl
JoeyT1994: Okay I see, I will go with `index, value` then for `update_factors`

src/caches/beliefpropagationcache.jl
JoeyT1994: Yes I will rewrite it all as `incoming_messages` now

src/caches/beliefpropagationcache.jl
mtfishman: Yeah, that sounds like a good idea (though better to call it `src/partitioneditensornetwork.jl`, i.e. make it lower case).

src/gauging.jl
JoeyT1994: Okay for now I will just forward the relevant functions but I agree with you that it seems a lot of the `PartitionedGraph` code is not really specific to the type and more interface-like. In some future PR we could try loosening things that way and see what happens.

src/gauging.jl
JoeyT1994: Sure. And yeah you're right `NamedEdge` is no longer needed.

src/gauging.jl
```diff
@@ -73,98 +78,66 @@ function vidal_gauge(
...
-    svd_kwargs...,
-  )
+function symmetric_gauge(ψ::ITensorNetwork; kwargs...)
+  ψ_vidal, bond_tensors = vidal_gauge(ψ; kwargs...)
 
   return vidal_to_symmetric_gauge(ψ_vidal, bond_tensors)
 end
 
 """Transform from the Symmetric Gauge (message tensors) to the Vidal Gauge (bond tensors)"""
 function symmetric_to_vidal_gauge(
```
mtfishman, line 138: Is this function needed, or can we just use `vidal_gauge`?

src/gauging.jl
```diff
@@ -73,98 +78,66 @@ function vidal_gauge(
...
-    ψ, pψψ, mts; eigen_message_tensor_cutoff, regularization, svd_kwargs...
-  )
+"""Put an ITensorNetwork into the vidal gauge (by computing the message tensors), return the network and the bond tensors."""
+function vidal_gauge(ψ::ITensorNetwork; kwargs...)
+  bond_tensors = initialize_bond_tensors(ψ)
+  return vidal_gauge(ψ, bond_tensors; kwargs...)
 end
 
 """Transform from an ITensor in the Vidal Gauge (bond tensors) to the Symmetric Gauge (partitionedgraph, message tensors)"""
 function vidal_to_symmetric_gauge(ψ::ITensorNetwork, bond_tensors::DataGraph)
```
mtfishman, line 104: Maybe we can just call this `symmetric_gauge`, and assume that if you pass both an `ITensorNetwork` and a `DataGraph` then it is in the Vidal gauge.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,240 @@
+default_initial_message(inds_e) = ITensor[denseblocks(delta(inds_e))]
+default_messages(ptn::PartitionedGraph) = Dictionary()
+function default_message_update(contract_list::Vector{ITensor}; kwargs...)
+  return contract_exact(contract_list; kwargs...)
+end
+default_message_update_kwargs() = (; normalize=true, contraction_sequence_alg="optimal")
+@traitfn default_bp_maxiters(g::::(!IsDirected)) = is_tree(g) ? 1 : nothing
```
mtfishman, line 7: Let's use `maxiter` instead of `maxiters`, that is more common in other Julia packages.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,240 @@
...
+    end
+  end
+end
+
+function linkinds(bp_cache::BeliefPropagationCache, edge::PartitionEdge)
+  ptn = partitioned_itensornetwork(bp_cache)
+  src_e_itn = subgraph(ptn, src(edge))
+  dst_e_itn = subgraph(ptn, dst(edge))
+  return commoninds(src_e_itn, dst_e_itn)
+end
```
mtfishman, line 66: Is the plan to define `linkinds(partioned_tn::PartitionedGraph, edge::PartitionEdge)` and have this call that?

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,240 @@
...
+  bp_cache::BeliefPropagationCache,
+  edges::Vector{<:PartitionEdge};
+  (update_diff!)=nothing,
+  kwargs...,
+)
+  bp_cache_updated = copy(bp_cache)
+  mts = messages(bp_cache_updated)
+  for e in edges
+    haskey(mts, e) && delete!(mts, e)
+    insert!(mts, e, update_message(bp_cache_updated, e; kwargs...))
```
mtfishman, line 160: I think you can just use `set!` for this.

src/gauging.jl
```diff
@@ -73,98 +78,66 @@ function vidal_gauge(
...
     root_S = sqrt_diag(bond_tensors[e])
     setindex_preserve_graph!(ψsymm, noprime(root_S * ψsymm[vsrc]), vsrc)
     setindex_preserve_graph!(ψsymm, noprime(root_S * ψsymm[vdst]), vdst)
 
-    ψsymm_mts[pe], ψsymm_mts[reverse(pe)] = copy(ITensor[dense(bond_tensors[e])]),
-    copy(ITensor[dense(bond_tensors[e])])
+    haskey(mts, pe) && delete!(mts, pe)
+    haskey(mts, reverse(pe)) && delete!(mts, reverse(pe))
+    insert!(mts, pe, copy(ITensor[dense(bond_tensors[e])]))
+    insert!(mts, reverse(pe), copy(ITensor[dense(bond_tensors[e])]))
```
mtfishman, line 121: I think you can just use `set!` for this.

src/gauging.jl
```diff
@@ -209,22 +183,18 @@ end
 """Function to measure the 'canonicalness' of a state in the Vidal Gauge"""
 function vidal_itn_canonicalness(ψ::ITensorNetwork, bond_tensors::DataGraph)
```
mtfishman, line 184: Maybe we can just make a single function `gauge_error`, i.e. `gauge_error(ψ::ITensorNetwork, bond_tensors::DataGraph)` for the Vidal gauge error and `gauge_error(ψ::ITensorNetwork)` for the symmetric gauge error.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,240 @@
+default_initial_message(inds_e) = ITensor[denseblocks(delta(inds_e))]
```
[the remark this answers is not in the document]
JoeyT1994: This would overwrite `default_message(bp_cache) = bp_cache.default_message`t. In principle it should be called `default_default_message` I guess but that is a bit confusint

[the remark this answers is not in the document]
mtfishman: Can't you restrict the other version to `default_message(bp_cache::BeliefPropagationCache)`?

src/caches/beliefpropagationcache.jl
JoeyT1994: Yes, just adding that now!

src/caches/beliefpropagationcache.jl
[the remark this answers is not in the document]
JoeyT1994: Good point yes I forgot to make that restriction

src/gauging.jl
JoeyT1994: Good point, I think the same applies to `symmetric_to_vidal_gauge` -> `vidal_gauge` as well

src/caches/beliefpropagationcache.jl
JoeyT1994: Okay I have now designed the `update_message` function this way!

src/caches/beliefpropagationcache.jl
JoeyT1994: Good spot, I have changed some internal code to directly call this function now anyway so it is being used.

JoeyT1994:
@mtfishman  thanks for the comments! I have modified the code to take them into account and hopefully it is now more fleshed out and intuitive.
Let me know if you have any more suggestions: the tests will currently fail until my PR [https://github.com/mtfishman/NamedGraphs.jl/pull/51] to `NamedGraphs.jl` is merged as there is a function in there that is needed.

src/ITensorNetworks.jl
```diff
@@ -95,11 +95,12 @@ include("contract.jl")
 include("utility.jl")
 include("specialitensornetworks.jl")
 include("boundarymps.jl")
-include(joinpath("beliefpropagation", "beliefpropagation.jl"))
-include(joinpath("beliefpropagation", "beliefpropagation_schedule.jl"))
+include("partitioneditensornetwork.jl")
+include(joinpath("edge_sequences.jl"))
```
mtfishman, line 99:
```suggestion
include("edge_sequences.jl")
```

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,231 @@
...
+  messages = incoming_messages(bp_cache, partition_verts)
+  central_tensors = ITensor[
+    tensornetwork(bp_cache)[v] for v in setdiff(vertices(bp_cache, partition_verts), verts)
+  ]
+  return vcat(messages, central_tensors)
+end
+
+function incoming_messages(bp_cache::BeliefPropagationCache, vertex)
+  return incoming_messages(bp_cache, [vertex])
+end
```
mtfishman, line 116: This can't distinguish between the case where the vertex name itself is a `Vector`, or if that refers to a collection of vertices.

To fix that, I think we could just get rid of the version `incoming_messages(bp_cache::BeliefPropagationCache, vertex)` and require that it must be a collection of vertices.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,231 @@
...
+  kwargs...,
+)
+  new_mts = copy(messages(bp_cache))
+  for edges in edge_groups
+    bp_cache_t = update(bp_cache, edges; kwargs...)
+    for e in edges
+      new_mts[e] = message(bp_cache_t, e)
+    end
+  end
+  return BeliefPropagationCache(copy(partitioned_itensornetwork(bp_cache)), new_mts)
```
mtfishman, line 183: We could define `set_messages(cache::BeliefPropagationCache, messages) = BeliefPropagationCache(partitioned_itensornetwork(cache), messages, default_message(cache))` as a shorthand for this. In general, I like defining getters and setters for fields of types since it is pretty common to want to update just one field and keep the rest the same. Also, [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) provides nice syntax for implementing field setters.

src/caches/beliefpropagationcache.jl
```diff
@@ -0,0 +1,231 @@
...
+Update the tensornetwork inside the cache
+"""
+function update_factors(
+  bp_cache::BeliefPropagationCache, vertices::Vector, factors::Vector{ITensor}
+)
+  bp_cache = copy(bp_cache)
+  tn = tensornetwork(bp_cache)
+
+  for (vertex, factor) in zip(vertices, factors)
+    setindex_preserve_graph!(tn, factor, vertex)
```
mtfishman, line 224:
```suggestion
    # TODO: Add a check that this preserves the graph structure.
    setindex_preserve_graph!(tn, factor, vertex)
```

src/gauging.jl
```diff
@@ -1,3 +1,9 @@
+function default_cache(ψ::ITensorNetwork)
+  ψψ = norm_network(ψ)
+  return BeliefPropagationCache(ψψ, group(v -> v[1], vertices(ψψ)))
+end
```
mtfishman, line 4: I don't really like that this has such a general name like `default_cache`, but really is making a cache for a norm network. Maybe call it `default_norm_cache`?

src/gauging.jl
```diff
@@ -10,13 +16,12 @@ function initialize_bond_tensors(ψ::ITensorNetwork; index_map=prime)
...
-"""Use an ITensorNetwork ψ, its bond tensors and gauging mts to put ψ into the vidal gauge, return the bond tensors and ψ_vidal."""
+"""Use an ITensorNetwork ψ, its bond tensors and belief propagation cache to put ψ into the vidal gauge, return the bond tensors and ψ_vidal."""
 function vidal_gauge(
   ψ::ITensorNetwork,
-  pψψ::PartitionedGraph,
-  mts,
-  bond_tensors::DataGraph;
-  eigen_message_tensor_cutoff=10 * eps(real(scalartype(ψ))),
+  bond_tensors::DataGraph,
+  bp_cache::BeliefPropagationCache;
```
mtfishman, line 23: I think this should also get moved to a keyword argument, see my proposal here: https://github.com/mtfishman/ITensorNetworks.jl/pull/139#issuecomment-1954553578

src/gauging.jl
```diff
@@ -73,114 +78,58 @@ function vidal_gauge(
...
+
+"""Put an ITensorNetwork into the vidal gauge (by computing the message tensors), return the network and the bond tensors."""
+function vidal_gauge(ψ::ITensorNetwork; kwargs...)
+  bond_tensors = initialize_bond_tensors(ψ)
+  return vidal_gauge(ψ, bond_tensors; kwargs...)
 end
 
 """Transform from an ITensor in the Vidal Gauge (bond tensors) to the Symmetric Gauge (partitionedgraph, message tensors)"""
-function vidal_to_symmetric_gauge(ψ::ITensorNetwork, bond_tensors::DataGraph)
+function symmetric_gauge(ψ::ITensorNetwork, bond_tensors::DataGraph)
```
mtfishman, line 104: I think we should design this function to allow passing a `cache!` keyword argument, as opposed to building it internally and then returning it, to make the interface consistent across the the gauging functions. Ideally all of the gauging functions would be designed that way, and in fact I would imagine many `ITensorNetwork` functions could use that same design, such as the solver code, code for computing expectation values, etc.

src/gauging.jl
```diff
@@ -206,25 +155,19 @@ function vidal_itn_isometries(
   return isometries
 end
 
-"""Function to measure the 'canonicalness' of a state in the Vidal Gauge"""
-function vidal_itn_canonicalness(ψ::ITensorNetwork, bond_tensors::DataGraph)
+"""Function to measure the 'distance' of a state from the Vidal Gauge"""
+function gauge_error(ψ::ITensorNetwork, bond_tensors::DataGraph)
   f = 0
-
   isometries = vidal_itn_isometries(ψ, bond_tensors)
```
mtfishman, line 161: I think `vidal_gauge_isometries` would be a better name for this.

src/gauging.jl
```diff
@@ -206,25 +155,19 @@ function vidal_itn_isometries(
...
-
   for e in keys(isometries)
-    LHS = isometries[e] / sum(diag(isometries[e]))
-    id = dense(delta(inds(LHS)))
-    id /= sum(diag(id))
-    f += 0.5 * norm(id - LHS)
+    lhs = isometries[e]
+    f += message_diff(ITensor[lhs], ITensor[dense(delta(inds(lhs)))])
```
mtfishman, line 164: Probably should be `denseblocks` instead of `dense` to handle QNs properly.

src/gauging.jl
```diff
@@ -206,25 +155,19 @@ function vidal_itn_isometries(
...
 
   return f / (length(keys(isometries)))
 end
 
-"""Function to measure the 'canonicalness' of a state in the Symmetric Gauge"""
-function symmetric_itn_canonicalness(ψ::ITensorNetwork, pψψ::PartitionedGraph, mts)
-  ψ_vidal, bond_tensors = symmetric_to_vidal_gauge(ψ, pψψ, mts)
-
-  return vidal_itn_canonicalness(ψ_vidal, bond_tensors)
+function gauge_error(ψ::ITensorNetwork, bp_cache::BeliefPropagationCache)
```
mtfishman, line 170: I think we should pass the cache as a keyword argument `cache!`.

src/gauging.jl
mtfishman: I don't quite understand your comment, since my comment was on the function called `symmetric_to_vidal_gauge`, but you say:

> it puts an ITensorNetwork into the symmetric gauge

but it looks like this was changed to `vidal_gauge` by now, if I understand the code diffs correctly.

src/gauging.jl
mtfishman: Sorry, scratch that. I think in this case it makes sense to just pass in the Vidal gauge bond tensors, and then output the BP cache for the symmetric gauge, since in some sense the Vidal gauge bond tensors are playing the role of the cache. This points towards needing a better system for dealing with the Vidal gauge (i.e. a devoted Vidal gauge tensor network type), but that's a separate issue.

mtfishman: I'm going back and forth about this issue, mostly because of my initial issue with this code, which was that it's not always clear which functions are potentially updating the cache or improving the gauging, vs. just converting naively from one gauge to another without performing any BP updates. In other words, I find that it is unclear and hard to keep track of when a function assumes it is in the idealized version of a gauge (i.e. assumes it is in a proper Vidal gauge or symmetric gauge, with a small gauge error), vs. when it improves the messages or gauge while doing the conversion. I think we should clarify that point and come up with a design which makes that choice clear, either through function names or keyword arguments, which of those choices is being made when you call one of these gauging functions.

src/gauging.jl
```diff
@@ -73,114 +78,58 @@ function vidal_gauge(
...
 function vidal_gauge(
   ψ::ITensorNetwork,
-  pψψ::PartitionedGraph,
-  mts;
-  eigen_message_tensor_cutoff=10 * eps(real(scalartype(ψ))),
-  regularization=10 * eps(real(scalartype(ψ))),
-  edges=NamedGraphs.edges(ψ),
-  svd_kwargs...,
+  bond_tensors;
+  (cache!)=nothing,
```
mtfishman, line 84: Related to the discussion below, this function is confusing to me.

If we think about the Vidal gauge as the combination of site tensors `ψ` and the `bond_tensors`, why would there also be a message tensor cache being input into this function? Don't the `bond_tensors` play the role of the messages, in the Vidal gauge?

src/caches/beliefpropagationcache.jl
JoeyT1994: Good point, I will get ride of that version.

src/caches/beliefpropagationcache.jl
JoeyT1994: That makes sense. I have now defined that.

src/gauging.jl
JoeyT1994: But in this case the function is just returning the error associated with `psi` and its bp_cache. So having the ability to update the cache or have it build a new cache doesn't really make sense to me here.

It probably shouldn't be called `gauge_error` at all and is really more of a bp function for how good a `bp_cache` is. I have now removed the function.

JoeyT1994:
@mtfishman  I have updated the `gauging.jl` code now by introducing a `VidalITensorNetwork` type which contains the `ITensorNetwork` and the `bond_tensors`.

The gauging code is then built around `update(psiv::VidalITensorNetwork; update_kwargs...)` which builds a bp_cache by default, updates it and then passes it to `update(psiv::VidalITensorNetwork, bp_cache::BeliefPropagationCache; update_kwargs...)` 

Let me know what you think!

src/gauging.jl
```diff
@@ -59,172 +86,96 @@ function vidal_gauge(
...
-  mts;
-  eigen_message_tensor_cutoff=10 * eps(real(scalartype(ψ))),
-  regularization=10 * eps(real(scalartype(ψ))),
-  edges=NamedGraphs.edges(ψ),
-  svd_kwargs...,
+function update(
+  ψv::VidalITensorNetwork;
+  (cache!)=nothing,
+  cache_update_kwargs=default_cache_update_kwargs(cache!),
+  kwargs...,
```
mtfishman, line 107: I don't see why this version and the version above that accepts the cache as an argument both need to exist, I'm not really a fan of the function above anyway. Can these be combined?

I'm also losing track of the logic behind what the cache here is being used for, since I thought the bond tensors of the Vidal gauge play the role of the BP cache.

src/gauging.jl
```diff
@@ -59,172 +86,96 @@ function vidal_gauge(
...
+  return update(ψv, cache![]; kwargs...)
 end
 
 """Transform from an ITensor in the Vidal Gauge (bond tensors) to the Symmetric Gauge (partitionedgraph, message tensors)"""
-function vidal_to_symmetric_gauge(ψ::ITensorNetwork, bond_tensors::DataGraph)
-  ψsymm = copy(ψ)
-  ψψsymm = norm_network(ψsymm)
-  pψψsymm = PartitionedGraph(ψψsymm, group(v -> v[1], vertices(ψψsymm)))
-  ψsymm_mts = default_bp_cache(pψψsymm)
+function symmetric_gauge(ψv::VidalITensorNetwork)
```
mtfishman, line 119: It seems like there is a common theme throughout all of these functions that we should settle and that I keep getting confused by, which is: when is a function improving the current gauge/message tensor cache, and then performing a conversion to a different gauge/format, vs. just performing a conversion to a different gauge/format?

My opinion on this is that any function could update the gauge/message tensors or not, and that is chosen with keyword arguments. I think until we settle a general design principle about this, this PR will be very difficult to review and complete.

src/apply.jl
```diff
@@ -298,33 +298,29 @@ end
 """Apply() function for an ITN in the Vidal Gauge. Hence the bond tensors are required.
 Gate does not necessarily need to be passed. Can supply an edge to do an identity update instead. Uses Simple Update procedure assuming gate is two-site"""
 function vidal_apply(
```
mtfishman, line 300: Seems like this could now just be called `apply` since it dispatches on the `VidalITensorNetwork` type.

src/gauging.jl
```diff
@@ -1,39 +1,66 @@
...
+data_graph(ψv::VidalITensorNetwork) = data_graph(itensornetwork(ψv))
+function copy(ψv::VidalITensorNetwork)
+  return VidalITensorNetwork(copy(itensornetwork(ψv)), copy(bond_tensors(ψv)))
+end
+
+function VidalITensorNetwork(
+  ψ::ITensorNetwork;
+  bond_tensors_cache=default_bond_tensor_cache,
+  bond_tensor=default_bond_tensor,
+)
```
mtfishman, line 25: I find this function a bit confusing, which relates to my comment below about the convention behind whether a certain function updates/improves the gauge/cache vs. just performs a conversion between formats.

If I were to guess what this function was doing without looking at the code, I would guess that `VidalITensorNetwork(ψ::ITensorNetwork)` might run BP on `ψ`, and then convert to the Vidal gauge using the BP message tensors. Instead, this just inserts identity tensors as the bond tensors. Instead of choosing between those two, I think it would make sense for this function to optionally accept a BP cache, optionally update that BP cache, and then convert to the Vidal gauge.

So, if no BP cache is passed, and the user specifies they don't want BP to be run, then it could fall back to this current behavior (using identity tensors as bond tensors).

src/gauging.jl
```diff
@@ -1,39 +1,66 @@
...
+  return DataGraph{vertextype(ψ),Nothing,ITensor}(underlying_graph(ψ))
+end
+default_bond_tensor(index) = denseblocks(delta(index, prime(index)))
+
+struct VidalITensorNetwork{V,BTS} <: AbstractITensorNetwork{V}
+  itensornetwork::ITensorNetwork{V}
+  bond_tensors::BTS
+end
 
+itensornetwork(ψv::VidalITensorNetwork) = ψv.itensornetwork
```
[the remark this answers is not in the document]
JoeyT1994: Yeah `site_tensors` seems reasonable.

src/apply.jl
JoeyT1994: Agreed.

src/gauging.jl
mtfishman: I think the devil is in the details there, and I find the current code logic to be confusing and hard to follow.

JoeyT1994:
@mtfishman I am thinking about how best to improve the logic of this code based on your comments.

I think a helper constructor `ITensorNetwork(psiv::VidalITensorNetwork)` which absorbs all bond tensors onto the sites symmetrically (by taking square roots) and returns to the arbitrary gauge is going to be helpful.

Then you could have the constructor `VidalITensorNetwork(psi::ITensorNetwork, bp_cache::BeliefPropagationCache)` which constructs a `VidalITensorNetwork` via the algorithm described in our paper.

Then you could have `update(psiv:VidalTensorNetwork; (cache!) = nothing, cache_update_kwargs)` which first converts to the symmetric gauge as you suggested in your comments. Then does whatever level of BP that you desire, and then calls the constructor above. That way you never simultaneously have `bond_tensors` and a `bp_cache`

What do you think? I know that `VidalITensorNetwork(psi::ITensorNetwork, bp_cache::BeliefPropagationCache)` doesn't do any cache updating and just takes it at face value but I personally think that is okay and a logical function to have?

src/gauging.jl
```diff
@@ -59,172 +86,96 @@ function vidal_gauge(
...
-    niters,
-    target_canonicalness,
-    svd_kwargs...,
-  )
-
-  return vidal_to_symmetric_gauge(ψ_vidal, bond_tensors)
-end
+function symmetric_gauge(ψ::ITensorNetwork; kwargs...)
+  ψv = VidalITensorNetwork(ψ)
+  ψv = update(ψv; kwargs...)
```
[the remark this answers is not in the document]
JoeyT1994: I have now written it so that this works. Although I still have the constructor `VidalITensorNetwork(ψ, bp_cache::BeliefPropagationCache)` in the code which this function passes to once it has done the relevant cache updating

src/gauging.jl
```diff
@@ -70,161 +110,56 @@ function vidal_gauge(
...
-    setindex_preserve_graph!(ψ_vidal, noprime(invroot_S * ψ_vidal[vsrc]), vsrc)
-    setindex_preserve_graph!(ψ_vidal, noprime(invroot_S * ψ_vidal[vdst]), vdst)
-  end
-
-  return ψ_vidal, bond_tensors
+function update(ψv::VidalITensorNetwork; kwargs...)
+  return VidalITensorNetwork(ITensorNetwork(ψv); kwargs...)
 end
 
 """Function to measure the 'isometries' of a state in the Vidal Gauge"""
```
mtfishman, line 133:
```suggestion
"""Function to construct the 'isometries' of a state in the Vidal Gauge"""
```

src/gauging.jl
```diff
@@ -70,161 +110,56 @@ function vidal_gauge(
...
-function vidal_itn_isometries(
-  ψ::ITensorNetwork,
-  bond_tensors::DataGraph;
-  edges=vcat(NamedGraphs.edges(ψ), reverse.(NamedGraphs.edges(ψ))),
+function vidal_gauge_isometries(
+  ψv::VidalITensorNetwork;
+  edges=vcat(NamedGraphs.edges(ψv), reverse.(NamedGraphs.edges(ψv))),
```
mtfishman, line 136:
```suggestion
  ψv::VidalITensorNetwork,
  edges=vcat(NamedGraphs.edges(ψv), reverse.(NamedGraphs.edges(ψv))),
```
I think it makes more sense for the edges to be passed as a positional argument.

Also, could we refactor this in terms of a core function `vidal_gauge_isometry(ψ::VidalITensorNetwork, edge)`, and then `vidal_gauge_isometries` loops over that function?

Another point is that I find it confusing when you call the variable `ψv`, at first I thought `v` was referring to a vertex, but I see now it is probably a shorthand for `vidal`. I think we should either just call it `ψ` when it is unambiguous what it refers to, or `ψ_vidal` if it seems necessary to be more explicit.

JoeyT1994: Seems reasonable! I have refactored this way and used `ψ::VidalITensorNetwork` aside from in a few functions where I need to differentiate and use `ψ_vidal`

src/apply.jl
```diff
@@ -297,34 +297,30 @@ end
 #In the future we will try to unify this into apply() above but currently leave it mostly as a separate function
 """Apply() function for an ITN in the Vidal Gauge. Hence the bond tensors are required.
 Gate does not necessarily need to be passed. Can supply an edge to do an identity update instead. Uses Simple Update procedure assuming gate is two-site"""
-function vidal_apply(
-  o::Union{ITensor,NamedEdge},
-  ψ::AbstractITensorNetwork,
-  bond_tensors::DataGraph;
-  normalize=false,
-  apply_kwargs...,
+function apply(
```
mtfishman, line 300:
```suggestion
function ITensors.apply(
```

src/apply.jl
```diff
@@ -297,34 +297,30 @@ end
...
+  o::Union{ITensor,NamedEdge}, ψv::VidalITensorNetwork; normalize=false, apply_kwargs...
```
mtfishman, line 301: Let's change `ψv` to `ψ_vidal`.

mtfishman: Or just `ψ`, there doesn't seem to be any ambiguity in this function.

src/gauging.jl
```diff
@@ -1,39 +1,88 @@
-"""initialize bond tensors of an ITN to identity matrices"""
-function initialize_bond_tensors(ψ::ITensorNetwork; index_map=prime)
-  bond_tensors = DataGraph{vertextype(ψ),Nothing,ITensor}(underlying_graph(ψ))
+function default_bond_tensor_cache(ψ::ITensorNetwork)
```
mtfishman, line 1: Seems like this could just be called `default_bond_tensors`.

src/gauging.jl
```diff
@@ -1,39 +1,88 @@
...
-function vidal_gauge(
-  ψ::ITensorNetwork,
-  pψψ::PartitionedGraph,
-  mts,
-  bond_tensors::DataGraph;
-  eigen_message_tensor_cutoff=10 * eps(real(scalartype(ψ))),
+"""Use an ITensorNetwork ψ, its bond tensors and belief propagation cache to put ψ into the vidal gauge, return the bond tensors and updated_ψ."""
+function vidalitensornetwork_preserve_cache(
+  ψ::ITensorNetwork;
+  bp_cache::BeliefPropagationCache=default_norm_cache(ψ),
```
mtfishman, line 65:
```suggestion
  cache=default_norm_cache(ψ),
```

src/gauging.jl
```diff
@@ -1,39 +1,88 @@
...
+  bond_tensors_cache=default_bond_tensor_cache,
```
mtfishman, line 66:
```suggestion
  bond_tensors=default_bond_tensors,
```

src/gauging.jl
```diff
@@ -1,39 +1,88 @@
...
+  message_cutoff=10 * eps(real(scalartype(ψ))),
   regularization=10 * eps(real(scalartype(ψ))),
   edges=NamedGraphs.edges(ψ),
   svd_kwargs...,
 )
-  ψ_vidal = copy(ψ)
+  ψ_vidal_site_tensors = copy(ψ)
+  bond_tensors = bond_tensors_cache(ψ)
```
mtfishman, line 73:
```suggestion
  ψ_vidal_bond_tensors = bond_tensors_cache(ψ)
```

src/apply.jl
```diff
@@ -339,34 +335,34 @@ function vidal_apply(
...
-        ψv2 = noprime(ψv2 * inv_diag(bond_tensors[vn => dst(e)]))
+        ψv2 = noprime(ψv2 * inv_diag(bond_tensor(ψv, vn => dst(e))))
       end
     end
 
     if normalize
       ψv1 /= norm(ψv1)
       ψv2 /= norm(ψv2)
-      normalize!(bond_tensors[e])
+      normalize!(updated_bond_tensors[e])
```
mtfishman, line 355:
```suggestion
      updated_bond_tensors[e] /= norm(updated_bond_tensors[e])
```
Best to avoid in-place updates of tensor data unless there is a compelling reason.

src/gauging.jl
```diff
@@ -70,161 +118,60 @@ function vidal_gauge(
...
-
-"""Function to measure the 'isometries' of a state in the Vidal Gauge"""
-function vidal_itn_isometries(
-  ψ::ITensorNetwork,
-  bond_tensors::DataGraph;
-  edges=vcat(NamedGraphs.edges(ψ), reverse.(NamedGraphs.edges(ψ))),
-)
-  isometries = Dict()
+  ψ_vsrcdag = dag(ψ_vsrc)
+  replaceind!(ψ_vsrcdag, commonind(ψ_vsrc, ψ[vdst]), commonind(ψ_vsrc, ψ[vdst])')
```
mtfishman, line 152:
```suggestion
  ψ_vsrcdag = replaceind(ψ_vsrcdag, commonind(ψ_vsrc, ψ[vdst]) => commonind(ψ_vsrc, ψ[vdst])')
```
Best to avoid in-place operations on ITensors unless there is a compelling reason.

JoeyT1994: Okay sure, I have changed that and try to avoid in-place operations from now on in development.

src/apply.jl
```diff
@@ -170,7 +170,7 @@ end
 
 function ITensors.apply(
   o::ITensor,
-  ψ::ITensorNetwork;
+  ψ::Union{ITensorNetwork,TreeTensorNetwork};
```
mtfishman, line 173: Can we use `AbstractITensorNetwork` here? My understanding is that if the network is `VidalITensorNetwork` it will call the other version, otherwise it will call this version (i.e. the Julia compiler will dispatch to the most specialized version it can find).

JoeyT1994: I was thinking this too but when I did this I got the following bug from `test/test_apply.jl` which is trying to use `apply` on an `VidalITensorNetwork`
image

Suggesting that Julia doesn't see it as a more specialized version, which I am surprised by?

mtfishman: I think that issue is caused by the operator being constrained to `o::ITensor` in one and `o::Union{ITensor,NamedEdge}` in the other. That may be confusing the compiler, since it can't decide if it should choose the version of the method based on the type of `o` or `ψ`.

What if we just remove the type constraint on `o` in both?

JoeyT1994: Yeah good spot, that's why it's happening.

Okay I have removed the constraint.

Issue #140, opened 2024-02-27
`alternating_update` breaks for certain vertextypes


b-kloss:
Algorithms based on alternating update should work for any vertextype, given that the input state and operator are TTNs. Setting up a `path_graph` with non-standard `vertextype` as in the following test

```
using ITensors
using ITensorNetworks
using ITensorNetworks: exponentiate_updater,add_edge!
using Test

@testset "MPS TDVP with other vertexnames" begin
    ## make path_graph with vertextype(g)==Tuple(Int,Int,Vararg{Int})
    g = ITensorNetworks.NamedGraph([(1,1),(1,2,1),(1,2,3,4),(1,4)])
    add_edge!(g,(1,1)=>(1,2,1))
    add_edge!(g,(1,2,1)=>(1,2,3,4))
    add_edge!(g,(1,2,3,4)=>(1,4))
    @show ITensorNetworks.vertextype(g)
    
    s = siteinds("S=1/2", g)
    os=ITensorNetworks.heisenberg(s)
    H = TTN(os, s)
    ψ0 = random_ttn(s)
    @test is_tree(H)
    @test is_tree(ψ0)
    
    # Time evolve forward:
    @test (ψ1 = tdvp(H, -0.1im, ψ0; nsteps=1, cutoff=1e-10, nsites=1))
end
```
leads to output
```
julia> include("test_treetensornetworks/test_solvers/test_tdvp_anyvertextype.jl")
ITensorNetworks.vertextype(g) = Tuple{Int64, Int64, Vararg{Int64}}
MPS TDVP with other vertexnames: Error During Test at /mnt/home/bkloss/.julia/v1.9/dev/ITensorNetworks/test/test_treetensornetworks/test_solvers/test_tdvp_anyvertextype.jl:25
  Test threw exception
  Expression: ψ1 = tdvp(H, -0.1im, ψ0; nsteps = 1, cutoff = 1.0e-10, nsites = 1)
  AssertionError: is_tree(itensor_network)
  Stacktrace:
    [1] TreeTensorNetwork (repeats 2 times)
      @ ~/.julia/v1.9/dev/ITensorNetworks/src/treetensornetworks/ttn.jl:16 [inlined]
    [2] #orthogonalize#537
      @ ~/.julia/v1.9/dev/ITensorNetworks/src/treetensornetworks/abstracttreetensornetwork.jl:108 [inlined]
[...]

PR #141, opened 2024-02-28
Improve compatibility with arbitrary vertextypes and tree structures 

b-kloss:
This PR addresses some bugs in `alternating_update` when using certain `vertextype`s.

- Removed `maybe_only` from `utils.jl` and `models.jl` since it is not needed anymore
- Added default `link_space=1` to `random_ttn`, in line with `ITensors.randomMPS`, to avoid disconnecting product tree tebsir networks during first sweep.#140
- Make `alternating_update` compatible with e.g. binary trees. This used to fail due to [loss of type-information](https://github.com/andyferris/Dictionaries.jl/issues/97) of `Dictionaries.getindices` when applied to certain empty `Dictionary`.
- Added test for the latter functionality.

It does not address issues relating to ill-defined `isless` comparison on arbitrary `vertextype` in `ttn_svd`, which should be resolved in a separate PR.

### DIFF
--- a/src/treetensornetworks/projttns/abstractprojttn.jl
+++ b/src/treetensornetworks/projttns/abstractprojttn.jl
@@ -129,13 +129,19 @@ end
 function invalidate_environments(P::AbstractProjTTN)
   ie = internal_edges(P)
   newenvskeys = filter(!in(ie), keys(environments(P)))
-  P = set_environments(P, getindices(environments(P), newenvskeys))
+  T=typeof(environments(P))
+  newenvs=getindices(environments(P),newenvskeys )
+  !(typeof(newenvs)==T) && (newenvs=T(newenvs))
+  P = set_environments(P, newenvs)
   return P
 end
 
 function invalidate_environment(P::AbstractProjTTN, e::AbstractEdge)
+  T=typeof(environments(P))
   newenvskeys = filter(!isequal(e), keys(environments(P)))
-  P = set_environments(P, getindices(environments(P), newenvskeys))
+  newenvs = getindices(environments(P),newenvskeys )
+  !(typeof(newenvs)==T) && (newenvs=T(newenvs))
+  P = set_environments(P, newenvs)
   return P
 end
 
--- a/src/treetensornetworks/projttns/abstractprojttn.jl
+++ b/src/treetensornetworks/projttns/abstractprojttn.jl
@@ -129,19 +129,14 @@ end
 function invalidate_environments(P::AbstractProjTTN)
   ie = internal_edges(P)
   newenvskeys = filter(!in(ie), keys(environments(P)))
-  T = typeof(environments(P))
-  newenvs = getindices(environments(P), newenvskeys)
-  !(typeof(newenvs) == T) && (newenvs = T(newenvs))
-  P = set_environments(P, newenvs)
+  P = set_environments(P, getindices_narrow_keytype(environments(P), newenvskeys))
   return P
 end
 
 function invalidate_environment(P::AbstractProjTTN, e::AbstractEdge)
   T = typeof(environments(P))
   newenvskeys = filter(!isequal(e), keys(environments(P)))
-  newenvs = getindices(environments(P), newenvskeys)
-  !(typeof(newenvs) == T) && (newenvs = T(newenvs))
-  P = set_environments(P, newenvs)
+  P = set_environments(P, getindices_narrow_keytype(environments(P), newenvskeys))
   return P
 end
 
--- a/src/treetensornetworks/projttns/abstractprojttn.jl
+++ b/src/treetensornetworks/projttns/abstractprojttn.jl
@@ -129,18 +129,18 @@ end
 function invalidate_environments(P::AbstractProjTTN)
   ie = internal_edges(P)
   newenvskeys = filter(!in(ie), keys(environments(P)))
-  T=typeof(environments(P))
-  newenvs=getindices(environments(P),newenvskeys )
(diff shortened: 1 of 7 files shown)


src/treetensornetworks/projttns/abstractprojttn.jl
```diff
@@ -129,13 +129,19 @@ end
 function invalidate_environments(P::AbstractProjTTN)
   ie = internal_edges(P)
   newenvskeys = filter(!in(ie), keys(environments(P)))
-  P = set_environments(P, getindices(environments(P), newenvskeys))
+  T = typeof(environments(P))
```
b-kloss, line 132: Alternatively, we may want to define a function `safe_getindices` and use it instead of `get_indices`, and reference the issue there?

src/treetensornetworks/ttn.jl
```diff
@@ -66,8 +66,8 @@ end
 # TODO: Implement `random_circuit_ttn` for non-trivial
 # bond dimensions and correlations.
 # TODO: Implement random_ttn for QN-Index
-function random_ttn(args...; kwargs...)
-  T = TTN(args...; kwargs...)
+function random_ttn(args...; link_space=1, kwargs...)
+  T = TTN(args...; link_space, kwargs...)
```
mtfishman, line 70: Maybe instead `TTN` should use a default of `link_space=1`?

src/treetensornetworks/projttns/abstractprojttn.jl
mtfishman: That sounds like a good strategy. We could call it `getindices_narrow_keytype` to be more explicit about what it is doing and how it is different from `getindices`.

src/treetensornetworks/ttn.jl
b-kloss: I addressed this, but I am not entirely sure whether that could have unwanted side-effects?

src/treetensornetworks/projttns/abstractprojttn.jl
b-kloss: I define it now in `src/utils.jl`.

src/treetensornetworks/ttn.jl
```diff
@@ -45,8 +45,8 @@ function TTN(g::AbstractGraph, args...; kwargs...)
   return TTN(Float64, g, args...; kwargs...)
 end
 
-function TTN(eltype::Type{<:Number}, graph::AbstractGraph, args...; kwargs...)
-  itensor_network = ITensorNetwork(eltype, graph; kwargs...)
+function TTN(eltype::Type{<:Number}, graph::AbstractGraph, args...; link_space=1, kwargs...)
+  itensor_network = ITensorNetwork(eltype, graph; link_space, kwargs...)
```
mtfishman, line 49: Sorry to be a bother, but probably we should define `link_space=1` as the default at the `ITensorNetwork` level, if that doesn't cause issues. It would be pretty arbitrary/inconsistent if `TTN` had that behavior but `ITensorNetwork`didn't.

src/utils.jl
```diff
@@ -25,3 +22,10 @@ function line_to_tree(line::Vector)
   end
   return [line_to_tree(line[1:(end - 1)]), line[end]]
 end
+
+function getindices_narrow_keytype(d::Dictionary, indices)
+  T = typeof(d)
+  d_at_indices = getindices(d, indices)
+  T != typeof(d_at_indices) && (d_at_indices = T(d_at_indices))
```
mtfishman, line 29: I think you can just use `d_at_indices = convert(typeof(d), d_at_indices)`, which won't do anything if `typeof(d) == typeof(d_at_indices)`.

So we could just write this as:
```julia
function getindices_narrow_keytype(d::Dictionary, indices)
  return convert(typeof(d), getindices(d, indices))
end
```

mtfishman:
Could you add tests for the `link_space` behavior of `TTN` and `ITensorNetwork`, i.e. test that they make link indices of dimension 1 by default?

src/treetensornetworks/ttn.jl
b-kloss: Sure, no problem. I was thinking the same, but again wasn't sure whether this may break other functionality.

src/utils.jl
b-kloss: Thanks, that's good to know.

src/treetensornetworks/ttn.jl
b-kloss: Moving it to `ITensorNetworks` is breaking at least the `mincut` tests in `test_itensornetworks.jl`.

mtfishman: Ok, I would be biased towards investigating failing tests, and either modifying them or disabling them, since setting the default of the `ITensorNetwork` constructor to `link_space=1` is probably the right way to go.

b-kloss: It looks like it's not passing the kwargs properly for some signatures, I'll take care of it.

mtfishman:
Also the `union` behavior of `IndsNetwork` (or more generally `DataGraph`) seems to be inconsistent with the behavior of `union` for other types, i.e.:
```julia
julia> union(Any[1.0, 3.0], Any[1, 2, 3])
3-element Vector{Any}:
 1.0
 3.0
 2
```
It should end up with values from the first `IndsNetwork` not the second one.

mtfishman:
Thanks!

emstoudenmire:
This is a big improvement – excited to see this

PR #143, opened 2024-02-29
Refactor sweeps interface

b-kloss:
This work-in-progress PR implements a more flexible `sweeps` interface for `alternating_update` based algorithms.
Currently, a proof-of-principle is implemented for `tdvp`, while other algorithms are broken. The purpose of this draft is to discuss the overall structure of the code, not the details of the implementation of specific functions.

The interface is based around passing a list of length `nsweeps`, each of which contains three `NamedTuples`: `extracter_kwargs`, `updater_kwargs`, `inserter_kwargs`. The latter hold the epynomous functions as well as their specified keyword args. These named_tuples are passed as kwargs to `tdvp`. Any other `kwargs` passed to `tdvp` will be moved into `inserter_kwargs` for convenience and continuity with prior interfaces.

There are two main drawbacks of this design:
- If we wish to pass a `kwarg` to both `inserter` and `updater`, we need to list it twice in the call to `tdvp`.
- Coupling of different components: `tdvp_sweeps_plan`, which is called inside the call to `tdvp` currently adds (;`time_step`, `substep`) to `updater_kwargs`, which the `updater` needs to be aware of. Changes w.r.t. this behaviour in `tdvp_sweeps_plan` require changes in the `updater`.

The latter two could be circumvented with a design that groups the function and a collection of symbols (i.e. keys of `kwargs`) it can digest. For example for the default TDVP behaviour, we would use something like `(exponentiate,[:time_step])`. `tdvp_sweeps_plan` could then be made aware of this signature, in order to not add `substep` to the `updater_kwargs`. Matt and I discussed this extensively already, discarding this idea as to complicated.

Other changes:
- Removed `sweeps_param` from `alternating_update`.
- Move `sweep_plan` related functionality to `tree_sweeping.jl`.
- Compute observables with `sweep_observer` instead of `region_observer` in tests.


src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -1,62 +1,17 @@
...
-  updater,
   projected_operator,
   init_state::AbstractTTN;
-  checkdone=(; kws...) -> false,
-  outputlevel::Integer=0,
-  nsweeps::Integer=1,
+  sweep_plans,  #this is really the only one beig pass all the way down
+  outputlevel, # we probably want to extract this one indeed for passing to observer etc.
+  checkdone=(; kws...) -> false,  ### move outside
   (sweep_observer!)=observer(),
```
mtfishman, line 8: Change to `(sweep_observer!)=nothing`.

src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -1,62 +1,17 @@
...
 )
-  maxdim, mindim, cutoff, noise, kwargs = process_sweeps(nsweeps; kwargs...)
-
   state = copy(init_state)
-
-  insert_function!(sweep_observer!, "sweep_printer" => sweep_printer) # FIX THIS
-
-  for which_sweep in 1:nsweeps
-    if !isnothing(write_when_maxdim_exceeds) &&
+  for (which_sweep, sweep_plan) in enumerate(sweep_plans)
```
mtfishman, line 13:
```suggestion
  for which_sweep in eachindex(sweep_plans)
    sweep_plan = sweep_plans[which_sweep]
```
I'm not a fan of `enumerate`, it doesn't necessarily give a valid index into the collection:
```julia
julia> using OffsetArrays

julia> collect(enumerate(OffsetVector(randn(10), -1)))
10-element Vector{Tuple{Int64, Float64}}:
 (1, 1.2097571700024081)
 (2, 1.9625931511876364)
 (3, -2.048843359679102)
 (4, -0.7910034346777152)
 (5, -1.7837242897474845)
 (6, 0.03179938683614588)
 (7, 0.4589608248671337)
 (8, 0.5987573798836898)
 (9, 0.3850021806565198)
 (10, 0.8128603681636092)

julia> collect(eachindex(OffsetVector(randn(10), -1)))
10-element Vector{Int64}:
 0
 1
 2
 3
 4
 5
 6
 7
 8
 9
```

src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -65,41 +20,27 @@ function alternating_update(
       end
       projected_operator = disk(projected_operator)
     end
-    sweep_params = (;
-      maxdim=maxdim[which_sweep],
-      mindim=mindim[which_sweep],
-      cutoff=cutoff[which_sweep],
-      noise=noise[which_sweep],
-    )
     sw_time = @elapsed begin
```
mtfishman, line 23:
```suggestion
    sweep_time = @elapsed begin
```

src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -65,41 +20,27 @@ function alternating_update(
...
-    checkdone(; state, which_sweep, outputlevel, kwargs...) && break
+    sweep_printer(; state, which_sweep, sw_time, outputlevel)
+    checkdone(; state, which_sweep, outputlevel, sweep_plan) && break
   end
-  select!(sweep_observer!, Observers.DataFrames.Not("sweep_printer"))
   return state
 end
 
-function alternating_update(updater, H::AbstractTTN, init_state::AbstractTTN; kwargs...)
+function alternating_update(H::AbstractTTN, init_state::AbstractTTN; kwargs...)
```
mtfishman, line 36:
```suggestion
function alternating_update(operator::AbstractTTN, init_state::AbstractTTN; kwargs...)
```

src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -121,14 +62,12 @@ each step of the algorithm when optimizing the MPS.
 Returns:
 * `state::MPS` - time-evolved MPS
 """
-function alternating_update(
-  updater, Hs::Vector{<:AbstractTTN}, init_state::AbstractTTN; kwargs...
-)
+function alternating_update(Hs::Vector{<:AbstractTTN}, init_state::AbstractTTN; kwargs...)
```
mtfishman, line 65:
```suggestion
function alternating_update(operators::Vector{<:AbstractTTN}, init_state::AbstractTTN; kwargs...)
```

src/treetensornetworks/solvers/tdvp.jl
```diff
@@ -30,56 +30,78 @@ function sub_time_steps(order)
...
-  updater,
   operator,
   t::Number,
   init_state::AbstractTTN;
   time_step::Number=t,
   nsites=2,
   nsteps=nothing,
   order::Integer=2,
+  outputlevel=0,
   (sweep_observer!)=observer(),
```
mtfishman, line 60:
```suggestion
  (sweep_observer!)=nothing,
```

mtfishman, line 60: Let's define `default_*()` functions for these defaults. Also, isn't that `time_step` value incorrect if there is more than one sweep?

src/treetensornetworks/solvers/tdvp.jl
```diff
@@ -30,56 +30,78 @@ function sub_time_steps(order)
...
+      order,
+      nsites,
+      time_step,
+      init_state;
+      root_vertex,
+      reverse_step,
+      pre_region_args=processed_kwarg_list[i],
+    )
+    push!(sweep_plans, sweep_plan)
+  end
```
mtfishman, line 99: Reminder to absorb this loop into a `tdvp_sweep_plans(...)` function.

src/treetensornetworks/solvers/tdvp.jl
```diff
@@ -30,56 +30,78 @@ function sub_time_steps(order)
...
 
   function sweep_time_printer(; outputlevel, which_sweep, kwargs...)
     if outputlevel >= 1
       sweeps_per_step = order ÷ 2
-      if sweep % sweeps_per_step == 0
+      if which_sweep % sweeps_per_step == 0
         current_time = (which_sweep / sweeps_per_step) * time_step
         println("Current time (sweep $which_sweep) = ", round(current_time; digits=3))
       end
```
mtfishman, line 107: This should be defined outside of the `tdvp` function.

src/treetensornetworks/solvers/tdvp.jl
```diff
@@ -90,42 +112,11 @@ function tdvp(
   insert_function!(sweep_observer!, "sweep_time_printer" => sweep_time_printer)
```
mtfishman, line 112: Should this get absorbed into the `sweep_printer`?

src/treetensornetworks/solvers/tree_sweeping.jl
```diff
@@ -63,3 +63,117 @@ function half_sweep(dir::Base.ReverseOrdering, args...; kwargs...)
     reverse(half_sweep(Base.Forward, args...; kwargs...)),
   )
 end
+
+function default_region_args_func(internal_kwargs::NamedTuple, pre_region_args::NamedTuple)
+  #wrap args in another NamedTuple
+  internal_kwargs = merge(get(pre_region_args, :internal_kwargs, (;)), internal_kwargs)
+  return merge(pre_region_args, (; internal_kwargs))
+end
```
mtfishman, line 71: Remove this, just use named tuple splatting.

src/treetensornetworks/solvers/tree_sweeping.jl
```diff
@@ -63,3 +63,117 @@ function half_sweep(dir::Base.ReverseOrdering, args...; kwargs...)
...
+  nsites,
+  graph::AbstractGraph;
+  region_args_func=default_region_args_func,
+  reverse_args_func=default_region_args_func,
+  pre_region_args=(;),
+  kwargs...,
+)  ###move this to a different file, algorithmic level idea
+  return vcat(
+    [
+      half_sweep(
```
mtfishman, line 83: Rename `forward_sweep(...)` or something similar.

Remove the direction, and define a reverse sweep as `reverse(forward_sweep(...))`.

src/treetensornetworks/solvers/update_step.jl
```diff
@@ -1,20 +1,8 @@
...
-        nsites,
-        region_args=(; half_sweep=half),
-        kwargs...,
-      ) for half in 1:2
-    ]...,
-  )
-end
-
-function region_update_printer(;
+function default_region_update_printer(;
```
mtfishman, line 5: Maybe `default_region_printer`?

src/treetensornetworks/solvers/update_step.jl
```diff
@@ -106,15 +73,34 @@ end
 # apart and puts it back into the network.
 #
 
-function extract_local_tensor(state::AbstractTTN, pos::Vector)
-  return state, prod(state[v] for v in pos)
+function extract_prolog(state::AbstractTTN, region)
+  return state = orthogonalize(state, current_ortho(region))
```
mtfishman, line 77:
```suggestion
  return orthogonalize(state, current_ortho(region))
```

src/treetensornetworks/solvers/update_step.jl
```diff
@@ -200,38 +184,47 @@ function region_update(
...
-    normalize,
-    maxdim=region_kwargs.maxdim,
-    mindim=region_kwargs.mindim,
-    cutoff=region_kwargs.cutoff,
+    region,
+    which_sweep,
+    spec,
+    outputlevel,
+    info...,
+    region_kwargs...,
```
mtfishman, line 224: Collect into a named tuple and pass to both `update!` and `region_printer`.

src/treetensornetworks/solvers/update_step.jl
```diff
@@ -200,38 +184,47 @@ function region_update(
...
   )
 
-  update!(
-    region_observer!;
+  printer(;
```
mtfishman, line 227: Call this `region_printer`, pass down from the top level functions `alternating_update`, `tdvp`, etc.

src/treetensornetworks/solvers/alternating_update.jl
```diff
@@ -1,62 +1,17 @@
...
-
-  insert_function!(sweep_observer!, "sweep_printer" => sweep_printer) # FIX THIS
-
-  for which_sweep in 1:nsweeps
-    if !isnothing(write_when_maxdim_exceeds) &&
+  for (which_sweep, sweep_plan) in enumerate(sweep_plans)
+    if !isnothing(write_when_maxdim_exceeds) && #fix passing this
       maxdim[which_sweep] > write_when_maxdim_exceeds
       if outputlevel >= 2
         println(
```
mtfishman, line 17: This entire logic around `write_when_maxdim_exceeds` could be handled through a callback function that gets passed down from the top level. For example, we could pass down a function like:
```julia
function cache_to_disk(operator; which_sweep, maxdim, outputlevel, write_when_maxdim_exceeds)
    maxdim[which_sweep] > write_when_maxdim_exceeds
    if outputlevel >= 2
      println(
        "write_when_maxdim_exceeds = $write_when_maxdim_exceeds and maxdim[which_sweep] = $(maxdim[which_sweep]), writing environment tensors to disk",
      )
    end
    operator = disk(operator)
  end
  return operator
end
```
as a keyword argument `transform_operator` passed down from `alternating_update`, `tdvp`, etc. `write_when_maxdim_exceeds` could get passed down from a `transform_operator_kwargs` named tuple, also passed down from `alternating_update`, `tdvp`, etc.

Then, this could would just be replaced by a call to `transform_operator(projected_operator; which_sweep, maxdim, outputlevel, transform_operator_kwargs...)`, which is simpler for this code and more general.

src/treetensornetworks/solvers/tdvp.jl
```diff
@@ -30,56 +30,78 @@ function sub_time_steps(order)
...
   kwargs...,
 )
+  # slurp unbound kwargs into inserter
+  ### if unbound kwargs are required in e.g. updater, they will have to be explicitly listed inside e.g. updater_kwargs
+  ### if non-standard inserter (with different kwarg signature) is used, it is up to the user to not pass unsupported kwargs
+  inserter_kwargs = (; inserter_kwargs..., kwargs...) # slurp unbound kwargs into inserter
+  # move inserter etc. into the respective kwargs
+  inserter_kwargs = merge((; inserter), inserter_kwargs)
+  updater_kwargs = merge((; updater), updater_kwargs)
+  extracter_kwargs = merge((; extracter), extracter_kwargs)
```
mtfishman, line 78: Don't merge the functions into the named tuples in this way, instead just pass to `process_kwargs_for_sweeps` (or whatever that will be called).

mtfishman:
@b-kloss here's a design idea for `process_kwargs_for_sweeps` (or whatever it will be called) for dealing with nested named tuples:
```julia
kwargs = (;
  update=expand_and_eigsolve_update,
  update_kwargs=(; eigsolve_kwargs=(; krylovdim=20), expand_kwargs=(; maxdim=30)),
  insert_kwargs=(; cutoff=1e-4),
)
```
can get processed recursively to:
```julia
(;
  update=[expand_and_eigsolve_update, expand_and_eigsolve_update, expand_and_eigsolve_update],
  update_kwargs=(; eigsolve_kwargs=(; krylovdim=[20, 20, 20]), expand_kwargs=(; maxdim=[30, 30, 30])),
  insert_kwargs=(; cutoff=[1e-4, 1e-4, 1e-4]),
)
```
for example for `nsweeps=3`. This then gets split to:
```julia
[kwargs, kwargs, kwargs]
```
which could again be done recursively. From what I can tell this doesn't lead to any ambiguity, since if you want a keyword argument to be a `NamedTuple`, you can wrap it in a `Vector`, but maybe I'm missing something there.

mtfishman:
By the way, it may be helpful to use a package like [StructWalk.jl](https://github.com/chengchingwen/StructWalk.jl) for recursively calling into the nested NamedTuples.

Also, I guess we want to think about a more generic interface for adding on subspace expansion to any local solver. An interface I could imagine is:
```julia
kwargs = (;
  updater=compose_updaters(; solver=eigsolve_updater, expander=local_subspace_expansion),
  updater_kwargs=(; solver_kwargs=(; krylovdim=20), expander_kwargs=(; maxdim=30)),
  inserter_kwargs=(; cutoff=1e-4),
)
```
Then, `compose_updaters(...)` can be a function that builds a nested function that includes the subspace expansion and the solver. It could be quite general, i.e. not be hardcoded to assume it accepts keyword arguments `solver` and `expander`, and for each updater `*` passed to `compose_updaters` it will look for the corresponding keyword arguments `*_kwargs`.

src/treetensornetworks/solvers/tree_sweeping.jl
b-kloss: This will also require fixing places in the code which rely on the direction of edges (currently the `inserter` function does for example, but it should be rewritten using `region` and `next_region` logic), but that should of course be done anyway.

src/treetensornetworks/solvers/tdvp.jl
b-kloss: Yes, in principle. Also the current implementation of this is not general enough. One should be able to give a a vector of timesteps (which gets processed together with the other ways of specifying the timestep). We could pass the `region_observer` to the `region_printer`, and similar the `sweep_observer` to the `sweep_printer` in order to access an auxiliary variable like the physical time, computed by some function in the observer. `tdvp` would then have its own default observers and printers (since physical time is not something that matters in most other algorithms).

src/treetensornetworks/solvers/tdvp.jl
b-kloss: Should this be done in a single file `defaults.jl` for the whole package?

mtfishman: Maybe for now we could make a file `src/solvers/defaults.jl`.

b-kloss:
> Refactor make_regions, define forward_regions(nsite, graph) such that forward_regions(2, path_graph(10)) = [(1, 2), (2, 3), (3, 4), ..., (9, 10)] and insert_region_intersections(forward_regions(2, path_graph(10))) = [(1, 2), (2,), (2, 3), (3,), (3, 4), ...(9,), (9, 10)].
There's a problem with this on trees, at least for `nsites=1`, and to a lesser extent also `nsites=2`. It seems impossible to recover for `nsites=1` what the reverse edge is supposed to be when two regions are not contiguous, as happen's on general trees. For `nsites=2`, this also happens but it can be recovered by looking at the past/future regions.
Currently, there's an implementation of `reverse_region` that's working and similar to the logic in `forward_region` --- both of them are specialized to `nsites=1,2`. I suggest we stick with that for now and leave a more general `insert_region_intersections` logic for the future?

mtfishman:
> There's a problem with this on trees, at least for `nsites=1`, and to a lesser extent also `nsites=2`. It seems impossible to recover for `nsites=1` what the reverse edge is supposed to be when two regions are not contiguous, as happen's on general trees. For `nsites=2`, this also happens but it can be recovered by looking at the past/future regions. Currently, there's an implementation of `reverse_region` that's working and similar to the logic in `forward_region` --- both of them are specialized to `nsites=1,2`. I suggest we stick with that for now and leave a more general `insert_region_intersections` logic for the future?

It seems like it is easy to extend to the the case of `nsites == 1` by including edges incident to the vertex when considering region intersections.

For `nsites >= 2`, I was picturing that at first you would compute the list of intersections of any region with any other region, independent of the update order. Then you could order the resulting region intersections based on the order of the original region updates.

Anyway, I see that it is a bit of an involved sub-project to get all of that logic correct and make sure it really works for many different types of graphs and update choices, so it sounds reasonable to use the existing code for now.

src/treetensornetworks/solvers/alternating_update.jl
b-kloss: Are you suggesting to put `transform_operator_kwargs` in the sweep_plan, like `updater_kwargs` etc? Or to treat it separately and pass it to `alternating_update` only? The former would move it from `alternating_update` to `update_region`, not sure if that's a problem in terms of repeating the call to `disk`?

mtfishman: That's a good question. Probably ultimately it does make sense for it to be processed as part of the sweep plan, but maybe we can leave this change for future work since there are some nontrivial choices to make.

src/solvers/local_solvers/eigsolve.jl
```diff
@@ -0,0 +1,36 @@
...
+    maxiter,
+    verbosity,
+    eager,
+  )
+  return vecs[1], (; info, eigvals=vals)
+end
+
+function _pop_which_eigenvalue(; which_eigenvalue, kwargs...)
+  return which_eigenvalue, NamedTuple(kwargs)
+end
```
mtfishman, line 36: Is this needed?

src/solvers/region_update/update_step.jl
mtfishman, line 1: I think a better name for this file would be `region_update.jl`. Also, maybe this could get moved to the `src/solvers/alternating_update` folder.

src/solvers/sweep_plans/sweep_plans.jl
```diff
@@ -0,0 +1,236 @@
...
+      intersecting_region = NamedGraphs.NamedEdge(only(regions[i]), only(regions[i + 1]))
+      if !has_edge(graph, intersecting_region)
+        error("Edge not in graph")
+      end
+    end
+    push!(intersecting_steps, (intersecting_region, region_kwargs))
+  end
+  return interleave(steps, intersecting_steps)
+end
+=#
```
mtfishman, line 77: Let's just delete this for now.

mtfishman:
Thanks again @b-kloss, this is a big improvement for making the code both simpler and more flexible and customizable!

emstoudenmire:
Huge PR! And a lot of work. The code looks so much cleaner now and I'm enthused about the direction this is going.

PR #144, opened 2024-03-07
Improve compatibility with tensor networks without internal vertices

b-kloss:
This PR addresses some issues related to using Tensor Networks where the `vertex_data` at some vertices is empty, e.g. `Index[]`. In particular the behaviour of `check_hascommoninds` which is for example called when computing an inner product between two tensor networks is now comparing empty `vertex_data` as equal. A test has been added that verifies that an inner product can be taken.

ToDo:
- [ ] Maybe move test for missing siteinds functionality to a different file.
- [ ] Fix issue with specific state constructor for the case of missing siteinds. For example, `TTN(s,"Up")` is currently not functional if `s` is an `IndsNetwork` where at least one vertex has empty `vertex_data`.

### DIFF
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -798,7 +798,11 @@ function hascommoninds(
   ::typeof(siteinds), A::AbstractITensorNetwork{V}, B::AbstractITensorNetwork{V}
 ) where {V}
   for v in vertices(A)
-    !hascommoninds(siteinds(A, v), siteinds(B, v)) && return false
+    if isempty(siteinds(A, v))
+      !isempty(siteinds(B, v)) && return false
+    else
+      !hascommoninds(siteinds(A, v), siteinds(B, v)) && return false
+    end
   end
   return true
 end
@@ -815,9 +819,13 @@ function check_hascommoninds(
     )
   end
   for v in vertices(A)
-    !hascommoninds(siteinds(A, v), siteinds(B, v)) && error(
-      "$(typeof(A)) A and $(typeof(B)) B must share site indices. On vertex $v, A has site indices $(siteinds(A, v)) while B has site indices $(siteinds(B, v)).",
-    )
+    if isempty(siteinds(A, v))
+      !(isempty(siteinds(B, v))) && return false
+    else
+      !hascommoninds(siteinds(A, v), siteinds(B, v)) && error(
+        "$(typeof(A)) A and $(typeof(B)) B must share site indices. On vertex $v, A has site indices $(siteinds(A, v)) while B has site indices $(siteinds(B, v)).",
+      )
+    end
   end
   return nothing
 end
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -820,7 +820,9 @@ function check_hascommoninds(
   end
   for v in vertices(A)
     if isempty(siteinds(A, v))
-      !(isempty(siteinds(B, v))) && return false
+      !(isempty(siteinds(B, v))) && error(
+        "$(typeof(A)) A and $(typeof(B)) B must share site indices. On vertex $v, A has no site indices while B has site indices $(siteinds(B, v)).",
+      )
     else
       !hascommoninds(siteinds(A, v), siteinds(B, v)) && error(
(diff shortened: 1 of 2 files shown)


src/abstractitensornetwork.jl
```diff
@@ -798,7 +798,11 @@ function hascommoninds(
   ::typeof(siteinds), A::AbstractITensorNetwork{V}, B::AbstractITensorNetwork{V}
 ) where {V}
   for v in vertices(A)
-    !hascommoninds(siteinds(A, v), siteinds(B, v)) && return false
+    if isempty(siteinds(A, v))
+      !isempty(siteinds(B, v)) && return false
+    else
+      !hascommoninds(siteinds(A, v), siteinds(B, v)) && return false
+    end
```
mtfishman, line 805: I'm confused by this change. It seems like this catches the case when `A` doesn't have site indices on vertex `v`, but `B` has site indices on vertex `v`. That seems like a case that `hascommoninds` should be working for. What goes wrong in `hascommoninds` in that case?

src/abstractitensornetwork.jl
```diff
@@ -815,9 +819,15 @@ function check_hascommoninds(
     )
   end
   for v in vertices(A)
-    !hascommoninds(siteinds(A, v), siteinds(B, v)) && error(
-      "$(typeof(A)) A and $(typeof(B)) B must share site indices. On vertex $v, A has site indices $(siteinds(A, v)) while B has site indices $(siteinds(B, v)).",
-    )
+    if isempty(siteinds(A, v))
```
mtfishman, line 822: It seems like `check_hascommoninds` could just call `hascommoninds(siteinds, A, B)` and return `nothing` if it is true and error if it is false.

src/abstractitensornetwork.jl
```diff
@@ -798,7 +798,11 @@ function hascommoninds(
   ::typeof(siteinds), A::AbstractITensorNetwork{V}, B::AbstractITensorNetwork{V}
 ) where {V}
```
mtfishman, line 799:
```suggestion
  ::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork
)
```

mtfishman: Same for `check_hascommoninds`.

test/test_opsum_to_ttn.jl
```diff
@@ -257,6 +256,10 @@ end
         Tmpo_lr = contract(Hsvd_lr)
       end
       @test Tttno_lr ≈ Tmpo_lr rtol = 1e-6
+      # ToDo: move to a more appropriate file, since this is not related to OpSum to TTN conversion
+      # test whether Hamiltonian contracts with itself (used to error due to missing siteindices)
+      @test isa(inner(Hsvd, Hsvd), Number)
```
mtfishman, line 261: Agreed. Also, it would be good to add tests for `hascommoninds` and `check_hascommoninds` with missing site indices.

src/abstractitensornetwork.jl
b-kloss: When `A` doesn't have site indices on vertex `v`, but `B` has site indices on vertex `v`, it will return `false`, while it will continue the loop if both don't have site indices on vertex `v`. The `false` would indeed also be caught in the `else` clause, so it would indeed be more transparent to write

src/abstractitensornetwork.jl
b-kloss: Yes.

mtfishman:
@b-kloss another approach here would be to just remove `hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` and `check_hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)`, I think basically they were copied over to this package from the MPS code in ITensors.jl where we were being more strict about inputs of certain functions but they may be more trouble than they are worth, and inherently may restrict the tensor network structure of certain functions more than we want in general.

src/abstractitensornetwork.jl
mtfishman: Gotcha, thanks. I see that is a strange corner case. See my comment https://github.com/mtfishman/ITensorNetworks.jl/pull/144#issuecomment-1983581282, I think it would be simpler to just remove these functions entirely, since they are likely doing more harm than good, and we can reassess what kind of checks we really want in this functions, if any.

mtfishman:
This was addressed in #148.

PR #145, opened 2024-03-18
Environments

JoeyT1994:
This PR adds the functionality `environment(tn::AbstractITensorNetwork, verbs::Vector)` to the codebase with returns the environment (related to the derivative) of an `ITensorNetwork` with respect to some subset of its vertices.

Algorithms are defined for getting the environments/derivates which currently include either exact contraction or belief propagation. This new functionality has been hooked to interface `environment(f::AbstractFormNetwork, state_vertices::Vector)`.

Testing it added for this within `test/test_forms.jl`


mtfishman:
My thinking on `derivative(f::AbstractFormNetwork, state_vertices::Vector)` is that for a form, you can only take the derivative with respect to the state vertices, so a separate function `derivative_state` wouldn't be necessary.

src/derivative.jl
```diff
@@ -0,0 +1,38 @@
...
+  return ITensor[contract(ψ_reduced; sequence, kwargs...)]
+end
+
+function derivative(
+  ::Algorithm"bp",
+  ψ::AbstractITensorNetwork,
+  verts::Vector;
+  (bp_cache!)=nothing,
+  update_bp_cache=isnothing(bp_cache!),
+  bp_cache_update_kwargs=default_cache_update_kwargs(bp_cache!),
```
mtfishman, line 27: Maybe just `cache!`, `update_cache`, and `update_kwargs`? Specifying it is `bp` seems redundant since it's already been specified that it is the `bp` through the algorithm selection.

test/test_forms.jl
```diff
@@ -48,4 +51,19 @@ using Random
 
   @test underlying_graph(ket_network(qf)) == underlying_graph(ψket)
   @test underlying_graph(operator_network(qf)) == underlying_graph(A)
+
+  ∂qf_∂v = only(derivative_state(qf, [v]))
+  @test (∂qf_∂v) * (qf[ket_vertex_map(qf)(v)] * qf[bra_vertex_map(qf)(v)]) ≈
```
mtfishman, line 56: Maybe we should have a shorthand `ket_vertex(qf, v)` for spelling `ket_vertex_map(qf)(v)`.

test/test_forms.jl
```diff
@@ -48,4 +51,19 @@ using Random
...
+    ITensors.contract(qf)
+
+  ∂qf_∂v_bp = derivative_state(qf, [v]; alg="bp", update_bp_cache=false)
+  ∂qf_∂v_bp = ITensors.contract(∂qf_∂v_bp)
+  ∂qf_∂v_bp = normalize!(∂qf_∂v_bp)
+  ∂qf_∂v = normalize!(∂qf_∂v)
```
mtfishman, line 62: Let's use `normalize` instead of `normalize!`, same down below.

test/test_forms.jl
```diff
@@ -48,4 +51,19 @@ using Random
 
   @test underlying_graph(ket_network(qf)) == underlying_graph(ψket)
   @test underlying_graph(operator_network(qf)) == underlying_graph(A)
+
+  ∂qf_∂v = only(derivative_state(qf, [v]))
+  @test (∂qf_∂v) * (qf[ket_vertex_map(qf)(v)] * qf[bra_vertex_map(qf)(v)]) ≈
+    ITensors.contract(qf)
+
+  ∂qf_∂v_bp = derivative_state(qf, [v]; alg="bp", update_bp_cache=false)
+  ∂qf_∂v_bp = ITensors.contract(∂qf_∂v_bp)
```
mtfishman, line 60:
```suggestion
  ∂qf_∂v_bp = contract(∂qf_∂v_bp)
```

mtfishman: If the issue is that `contract` conflicts with one defined in `Graphs`, we should instead do `using Graphs: nv, ...` to bring the functions from `Graphs` into the namespace that we are actually using in the tests.

src/derivative.jl
```diff
@@ -0,0 +1,38 @@
...
+)
+  if isnothing(bp_cache!)
+    bp_cache! = Ref(BeliefPropagationCache(ψ))
+  end
+
+  if update_bp_cache
+    bp_cache![] = update(bp_cache![]; bp_cache_update_kwargs...)
+  end
+
+  return incoming_messages(bp_cache![], setdiff(vertices(ψ), verts))
```
mtfishman, line 37: In general, shouldn't the derivative include tensors both from the factor and from the messages? I.e. to build a projected MPO, we need both the environment tensors and the MPO tensors in the region we are updating.

JoeyT1994: Yes, `incoming_messages` automatically takes care of this in the BP code and will include the tensors on the vertices from the region which are not specified in `verts`. Obviously that maybe suggests `incoming_messages` is a misnomer. So if you think there might be a more appropriate name like `environment_tensors` I am happy to revert to that

test/test_forms.jl
JoeyT1994: Yes this appears to be due to `WARNING: both Graphs and ITensors export "contract"; uses of it in module Main must be qualified` I get on compilation. Let me implement your suggestion.

test/test_forms.jl
JoeyT1994: I believe `normalize(T::ITensor)` is not defined and only the in-place version exists. I can do `∂qf_∂v /= norm(∂qf_∂v)` instead

test/test_forms.jl
JoeyT1994: Sounds sensible!

JoeyT1994:
@mtfishman In terms of `derivative(f::AbstractFormNetwork, state_vertices::Vector)` instead of `derivative_state(f::AbstractFormNetwork, state_vertices::Vector)`, the issue is that it conflicts with `derivative(itn::AbstractITensorNetwork, vertices::Vector)` in `src/derivative.jl` which I want `derivative_state` to forward to. We have the hierarchy `FormNetwork <: AbstractITensorNetwork` and so currently that would just create a circular loop.

I was originally defining  `derivative(itn::AbstractITensorNetwork, vertices::Vector)` and having `derivative(f::AbstractFormNetwork, state_vertices::Vector)` pass to that by calling `derivative(itensornetwork(f), vertices(f, state_vertices))` but the logic for building the `bp_cache` (if it is not defined) is in `derivative(itn::AbstractITensorNetwork, vertices::Vector)` and I want the default cache (default partitioning) to be different for an `AbstractFormNetwork vs an ` AbstractITensorNetwork`. I.e I have:

```
default_partitioning(ψ::AbstractITensorNetwork) = group(v -> v, vertices(ψ))
default_partitioning(f::AbstractFormNetwork) = group(v -> first(v), vertices(f))
```

at the moment.

src/derivative.jl
mtfishman: Yeah, we should think of a better name for that functionality, `incoming_messages` is misleading. `environment_tensors` or just `environment` seems reasonable. I think I prefer `environment`, we can see if that becomes confusing or unclear.

mtfishman:
Can't you specify a different default partitioning in `derivative(f::AbstractFormNetwork, state_vertices::Vector)`?

src/derivative.jl
JoeyT1994: Okay let me go with that!

JoeyT1994:
Yeah good point: I can make the choice `partitioning` a `kwarg` of `derivative(Alg::"BP", ...)` and have `derivative(f::AbstractForm, ...)` define that and pass it on

src/caches/beliefpropagationcache.jl
```diff
@@ -1,13 +1,18 @@
...
+  updated_messages = contract(contract_list; sequence, kwargs...)
+  updated_messages /= norm(updated_messages)
+  return ITensor[updated_messages]
 end
-default_message_update_kwargs() = (; normalize=true, contraction_sequence_alg="optimal")
 @traitfn default_bp_maxiter(g::::(!IsDirected)) = is_tree(g) ? 1 : nothing
 @traitfn function default_bp_maxiter(g::::IsDirected)
   return default_bp_maxiter(undirected_graph(underlying_graph(g)))
 end
+default_partitioning(ψ::AbstractITensorNetwork) = group(v -> v, vertices(ψ))
```
mtfishman, line 13: It's a bit long, but I think I would prefer `default_partitioned_vertices`, to be more explicit that it is outputting groups of vertices (rather than a partitioned tensor network, for example).

src/environment.jl
```diff
@@ -0,0 +1,42 @@
...
+  sequence = contraction_sequence(ψ_reduced; alg=contraction_sequence_alg)
+  return ITensor[contract(ψ_reduced; sequence, kwargs...)]
+end
+
+function environment(
+  ::Algorithm"bp",
+  ψ::AbstractITensorNetwork,
+  verts::Vector;
+  (cache!)=nothing,
+  partitions=default_partitioning(ψ),
```
mtfishman, line 29: It's a bit long, but I think I would prefer `partitioned_vertices` instead of `partitions`, to be more explicit that it is expecting groups of vertices (rather than a partitioned tensor network, for example).

src/environment.jl
```diff
@@ -0,0 +1,42 @@
...
+end
+
+function environment(
+  ::Algorithm"exact",
+  ψ::AbstractITensorNetwork,
+  vertices::Vector;
+  contraction_sequence_alg="optimal",
+  kwargs...,
+)
+  ψ_reduced = Vector{ITensor}(subgraph(ψ, vertices))
```
mtfishman, line 19: Shouldn't this be `setdiff(vertices(ψ), verts)` (where `verts` are the vertices input into the function)? I.e. the environment would be calculated by contracting the rest of the system.

JoeyT1994: Good spot. I was handling the logic for that in the wrong place (at the level of the `form network` and so hence why it didn't come up in testing). I fixed it now and handle the `setdiff` logic at the lowest possible level.

mtfishman:
Looks good, thanks! One more step towards BP-based DMRG/TDVP...

PR #146, opened 2024-03-19
Implement `inner(x,A,y)` for `TTN` via `BP`.

b-kloss:
This implements `inner(x,A,y)` for `x,A,y` all `AbstractTTN` with the same underlying graph, passing through the BP-backend. It addresses a performance issue of the previous implementation for certain `TTN`, leading to higher than necessary scaling in some situations (due to the traversal order). Thanks @JoeyT1994 for providing the `contract_with_BP` code.
Tests have been added to `test_forms.jl` to verify the correctness of the implementation.

This is not intended as a complete overhaul of the current implementations of `inner`, `expect` etc, which could be addressed in future PRs.

### DIFF
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -721,6 +721,35 @@ function contract_inner(
   return contract(tn; sequence)[]
 end
 
+function contract_with_BP(
+  itn::AbstractITensorNetwork; outputlevel=1, partitioning=group(v -> v, vertices(itn))
+)
+  @assert isempty(externalinds(itn))
+  bp_cache = BeliefPropagationCache(copy(itn), partitioning)
+
+  bp_cache = update(bp_cache)
+
+  pg = partitioned_itensornetwork(bp_cache)
+
+  if !is_tree(partitioned_graph(pg)) && outputlevel > 0
+    println("Partitioned graph is not a tree, result will be approximate!!")
+  end
+
+  log_numerator, log_denominator = 0, 0
+  for pv in partitionvertices(pg)
+    incoming_mts = incoming_messages(bp_cache, [pv])
+    local_state = ITensor[itn[v] for v in vertices(pg, pv)]
+    log_numerator += log(ITensors.contract(vcat(incoming_mts, local_state))[])
+  end
+  for pe in partitionedges(pg)
+    log_denominator += log(
+      ITensors.contract(vcat(message(bp_cache, pe), message(bp_cache, reverse(pe))))[]
+    )
+  end
+
+  return exp(log_numerator - log_denominator)
+end
+
 # TODO: rename `sqnorm` to match https://github.com/JuliaStats/Distances.jl,
 # or `norm_sqr` to match `LinearAlgebra.norm_sqr`
 norm_sqr(ψ::AbstractITensorNetwork; sequence) = contract_inner(ψ, ψ; sequence)
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -724,6 +724,17 @@ end
 function contract_with_BP(
   itn::AbstractITensorNetwork; outputlevel=1, partitioning=group(v -> v, vertices(itn))
 )
+  return contract_with_BP(
+    ComplexF64, itn::AbstractITensorNetwork; outputlevel, partitioning
+  )
+end
+
+function contract_with_BP(
+  T::Type,
+  itn::AbstractITensorNetwork;
+  outputlevel=1,
+  partitioning=group(v -> v, vertices(itn)),
+)
   @assert isempty(externalinds(itn))
   bp_cache = BeliefPropagationCache(copy(itn), partitioning)
 
@@ -739,15 +750,17 @@ function contract_with_BP(
   for pv in partitionvertices(pg)
     incoming_mts = incoming_messages(bp_cache, [pv])
     local_state = ITensor[itn[v] for v in vertices(pg, pv)]
-    log_numerator += log(ITensors.contract(vcat(incoming_mts, local_state))[])
+    log_numerator += log(complex(ITensors.contract(vcat(incoming_mts, local_state))[]))
   end
   for pe in partitionedges(pg)
     log_denominator += log(
-      ITensors.contract(vcat(message(bp_cache, pe), message(bp_cache, reverse(pe))))[]
+      complex(
+        ITensors.contract(vcat(message(bp_cache, pe), message(bp_cache, reverse(pe))))[]
+      ),
     )
   end
-
-  return exp(log_numerator - log_denominator)
+  res = exp(log_numerator - log_denominator)
+  return T(res)
 end
 
 # TODO: rename `sqnorm` to match https://github.com/JuliaStats/Distances.jl,
--- a/src/treetensornetworks/abstracttreetensornetwork.jl
+++ b/src/treetensornetworks/abstracttreetensornetwork.jl
@@ -365,20 +365,14 @@ end
 # Inner products
 #
 
-# TODO: implement using multi-graph disjoint union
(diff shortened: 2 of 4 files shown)


src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -365,20 +365,14 @@ end
...
+# TODO: Remove dispatch on AbstractTTN (unless we want to trigger warning for trees if BP is nonexact)  
+function inner(y::AbstractTTN, A::AbstractTTN, x::AbstractTTN; kwargs...)
   ydag = sim(dag(y); sites=[])
-  x = sim(x; sites=[])
-  O = ydag[root_vertex] * A[root_vertex] * x[root_vertex]
-  for v in traversal_order[2:end]
-    O = O * ydag[v] * A[v] * x[v]
-  end
-  return O[]
+  blf = BilinearFormNetwork(A, ydag, x)
```
mtfishman, line 371: I don't think we should be using a form here, that is mostly meant for optimization.

mtfishman: Is your goal to make a tensor network out of the network `y`, `A`, and `x`? If so, you can use `disjoint_union`.

src/abstractitensornetwork.jl
```diff
@@ -721,6 +721,48 @@ function contract_inner(
...
 
+function contract_with_BP(
+  itn::AbstractITensorNetwork; outputlevel=1, partitioning=group(v -> v, vertices(itn))
+)
+  return contract_with_BP(
+    ComplexF64, itn::AbstractITensorNetwork; outputlevel, partitioning
+  )
+end
+
+function contract_with_BP(
```
mtfishman, line 732: `contract_with_BP` isn't a good function name, what about overloading `ITensors.contract(::Algorithm"bp", kwargs...)` and running with `contract(itn; alg="bp", kwargs...)`.

src/abstractitensornetwork.jl
```diff
@@ -721,6 +721,48 @@ function contract_inner(
...
+  T::Type,
```
mtfishman, line 733: I'm confused by this part of the interface, why can you set the element type in this way?

src/abstractitensornetwork.jl
```diff
@@ -721,6 +721,48 @@ function contract_inner(
...
+  @assert isempty(externalinds(itn))
+  bp_cache = BeliefPropagationCache(copy(itn), partitioning)
+
+  bp_cache = update(bp_cache)
+
+  pg = partitioned_itensornetwork(bp_cache)
+
+  if !is_tree(partitioned_graph(pg)) && outputlevel > 0
+    println("Partitioned graph is not a tree, result will be approximate!!")
+  end
```
mtfishman, line 747: Let's remove this, I think this is inherent if someone selects that they are using BP.

src/abstractitensornetwork.jl
```diff
@@ -721,6 +721,48 @@ function contract_inner(
...
+
+  log_numerator, log_denominator = 0, 0
+  for pv in partitionvertices(pg)
+    incoming_mts = incoming_messages(bp_cache, [pv])
+    local_state = ITensor[itn[v] for v in vertices(pg, pv)]
+    log_numerator += log(complex(ITensors.contract(vcat(incoming_mts, local_state))[]))
```
mtfishman, line 753:
```suggestion
    log_numerator += log(complex(contract(vcat(incoming_mts, local_state))[]))
```

src/abstractitensornetwork.jl
```diff
@@ -721,6 +721,48 @@ function contract_inner(
...
+  end
+  for pe in partitionedges(pg)
+    log_denominator += log(
+      complex(
+        ITensors.contract(vcat(message(bp_cache, pe), message(bp_cache, reverse(pe))))[]
```
mtfishman, line 758:
```suggestion
        contract(vcat(message(bp_cache, pe), message(bp_cache, reverse(pe))))[]
```

src/abstractitensornetwork.jl
```diff
@@ -721,6 +721,48 @@ function contract_inner(
...
+
+  pg = partitioned_itensornetwork(bp_cache)
+
+  if !is_tree(partitioned_graph(pg)) && outputlevel > 0
+    println("Partitioned graph is not a tree, result will be approximate!!")
+  end
+
+  log_numerator, log_denominator = 0, 0
+  for pv in partitionvertices(pg)
+    incoming_mts = incoming_messages(bp_cache, [pv])
```
mtfishman, line 751: @JoeyT1994 so is this inherently outputting both the messages and the tensors in partition `pv` of the tensor network?

mtfishman: Never mind, I see those are being extracted in the next line.

src/abstractitensornetwork.jl
```diff
@@ -721,6 +721,48 @@ function contract_inner(
...
+    local_state = ITensor[itn[v] for v in vertices(pg, pv)]
```
mtfishman, line 752: Can't you use `pg[pv]` here?

src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -365,20 +365,14 @@ end
...
+  blf_scalar_bp = contract_with_BP(
```
mtfishman, line 372: If we just use `contract` here, then we can allow an `alg` keyword argument to get passed from `inner`, and users can control what algorithm is used to contract the network.

To handle partitioning in the more general case, something we could do is that instead of passing an `ITensorNetwork` and a partitioning, we pass a `PartitionedGraph` to `contract`, then it is up to the `contract` backends to contract the partitioned tensor network (either using or ignoring the partitioning).

src/treetensornetworks/abstracttreetensornetwork.jl
b-kloss: Yes, as far as I understand the `BilinearFormNetwork` just stacks the network here.

src/abstractitensornetwork.jl
b-kloss: Yes, but it may be good to allow to pass a flag like `assert_exact=true`, to check that the `BP` will indeed be exact for the contraction (for the case of contracting `TTN`s)?

mtfishman: Let's think about the design of that, but I don't think we have to address that in this PR. I agree with the sentiment that we don't want to secretly perform approximate contractions but give users the impression that we are doing exact contractions.

src/abstractitensornetwork.jl
mtfishman: Or `factor(bp_cache, pv)`?

src/treetensornetworks/abstracttreetensornetwork.jl
mtfishman: Ok, let's do that without going through `BilinearFormNetwork`.

src/abstractitensornetwork.jl
b-kloss: This is not the right way to do it. Basically, due to the `log` in this implementation, we need to promote the scalars of which the element type is taken to `complex`. Then the idea was that `contract_with_BP` returns a complex scalar, unless you specify a type, in which case it converts the scalar to the specified type. This should of course be handled differently for a proper implementation as `contract` with `BP` backend.

mtfishman: Ok. I think `contract_with_BP` as it stands is too far away from being a library function, so we should think about how to deal with that before continuing forward with this PR.

mtfishman:
Closing in favor of first writing a better designed `contract(::AbstractITensorNetwork)` function based on BP.

PR #147, opened 2024-03-26
Inner refactor

JoeyT1994:
This PR refactors the `inner` interface in `ITensorNetworks.jl`. Most importantly it adds algorithmic backends where "bp" and "exact" are currently supported.

For instance `inner(x::AbstractITensorNetwork, A::AbstractITensorNetwork, y::AbstractITensorNetwork; alg = default_alg)` and `inner(x::AbstractITensorNetwork, y::AbstractITensorNetwork; alg = default_alg)` will default to "bp" with a sensible partitioning if all the arguments are trees and "exact" otherwise.

The assumption in `inner(...)` is that the `externalinds` of the arguments are all aligned and thus the networks stack properly under `disjoint_union` and don't need mapping: although an optional `kwarg` link index map (which is defaulted to `sim`) is provided for `x` to avoid clashes. This is consistent with the interface for `ITensors.jl` at the moment.

Logarithmic versions of `inner`, i.e.  `loginner()` are also provided. In the "bp" case this is neatly defined and helps to avoid overflow/ underflow.  In the "exact" case it just calls `log(inner(...))` so perhaps it is a bit redundant. 

Testing is added via `test_inner.jl` and other standard tests which rely on `inner` will now direct to this and all pass. Additionally, tests in `test_treetensornetworks` will pass to this version and not the version in `src/treetensornetworksabstracttreetensornetwork.jl` unless the tensor network is specifically wrapped in the `TTN` type.

A few points of note/ things to think about resolving:

1. If all the elements of a tensor network `x` are 0 which means that `norm_sqr(x::AbstractITensorNetwork) = 0` then the BP method will return `NaN` as it will return the ratio 0/0. Hence this is an edge case that isn't well determined by the BP code. This is because the current BP backend computes `inner()` via a global method. A local method is possible to define for a tree to solve this (it is how Benedikt's code in `src/treetensornetworksabstracttreetensornetwork.jl` works), and we could then define `alg = "bp_global" ` and `alg = "bp_local_tree"` to let the user choose algorithms.
2. Currently `inner(alg::Algorithm"exact", ...)` passes to the `inner_network(...)` constructor which I cleaned up a bit: `inner_network` now forwards to `flatten_networks` but takes the `dag` on the first argument and its keyword arguments default to no flattening or combining (which are useful `kwargs` for "exact"). `inner(alg::Algorithm"bp",...)`, instead, internally builds the relevant `ITensorNetwork` via `disjoint_union`. This is because the `flatten_networks` builder results in confusing vertex names for more than two networks and thus it is difficult to parse when trying to define a good `default_partitioning`.

@mtfishman @b-kloss


mtfishman:
Can't the BP code detect if the norm is zero at some point? Presumably the message tensors become zero?

mtfishman:
Note that I was advocating for removing `hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` and `check_hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` here: https://github.com/mtfishman/ITensorNetworks.jl/pull/144#issuecomment-1983581282, in case that is relevant.

JoeyT1994:
> Note that I was advocating for removing `hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` and `check_hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` here: [#144 (comment)](https://github.com/mtfishman/ITensorNetworks.jl/pull/144#issuecomment-1983581282), in case that is relevant.

Hmm yeah I'm not sure what the bug is here because I don't see it locally on my machine. I thought something had changed in `ITensors.jl` and so it was importing something that no longer exists. Let me try to find the issue.

mtfishman:
I think probably for running BP on a norm network you can show that the only way a message can end up zero is if the network is zero or if it is initialized to zero, since each update is of the form of a CP-map.

mtfishman:
Note that I was advocating for removing `hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` and `check_hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` here: https://github.com/mtfishman/ITensorNetworks.jl/pull/144#issuecomment-1983581282, in case that is relevant.

> > Note that I was advocating for removing `hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` and `check_hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` here: [#144 (comment)](https://github.com/mtfishman/ITensorNetworks.jl/pull/144#issuecomment-1983581282), in case that is relevant.
> 
> Hmm yeah I'm not sure what the bug is here because I don't see it locally on my machine. I thought something had changed in `ITensors.jl` and so it was importing something that no longer exists. Let me try to find the issue.

It may be related to #1356, we moved all of the MPS/MPO code into an `ITensorMPS` submodule of `ITensors.jl`, which changed the namespaces of some internal functions, which probably included `check_hascommoninds` (i.e. `check_hascommoninds` probably isn't available from `ITensors`, and instead must be accessed from `ITensors.ITensorMPS`).

But if we have to circumvent issues around `check_hascommoninds` anyway, I think we may as well just remove it since I don't think it is serving a good purpose in ITensorNetworks.

JoeyT1994:
Okay I can remove it, shall I remove the calls to `check_hascommoninds` in the `treetensornetwork` code then? Presumably it is just being used for some assertions

mtfishman:
> Okay I can remove it, shall I remove the calls to `check_hascommoninds` in the `treetensornetwork` code then? Presumably it is just being used for some assertions

Yeah, thanks! Can you also remove `hascommoninds(::typeof(siteinds), A::AbstractITensorNetwork, B::AbstractITensorNetwork)` as well?

JoeyT1994:
@mtfishman  actually this seems to be happening with more than just `check_hascommoninds` due to the `ITensors.jl` changes, I can try to hunt them down

mtfishman:
Seems like we may as well just remove `src/treetensornetworks/solvers/deprecated` entirely instead of trying to update it.

mtfishman:
Also `src/treetensornetworks/deprecated_opsum_to_ttn.jl`.

mtfishman:
@JoeyT1994 reconsidering this, I'm concerned this PR is going to get messy with all of these other changes. Do you mind if I make a PR updating to the latest ITensor version, removing deprecated code, removing `check_hascommoninds`, etc. and then you can merge that into this PR?

JoeyT1994:
> I think probably for running BP on a norm network you can show that the only way a message can end up zero is if the network is zero or if it is initialized to zero, since each update is of the form of a CP-map.

Yes you are right that if it is a CP-map a message being zero directly implies on of those two things. Perhaps in a non CP-map there is a possibility the outgoing message ends up being zero by virtue of the incoming message being a zero eigenvector of the local tensor - but that seems a weird case and likely implies the tensor contracts to 0 too.

In principle if a message is 0 then the BP code should rapidly converge to 0 in a few sweeps of the code (as all the other messages end up zero) and it implies something about the tensornetwork. So it seems to me this isn't actually something that the BP code needs to `catch` for and is part of the algorithm. The exception is if we are trying to normalize the message tensors, in which case we should flag an error?

JoeyT1994:
@mtfishman I have updated the PR based on what we spoke about on Wednesday.

1. `inner(x, A, y)` passes to `BiLinearForm(A, x, y)` which does the mapping on `x` to bring it into the dual space. 
2. `logscalar()` is now generic to any algorithm where the cache has `scalar_factors(cache)` defined on it
3. The new namespace convention is used and updated for some of the tests `inner` applies to.

Let me know your thoughts.

src/abstractitensornetwork.jl
```diff
@@ -709,65 +709,37 @@ function flatten_networks(
...
-  sequence=nothing,
-  contraction_sequence_kwargs=(;),
-)
-  tn = inner_network(ϕ, ψ; combine_linkinds=true)
-  if isnothing(sequence)
-    sequence = contraction_sequence(tn; contraction_sequence_kwargs...)
-  end
-  return contract(tn; sequence)[]
+inner_network(x::AbstractITensorNetwork; kwargs...) = inner_network(x, x; kwargs...)
+function norm_sqr_network(ψ::AbstractITensorNetwork)
```
mtfishman, line 723: Is the plan for this to be implemented as `norm_sqr_network(x::AbstractITensorNetwork) = QuadraticForm(x)`?

src/abstractitensornetwork.jl
```diff
@@ -709,65 +709,37 @@ function flatten_networks(
...
+  end
+  ψket = rename_vertices(v -> (v, 2), data_graph(prime(ψdag; sites=[])))
+  ψψ = ITensorNetwork(union(ψbra, ψket))
+  for v in vertices(ψ)
+    if !isempty(commoninds(ψψ[(v, 1)], ψψ[(v, 2)]))
+      add_edge!(ψψ, (v, 1) => (v, 2))
+    end
+  end
+  return ψψ
+end
```
mtfishman, line 742: Reminder to delete this before merging. Why is `disjoint_union` slow?

src/abstractitensornetwork.jl
```diff
@@ -938,3 +910,29 @@ function ITensorMPS.add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork
...
+##       add_edge!(dg, v1 => v2)
+##     end
+##   end
+## 
+##   # TODO: Allow customization of the output type.
+##   ## return promote_type(typeof(tn1), typeof(tn2))(dg)
+##   ## return contract_output(typeof(tn1), typeof(tn2))(dg)
+## 
+##   return ITensorNetwork(dg)
+## end
```
mtfishman, line 938: Reminder to delete before merging.

src/caches/beliefpropagationcache.jl
```diff
@@ -208,7 +226,7 @@ function update(
   verbose=false,
   kwargs...,
 )
-  compute_error = !isnothing(tol)
+  compute_error = !isnothing(tol) && !hasqns(tensornetwork(bp_cache))
```
mtfishman, line 229: Is this just temporarily circumvent a QN issue?

src/caches/beliefpropagationcache.jl
```diff
@@ -14,12 +14,20 @@ end
...
+end
 
+#TODO: Define a version of this that works for QN supporting tensors
 function message_diff(message_a::Vector{ITensor}, message_b::Vector{ITensor})
   lhs, rhs = contract(message_a), contract(message_b)
-  return 0.5 *
-         norm((denseblocks(lhs) / sum(diag(lhs))) - (denseblocks(rhs) / sum(diag(rhs))))
+  tr_lhs = length(inds(lhs)) == 1 ? sum(lhs) : sum(diag(lhs))
+  tr_rhs = length(inds(rhs)) == 1 ? sum(rhs) : sum(diag(rhs))
+  return 0.5 * norm((denseblocks(lhs) / tr_lhs) - (denseblocks(rhs) / tr_rhs))
```
mtfishman, line 30: Could you explain this?

src/abstractitensornetwork.jl
```diff
@@ -938,3 +910,29 @@ function ITensorMPS.add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork
 end
 
 Base.:+(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork) = add(tn1, tn2)
+
+ITensors.hasqns(tn::AbstractITensorNetwork) = all([hasqns(tn[v]) for v in vertices(tn)])
```
mtfishman, line 914:
```suggestion
ITensors.hasqns(tn::AbstractITensorNetwork) = any(v -> hasqns(tn[v]), vertices(tn))
```
I think we should consider it to have QNs if any tensor has QNs.

src/boundarymps.jl
```diff
@@ -9,5 +9,5 @@ function contract_boundary_mps(tn::ITensorNetwork; kwargs...)
   end
   T = MPO([tn[i1, d2 - 1] for i1 in 1:d1])
   vR = MPS([tn[i1, d2] for i1 in 1:d1])
-  return inner(dag(vL), T, vR)[]
+  return ITensors.inner(dag(vL), T, vR)[]
```
mtfishman, line 12: Why not just `inner`?

src/caches/beliefpropagationcache.jl
```diff
@@ -40,8 +48,18 @@ function BeliefPropagationCache(tn, partitioned_vertices; kwargs...)
...
+function BeliefPropagationCache(
+  tn; partitioned_vertices=default_partitioned_vertices(tn), kwargs...
+)
+  return BeliefPropagationCache(tn, partitioned_vertices; kwargs...)
+end
+
+function cache(
+  alg::Algorithm"bp",
+  tn;
+  cache_construction_kwargs=default_cache_construction_kwargs(alg, tn),
```
mtfishman, line 60: I think it would be nicer to just use flattened keyword arguments here rather than a single keyword argument which accepts a NamedTuple of the keyword arguments.

src/caches/beliefpropagationcache.jl
```diff
@@ -244,3 +262,33 @@ end
...
+
+function edge_norms(bp_cache::BeliefPropagationCache, pes::Vector)
+  return [edge_norm(bp_cache, pe) for pe in pes]
+end
+
+function edge_norms(bp_cache::BeliefPropagationCache)
+  return edge_norms(bp_cache, partitionedges(partitioned_itensornetwork(bp_cache)))
+end
+
+function scalar_factors(bp_cache::BeliefPropagationCache)
```
mtfishman, line 292: I think a better name for this would be `scalar_factors_quotient` since it is outputting the factors of the numerator and denominator.

src/caches/beliefpropagationcache.jl
```diff
@@ -244,3 +262,33 @@ end
...
+  return vertex_norms(bp_cache, partitionvertices(partitioned_itensornetwork(bp_cache)))
+end
+
+function edge_norms(bp_cache::BeliefPropagationCache, pes::Vector)
+  return [edge_norm(bp_cache, pe) for pe in pes]
+end
+
+function edge_norms(bp_cache::BeliefPropagationCache)
+  return edge_norms(bp_cache, partitionedges(partitioned_itensornetwork(bp_cache)))
+end
```
mtfishman, line 290: I don't think `norm` is the right terminology for this.

Perhaps a better terminology would be `region_scalar`, and then:
1. `region_scalar(cache::BeliefPropagationCache, pv::PartitionVertex)` replaces `vertex_norm`.
2. `region_scalar(cache::BeliefPropagationCache, pv::PartitionEdge)` replaces `edge_norm`.
3. `vertex_scalars` replaces `vertex_norms`.
4. `edge_scalars` replaces `edge_norms`.

src/contract.jl
```diff
@@ -36,3 +40,66 @@ function contract_density_matrix(
...
+end
+
+function ITensors.scalar(
+  tn::Union{AbstractITensorNetwork,Vector{ITensor}}; alg="exact", kwargs...
+)
+  return scalar(Algorithm(alg), tn; kwargs...)
+end
+
+function logscalar(
+  tn::Union{AbstractITensorNetwork,Vector{ITensor}}; alg::String="exact", kwargs...
```
mtfishman, line 57:
```suggestion
  tn::Union{AbstractITensorNetwork,Vector{ITensor}}; alg="exact", kwargs...
```
We shouldn't constrain to `String` since a user could input an `Algorithm` directly.

src/contract.jl
```diff
@@ -36,3 +40,66 @@ function contract_density_matrix(
...
+)
+  return logscalar(Algorithm(alg), tn; kwargs...)
+end
+
+function logscalar(
+  alg::Algorithm"exact", tn::Union{AbstractITensorNetwork,Vector{ITensor}}; kwargs...
```
mtfishman, line 63: I think it would be better to define a type alias:
```julia
const AnyITensorNetwork = Union{AbstractITensorNetwork,Vector{ITensor}}
```
and then use `::AnyITensorNetwork` here and elsewhere.

src/contract.jl
```diff
@@ -36,3 +40,66 @@ function contract_density_matrix(
...
+)
+  s = scalar(alg, tn; kwargs...)
+  if s ≈ 0
+    tol = 1e-16
+    return -Inf
+  elseif isa(s, AbstractFloat) && s >= 0
+    return log(s)
+  else
+    return complex(s)
+  end
```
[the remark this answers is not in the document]
mtfishman: I believe `log` will output `-Inf` anyway for small `s` so I don't see why we need to handle that logic ourselves here.

src/contract.jl
```diff
@@ -36,3 +40,66 @@ function contract_density_matrix(
...
+)
+  return logscalar(Algorithm(alg), tn; kwargs...)
+end
+
+function logscalar(
+  alg::Algorithm"exact", tn::Union{AbstractITensorNetwork,Vector{ITensor}}; kwargs...
+)
+  s = scalar(alg, tn; kwargs...)
+  if s ≈ 0
+    tol = 1e-16
```
mtfishman, line 67: FYI you need to use the syntax `≈(s, 0; tol=1e-16)` here, `s ≈ 0 tol = 1e-16` only works inside of `@test`.

src/contract.jl
mtfishman: Though there is some question about when and why we should allow inputs of both `AbstractITensorNetwork` and `Vector{ITensor}`, I was generally trying to be conservative about that and require users to wrap `Vector{ITensor}` into `ITensorNetwork` themselves. Is there a particular reason for allowing that here?

src/contract.jl
```diff
@@ -36,3 +40,66 @@ function contract_density_matrix(
...
+
+  numerator_terms, denominator_terms = scalar_factors(cache![])
+  terms = vcat(numerator_terms, denominator_terms)
+  if any(≈(0), terms)
+    return -Inf
+  elseif all(t -> isa(t, AbstractFloat), terms) && all(>=(0), terms)
+    return sum(log.(numerator_terms)) - sum(log.((denominator_terms)))
+  else
+    return sum(log.(complex.(numerator_terms))) - sum(log.(complex.((denominator_terms))))
+  end
```
mtfishman, line 100: This logic seems very complicated, I'm sure we can simplify it by thinking about it in a nicer what, factorizing into functions, etc.

test/test_additensornetworks.jl
```diff
@@ -1,20 +1,15 @@
...
+using NamedGraphs: hexagonal_lattice_graph, rem_edge!, named_grid, NamedEdge
+using SplitApplyCombine: group
+using Random: seed!
+using ITensors: siteinds, scalar, apply, inner, op
+using LinearAlgebra: norm_sqr
 
 @testset "add_itensornetworks" begin
-  Random.seed!(5623)
-  g = named_grid((2, 3))
+  seed!(5623)
```
mtfishman, line 11: Let's keep it as `Random.seed!(5623)`.

test/test_additensornetworks.jl
```diff
@@ -1,20 +1,15 @@
...
-using NamedGraphs: hexagonal_lattice_graph, rem_edge!
-using Random
-using LinearAlgebra
-using SplitApplyCombine
+using ITensorNetworks
 
-using Random
+using NamedGraphs: hexagonal_lattice_graph, rem_edge!, named_grid, NamedEdge
+using SplitApplyCombine: group
+using Random: seed!
```
mtfishman, line 6:
```suggestion
using Random: Random
```

test/test_additensornetworks.jl
```diff
@@ -1,20 +1,15 @@
-using ITensorNetworks
-using ITensorNetworks: inner_network
 using Test
```
mtfishman, line 1:
```suggestion
using Test: @test, @testset
```

test/test_belief_propagation.jl
```diff
@@ -1,25 +1,35 @@
+using Test
```
mtfishman, line 1:
```suggestion
using Test: @test, @testset
```

test/test_apply.jl
```diff
@@ -1,31 +1,23 @@
+using Test
```
mtfishman, line 1:
```suggestion
using Test: @test, @testset
```

test/test_apply.jl
```diff
@@ -1,31 +1,23 @@
...
 using ITensorNetworks
```
mtfishman, line 2: Should remove this and list types and functions we are using.

test/test_belief_propagation.jl
```diff
@@ -1,25 +1,35 @@
...
 using ITensorNetworks
```
mtfishman, line 2: Should remove this and list types and functions we are using.

test/test_contraction_sequence.jl
```diff
@@ -5,6 +5,8 @@ using Random
 using Test
```
mtfishman, line 5:
```suggestion
using Test: @test, @testset
```

test/test_inner.jl
```diff
@@ -0,0 +1,49 @@
+using Test
+using ITensorNetworks
+
+using ITensorNetworks: logscalar, scalar, inner, loginner
+using ITensors: siteinds, dag
+using SplitApplyCombine: group
+using Graphs: SimpleGraph, uniform_tree
+using NamedGraphs: NamedGraph
+using Random: seed!
```
mtfishman, line 9:
```suggestion
using Random: Random
```

test/test_inner.jl
```diff
@@ -0,0 +1,49 @@
...
+
+@testset "Inner products, BP vs exact comparison" begin
+  seed!(1234)
```
mtfishman, line 12:
```suggestion
  Random.seed!(1234)
```

src/inner.jl
```diff
@@ -0,0 +1,143 @@
...
+  kwargs...,
+)
+  return loginner(Algorithm(alg), ϕ, ψ; kwargs...)
+end
+
+function ITensors.loginner(
+  ϕ::AbstractITensorNetwork,
+  A::AbstractITensorNetwork,
+  ψ::AbstractITensorNetwork;
+  alg=default_algorithm(ϕ, ψ),
```
mtfishman, line 69:
```suggestion
  alg=default_algorithm(ϕ, A, ψ),
```
Seems like it should allow accounting for `A` as well.

src/inner.jl
```diff
@@ -0,0 +1,143 @@
+using ITensors: inner, scalar, loginner
+using LinearAlgebra: norm, norm_sqr
+
+default_algorithm(tns::Tuple) = "bp"
```
mtfishman, line 4: I think this should be `default_contract_alg` to be more specific and also match the style that we use `alg` for the algorithm keyword argument.

src/inner.jl
```diff
@@ -0,0 +1,143 @@
...
+  if isnothing(sequence)
+    sequence = contraction_sequence(tn; contraction_sequence_kwargs...)
+  end
+  return scalar(tn; sequence)
+end
+
+function ITensors.loginner(
+  ϕ::AbstractITensorNetwork,
+  ψ::AbstractITensorNetwork;
+  alg=default_algorithm(ϕ, ψ),
```
mtfishman, line 59: Is this supposed to be a tuple input, i.e. `default_algorithm((ϕ, ψ))` (to be renamed `default_contract_alg` as suggested above).

test/test_apply.jl
```diff
@@ -1,31 +1,23 @@
...
-  VidalITensorNetwork
-using Test
-using Compat
-using ITensors
-using Metis
-using NamedGraphs
-using Random
-using LinearAlgebra
-using SplitApplyCombine
+  environment, update, BeliefPropagationCache, VidalITensorNetwork, norm_sqr_network
```
mtfishman, line 4: Please alphabetize, and put types first (in alphabetical order).

mtfishman: Same thing elsewhere.

test/test_belief_propagation.jl
```diff
@@ -1,25 +1,35 @@
...
   ising_network,
   split_index,
-  contract_inner,
   contract_boundary_mps,
   BeliefPropagationCache,
   tensornetwork,
   update,
   update_factor,
   environment,
   contract
```
mtfishman, line 12: Please alphabetize.

src/caches/beliefpropagationcache.jl
JoeyT1994: Yeah, the way we define `message_diff()` doesn't work for `QN` tensors at the moment, so I added that to avoid it.

src/caches/beliefpropagationcache.jl
JoeyT1994: If the message has only one index then `diag(A::ITensor)` is not defined and doesn't work - which I am not sure if it is intentional or not? Technically I might imagine it should return a vector of the elements of the `ITensor` if it has only one index

src/abstractitensornetwork.jl
JoeyT1994: Good point!

src/boundarymps.jl
JoeyT1994: No idea why it was that way. Have changed it

src/contract.jl
JoeyT1994: Yeah that is a good question. I think here I was taking the stance that lots of times those functions end up passing down to `contract(tns::Vector{ITensor})` so if the user wants `logscalar` on a `Vector{ITensor}` with `alg = "exact"` then they don't have to wrap it up as it will just get unwrapped.

I guess the main point is, depending on the algorithm, certain `scalar()` calls will unwrap to a `Vector{ITensor}` and pass down to something like `contract` whilst others will want to wrap up to an `ITensorNetwork` and pass to new code like `logscalar(alg::Algorithm"BP", ...)` so I was thinking of a generalized interface for calling either

src/contract.jl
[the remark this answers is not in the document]
JoeyT1994: Good point I didn't realize that was the case for `log(0.0)` in `Julia`

src/contract.jl
mtfishman: That's good to try and keep in mind but probably we should assess that more systematically and decide which kinds of functions should be able to accept `Vector{ITensor}` vs. `AbstractITensorNetwork` in a future PR. I could imagine changing many constraints from `AbstractITensorNetwork` to `AnyITensorNetwork` to make them more general when it makes sense, but let's think about that for a future PR.

src/caches/beliefpropagationcache.jl
JoeyT1994: Seem like reasonable names, I have changed to that.

src/contract.jl
JoeyT1994: Okay I will leave as is for now.

examples/contract_bp.jl
mtfishman, line 1: Seems like a mistake?

src/abstractitensornetwork.jl
```diff
@@ -709,65 +709,37 @@ function flatten_networks(
...
-  ψ::AbstractITensorNetwork;
-  sequence=nothing,
-  contraction_sequence_kwargs=(;),
-)
-  tn = inner_network(ϕ, ψ; combine_linkinds=true)
-  if isnothing(sequence)
-    sequence = contraction_sequence(tn; contraction_sequence_kwargs...)
-  end
-  return contract(tn; sequence)[]
+inner_network(x::AbstractITensorNetwork; kwargs...) = inner_network(x, x; kwargs...)
```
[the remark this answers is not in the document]
mtfishman: @JoeyT1994 this looks unresolved.

src/abstractitensornetwork.jl
mtfishman: @JoeyT1994 this looks unresolved.

src/caches/beliefpropagationcache.jl
mtfishman: That's just how `diag` is defined in Julia:
```julia
julia> using LinearAlgebra

julia> diag(randn(2, 2))
2-element Vector{Float64}:
 -0.7523337727717532
  1.1916982630642805

julia> diag(randn(2))
ERROR: ArgumentError: use diagm instead of diag to construct a diagonal matrix
Stacktrace:
 [1] diag(A::Vector{Float64})
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/generic.jl:445
 [2] top-level scope
   @ REPL[11]:1

julia> diag(randn(2, 2, 2))
ERROR: MethodError: no method matching diag(::Array{Float64, 3})

Closest candidates are:
  diag(::BitMatrix)
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/bitarray.jl:79
  diag(::ITensors.ITensor)
   @ ITensors ~/.julia/packages/ITensors/c1wvC/src/itensor.jl:1902
  diag(::UnitLowerTriangular)
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.2+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/triangular.jl:418
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[12]:1
```
but I agree that there should be a generic function, maybe called `diagonals`, for any array order that works as you say (I have a function `diagview` in `DiagonalArrays` that acts like that: https://github.com/ITensor/ITensors.jl/tree/v0.3.66/NDTensors/src/lib/DiagonalArrays).

But what happens if the message tensors have more than two indices? How should `message_diff` be defined? I think I'm just missing the broader picture/goal for this function.

src/caches/beliefpropagationcache.jl
```diff
@@ -251,3 +265,27 @@ end
...
+  incoming_mts = environment(bp_cache, [pv])
+  local_state = factor(bp_cache, pv)
+  return scalar(vcat(incoming_mts, local_state))
+end
+
+function region_scalar(bp_cache::BeliefPropagationCache, pe::PartitionEdge)
+  return scalar(vcat(message(bp_cache, pe), message(bp_cache, reverse(pe))))
+end
+
+function vertex_scalars(bp_cache::BeliefPropagationCache, pvs::Vector)
```
mtfishman, line 279:
```suggestion
function vertex_scalars(bp_cache::BeliefPropagationCache, pvs::Vector=partitionvertices(partitioned_tensornetwork(bp_cache)))
```

src/caches/beliefpropagationcache.jl
```diff
@@ -251,3 +265,27 @@ end
...
+  return [region_scalar(bp_cache, pv) for pv in pvs]
+end
+
+function edge_scalars(bp_cache::BeliefPropagationCache, pes::Vector)
```
mtfishman, line 283:
```suggestion
function edge_scalars(bp_cache::BeliefPropagationCache, pes::Vector=partitionedges(partitioned_tensornetwork(bp_cache)))
```

src/abstractitensornetwork.jl
JoeyT1994: Ideally I would implement it that way. The only issue for now is that currently there are a number of tests/ parts of the code which fail when you do this because they rely on `norm_sqr_network()` being two layer and not three layer (with the identity operator sandwiched). E.g. the `gauging` code relies on the message tensors being just `link_ind, link_ind` and I find the tests for `contraction_sequencs.jl` fail suddenly (probably because the size of the norm network gets too big for it to handle properly with optimum sequence finding)

src/abstractitensornetwork.jl
JoeyT1994: This is deleted now.

src/caches/beliefpropagationcache.jl
JoeyT1994: Any thoughts on this? Or do we leave as is and remind ourselves to come up with a better definition that works more streamlined with QN tensors and one index tensors.

examples/contract_bp.jl
JoeyT1994: My bad, good spot.

src/inner.jl
```diff
@@ -0,0 +1,143 @@
...
+  alg=default_contract_alg((ϕ, A, ψ)),
+  kwargs...,
+)
+  return loginner(Algorithm(alg), ϕ, A, ψ; kwargs...)
+end
+
+function ITensors.loginner(
+  alg::Algorithm"exact", ϕ::AbstractITensorNetwork, ψ::AbstractITensorNetwork; kwargs...
+)
+  return log(inner(alg, ϕ, ψ); kwargs...)
```
mtfishman, line 78: Hmm this doesn't seem like the right logic. The point of `loginner` is to try to avoid overflow if `inner` is very large. Should `inner` call `loginner`?

src/inner.jl
```diff
@@ -0,0 +1,143 @@
...
+end
+
+function ITensors.loginner(
+  alg::Algorithm"exact",
+  ϕ::AbstractITensorNetwork,
+  A::AbstractITensorNetwork,
+  ψ::AbstractITensorNetwork;
+  kwargs...,
+)
+  return log(inner(alg, ϕ, A, ψ); kwargs...)
```
mtfishman, line 88: Same comment here.

src/caches/beliefpropagationcache.jl
mtfishman: I see, I think it would be better to address the issue in `message_diff`.

src/caches/beliefpropagationcache.jl
mtfishman: It seems like a better design would be to allow customization of a norm for the message tensors. `tr` is just one choice, which I guess is good if it is a norm network but otherwise normalizing by `norm` of the message tensors may be better (and work for QNs).

src/abstractitensornetwork.jl
mtfishman: I see, thanks. Could you add a comment about that in the code, indicating ideally we would use `QuadraticFormNetwork` and explain what the problems with that are right now?

src/caches/beliefpropagationcache.jl
mtfishman: Also I'm sure we can fix issues with QNs in a naive way by just writing a specialized trace function as a loop over the diagonal values, which would also simplify this code.

src/abstractitensornetwork.jl
JoeyT1994: Sure. I will do that.

src/caches/beliefpropagationcache.jl
JoeyT1994: Okay let me come up with something that attempts to unify this and relies on the ability to pass a notion of `norm` as a keyword argument to the message diff function

src/inner.jl
JoeyT1994: We haven't actually defined an algorithm for computing `loginner(alg::Algorithm"exact", ....)` yet? So this is why I put the logic this way round for now.

But I see that it kind of defeats the point to have it because `log` is to deliberately avoid over/ underflow. I could remove this function for now and not support `loginner(alg::Algorithm"exact", ....)` until we actually incorporate an algorithm for it in? Then once we do we can pass `inner()` to it and take the `exp`.

mtfishman: Oh I see, sorry, I hadn't realized this was the `::Algorithm"exact"` version. I think it is ok to keep this as-is as a suboptimal stand-in.

src/caches/beliefpropagationcache.jl
JoeyT1994: Okay I refactored this so `message_diff()` excepts a keyword argument for how to compute the `message_norm` which defaults to `norm()` thus working for QNS and ITensors of dimension 1 in the default case.

Seems to be working well. Let me know what you think.

src/caches/beliefpropagationcache.jl
JoeyT1994: Okay I have addressed it below and will resolve this one.

src/caches/beliefpropagationcache.jl
mtfishman: Looks good, thanks! We could always bring back the previous trace-based norm as needed.

mtfishman:
I'll take another look through the PR.

src/contract.jl
mtfishman: I misunderstood your plan of action for this, my vote would be to restrict these functions to `tn::ITensorNetwork` for now and think about extending to other collections of ITensors like `::Vector{ITensor}` in future work.

test/test_belief_propagation.jl
```diff
@@ -21,149 +20,149 @@ using ITensorNetworks:
...
-  numerator = contract(vcat(env_tensors, ITensor[ψ[v], op("Sz", s[v]), dag(prime(ψ[v]))]))[]
-  denominator = contract(vcat(env_tensors, ITensor[ψ[v], op("I", s[v]), dag(prime(ψ[v]))]))[]
+  # bpc = BeliefPropagationCache(ψψ, group(v -> v[1], vertices(ψψ)))
+  # bpc = update(bpc)
+  # env_tensors = environment(bpc, [PartitionVertex(v)])
+  # numerator = contract(vcat(env_tensors, ITensor[ψ[v], op("Sz", s[v]), dag(prime(ψ[v]))]))[]
+  # denominator = contract(vcat(env_tensors, ITensor[ψ[v], op("I", s[v]), dag(prime(ψ[v]))]))[]
 
-  @test abs.((numerator / denominator) - exact_sz) <= 1e-14
+  # @test abs.((numerator / denominator) - exact_sz) <= 1e-14
```
mtfishman, line 87: I would rather not leave code commented out, we should delete it and bring it back later as needed. Same for down below.

Could you explain why it is being removed?

test/test_belief_propagation.jl
```diff
@@ -21,149 +20,149 @@ using ITensorNetworks:
...
+  # )
+
+  # env_tensors = environment(bpc, [v])
+  # numerator = contract(vcat(env_tensors, ITensor[ψOψ[v]]))[]
+  # denominator = contract(vcat(env_tensors, ITensor[ψψ[v]]))[]
+
+  # exact_sz =
+  #   contract_boundary_mps(ψOψ; cutoff=1e-16) / contract_boundary_mps(ψψ; cutoff=1e-16)
+
+  # @test abs.((numerator / denominator) - exact_sz) <= 1e-5
```
mtfishman, line 166: Delete instead of commenting out.

test/test_itensornetwork.jl
```diff
@@ -48,6 +50,8 @@ using NamedGraphs: NamedEdge, incident_edges, named_comb_tree, named_grid
 using Random: Random, randn!
 using Test: @test, @test_broken, @testset
 
+using ITensorNetworks: norm_sqr, neighbor_itensors
```
mtfishman, line 53: These should get merged into the list above.

test/test_belief_propagation.jl
JoeyT1994: Sorry yeah I had commented it out to focus on one part of the tests but forgot to change it back. Fixing that now.

test/test_belief_propagation.jl
JoeyT1994: Same as above

src/contract.jl
JoeyT1994: I see, I thought you meant to just leave as is. I will change the functions in this script to only apply to `tn::AbstractITensorNetwork`. I will remove `contract_density_matrix(contract_list::Vector{ITensor}, ...)` and just put it in the `test_belief_propagation.jl` file as that is the only place it is being used (to define an update routine for the message tensors based on `approx_itensornetwork`)

mtfishman: Makes sense, thanks. We probably need to refactor/redesign the approximate contraction backend anyway since it is a bit of a mess.

src/abstractitensornetwork.jl
```diff
@@ -728,66 +728,23 @@ function flatten_networks(
...
-  ψ::AbstractITensorNetwork;
-  sequence=nothing,
-  contraction_sequence_kwargs=(;),
-)
-  tn = inner_network(ϕ, ψ; combine_linkinds=true)
-  if isnothing(sequence)
-    sequence = contraction_sequence(tn; contraction_sequence_kwargs...)
-  end
-  return contract(tn; sequence)[]
+# TODO: We should make this pass to inner_network and then to BiLinearForm.
```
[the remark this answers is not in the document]
JoeyT1994: Yeah that's a good point. We could have a version of `BiLinearFormNetwork` which is only two layer when the identity operator is passed. And in that case calls like `operator_vertices(f)` could return empty or something.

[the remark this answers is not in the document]
mtfishman: I actually meant just as a post-processing of the output of `BilinearFormNetwork` and `QuadraticFormNetwork` for the sake of `inner_network` and `norm_sqr_network`, I think we probably still want the operator layer in the actual `BilinearFormNetwork` and `QuadraticFormNetwork` code since it is helpful when creating reduced maps for optimization.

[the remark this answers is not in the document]
JoeyT1994: I see. Although at that point that seems like more logic / lines of code than we have now in `norm_sqr_network()`?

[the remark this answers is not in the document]
mtfishman: We could wrap that in a helper function. Anyway, it is kind of besides the point for this PR, and as we discussed the main purpose of using `BilinearFormNetwork` or `QuadraticFormNetwork` to have a single place to handle logic about index maps.

mtfishman:
@JoeyT1994 I merged #155 into `main` and then tried merging it into this branch and fixing conflicts but it seems like it broke some BP tests... I had to fix some issues in the tests in #155 because of changes to the normalization of certain constructors but I'm not sure if that is related.

JoeyT1994:
Okay no worries let me pull that merge to my local machine and try to resolve the test failures.

mtfishman:
Thanks. That change was supposed to be pretty benign but perhaps the behavior of some of the constructors changed accidentally.

JoeyT1994:
Yeah the issue is that now `scalar(alg::Algorithm"bp", tn::AbstractITensorNetwork, ...)` seems to not be returning the correct answer any more. Can't quite figure out why yet.

JoeyT1994:
Okay actually I don't think its a BP issue:
The `inner_network` constructor (which passes to `BiLinearFormNetwork` and inserts an identity operator in the middle) is doing something different now than what it was, i.e.

```
ψ = random_tensornetwork(s; link_space=χ)
contract(inner_network(ψ, ψ)) == contract(ψ ⊗ prime(dag(ψ); sites=[]))
```

Now gives false.

JoeyT1994:
I see where the issue is, its the `delta_network(s)` constructor, it appears to be inserting matrices like:

```
1 0
0 0
```

on the vertices instead of identity matrices.

mtfishman:
I think that may have been an incorrect usage of `delta_network`, see my messages on Slack.

PR #149, opened 2024-03-27
Update style of `using` and `import`

mtfishman:


mtfishman:
In the latest commit, I updated the syntax for all of the remaining functions that were being overloading with `import M: f; function f(...) ... end` to the syntax `function M.f(...) ... end`.

For future reference, I used the function `replace_in_files` from a package I have been developing called [FileUtils.jl](https://github.com/mtfishman/FileUtils.jl), which recursively calls [Base.replace](https://docs.julialang.org/en/v1/base/strings/#Base.replace-Tuple{IO,%20AbstractString,%20Vararg{Pair}}) on every line of each file in a file path.

Here is the script I used:
```julia
# https://docs.julialang.org/en/v1/base/base/#Base.names
function imported_names(M::Module)
  return setdiff(names(M; imported=true), names(M; imported=false))
end

function imported_functions(M::Module)
  return map(f -> getfield(M, f), imported_names(M))
end

using FileUtils: replace_in_files

function prepend_imports(M::Module)
  fs = imported_functions(M)
  # foreach(f -> @show((parentmodule(f), f)), fs)
  src_path = joinpath(pkgdir(M), "src")
  @show src_path
  for f in fs
    function_name = last(split(string(f), "."))
    module_name = last(split(string(parentmodule(f)), "."))
    annotated_function_name = module_name * "." * function_name
    @show function_name, annotated_function_name
    replace_in_files(
      src_path,
      "function $(function_name)(" => "function $(annotated_function_name)(";
      recursive=true,
      ignore_dirs=[".git"],
      showdiffs=false,
    ) 
    replace_in_files(
      src_path,
      Regex("^$(function_name)\\(") => "function $(annotated_function_name)(";
      recursive=true,
      ignore_dirs=[".git"],
      showdiffs=false,
    )
  end
end

using ITensorNetworks: ITensorNetworks
prepend_imports(ITensorNetworks)
```

mtfishman:
@JoeyT1994 please try to be consistent with this style in new PRs, i.e. overloading functions defined in other modules with `function M.f(...) ... end` instead of `import M: f; `function f(...) ... end`, and also instead of writing `using M`, put a more explicit `using M: f` at the top of each file where `f` is being used.

mtfishman:
I've removed all of the `import` statements and some of the `using` statements, but I haven't changed over all of the `using` statements yet. I think I'll merge this as-is once tests pass and work on the `using` statements in a future PR.

`using` statements of examples and tests should also be updated from `using M` to `using M: f`, that will take some time but maybe we can update those incrementally, say as we make changes to a certain test or example file we also update that style.

Another style improvement will be to wrap each test file in a temporary module, which helps with namespace issues (i.e. helps to avoid having naming conflicts with previously defined variables when running tests), see for example https://github.com/ITensor/ITensors.jl/blob/v0.3.58/NDTensors/src/lib/BlockSparseArrays/test/test_basics.jl.

JoeyT1994:
> @JoeyT1994 please try to be consistent with this style in new PRs, i.e. overloading functions defined in other modules with `function M.f(...) ... end` instead of `import M: f; function f(...) ... end`, and also instead of writing `using M`, put a more explicit `using M: f` at the top of each file where `f` is being used.

Thanks @mtfishman I will make sure to try to stick with that style going forward and incorporate it into the `inner` PR

PR #157, opened 2024-04-12
Optimize `ttn_svd`

b-kloss:
This PR improves scaling of graph-related logic in `ttn_svd` and related functionality.
Tests `test/test_opsum_to_ttn.jl` are passing.


src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -38,6 +30,41 @@ function crosses_vertex(t::Scaled{C,Prod{Op}}, g::AbstractGraph, v) where {C}
   return v ∈ span(t, g)
 end
 
+# map all vertices `w` of `g` except for `v` to the incident edge of `v`
+# which lies in edge_path(g,w,v)
+function map_vertices_to_incident_edges(g::AbstractGraph, v, es)
```
mtfishman, line 35: I don't really understand what this function is doing. In what way is it a "map"?

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -13,21 +13,13 @@ using StaticArrays: MVector
...
 function span(t::Scaled{C,Prod{Op}}, g::AbstractGraph) where {C}
-  spn = eltype(g)[]
+  spn = Set{eltype(g)}()
   nterms = length(t)
-  for i in 1:nterms, j in i:nterms
-    path = vertex_path(g, ITensors.site(t[i]), ITensors.site(t[j]))
+  nterms == 1 && return Set([ITensors.site(t[1])])
+  for i in 1:nterms, j in (i + 1):nterms
+    path = Set(vertex_path(g, ITensors.site(t[i]), ITensors.site(t[j])))
     spn = union(spn, path)
```
mtfishman, line 23: Seems like this could be `union!`, not sure if it makes a performance difference in practice though.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -114,39 +143,50 @@ function ttn_svd(
...
+    edges = ordered_incident_edges(sites, v, es)
+
     # use the corresponding ordering as index order for tensor elements at this site
     dim_in = findfirst(e -> dst(e) == v, edges)
     edge_in = (isnothing(dim_in) ? [] : edges[dim_in])
     dims_out = findall(e -> src(e) == v, edges)
     edges_out = edges[dims_out]
 
+    which_incident_edge = map_vertices_to_incident_edges(sites, v, edges)
+    factor_to_edge(t) = which_incident_edge[ITensors.site(t)]
```
mtfishman, line 164: I find the name `factor_to_edge` confusing.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -38,6 +30,41 @@ function crosses_vertex(t::Scaled{C,Prod{Op}}, g::AbstractGraph, v) where {C}
...
+        deleteat!(subgraphs, i)
+        break
+      end
+    end
+  end
+  @assert isempty(subgraphs)
+  return Dict(zip(_vs, _es))
+end
+
+function ordered_incident_edges(g::AbstractGraph, v, ordered_edges)
```
mtfishman, line 61: Is the issue this is solving that `incident_edges` outputs a set of edges, but they might not be directed in the way you want? It seems a bit arbitrary to put all of that logic into a single function. Conceptually it seems better to split into two functions, like:
```julia
edges = incident_edges(g, v)
aligned_edges = align_edges(edges, reference_edges)
```
and call it like that in the source code. Regarding `align_edges`, here is a conceptually simple implementation:
```julia
julia> using Graphs

julia> edges = [Edge(1 => 2), Edge(2 => 3)]
2-element Vector{Graphs.SimpleGraphs.SimpleEdge{Int64}}:
 Edge 1 => 2
 Edge 2 => 3

julia> reference_edges = [Edge(2 => 1), Edge(3 => 2), Edge(4 => 3)]
3-element Vector{Graphs.SimpleGraphs.SimpleEdge{Int64}}:
 Edge 2 => 1
 Edge 3 => 2
 Edge 4 => 3

julia> intersect(reference_edges, Iterators.flatten((edges, reverse.(edges))))
2-element Vector{Graphs.SimpleGraphs.SimpleEdge{Int64}}:
 Edge 2 => 1
 Edge 3 => 2
```

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: It creates a `Dict` from vertices to incident edges.

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: We could just call it `which_incident_edge(t)`?

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: Thanks, that's cleaner! I changed it.

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: I renamed it to `vertices_to_incident_edges_dict` and added some comments to explain what it's doing.

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: I renamed it to `vertices_to_incident_edges_dict` and added some comments to explain what it's doing.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -38,6 +30,41 @@ function crosses_vertex(t::Scaled{C,Prod{Op}}, g::AbstractGraph, v) where {C}
   return v ∈ span(t, g)
 end
 
+function align_edges(edges, reference_edges)
+  return intersect(reference_edges, Iterators.flatten((edges, reverse.(edges))))
+end
+
+# return a dict from vertices `w` of `g`, except for `v`, to the incident edge of `v`
+# which lies in edge_path(g,w,v)
+function vertices_to_incident_edges_dict(g::AbstractGraph, v, incident_edges)
```
mtfishman, line 39: It feels like this function is trying to do too many things in a single function, it is difficult for me to wrap my head around what it is doing and what it's purpose is.

mtfishman: The description also feels underdefined. Can you try to refactor the function into smaller parts, and also rephrase what it is doing? Is the main idea that there is a path of vertices, and you want to output a dictionary from the vertices in that path to the incident edges of those vertices?

There is a function `boundary_edges` in `NamedGraphs.jl`: https://github.com/mtfishman/NamedGraphs.jl/blob/v0.1.25/src/Graphs/boundary.jl, does that relate to what this function is doing and can it help to simplify it? Perhaps it is some combination of `boundary_edges`, and then a `group` operation from https://github.com/JuliaData/SplitApplyCombine.jl which then groups those boundary edges by which ones share a vertex within the region? If it can be reframed in that way, ideally we wouldn't have it as a separate function, and instead just use a combination of `boundary_edges` and `group` inside `ttn_svd`. I think if a function is too specialized and not built around a simple and general conceptual framework that you can imagine using elsewhere, it isn't necessarily good to turn it into a separate function.

b-kloss: I wasn't aware of `boundary_edges`. I can probably simplify the code using that and move it into `ttn_svd`.

b-kloss: I simplified the code, and removed `span` in favor of reusing the `Dictionary` `which_incident_edges` to determine whether a term crosses a vertex. Unfortunately, it makes the code a little slower (but not by much), so once we have a faster algorithm for `span` we may want to switch the implementation back to the previous pattern of precomputing the `span` for every term.

src/treetensornetworks/opsum_to_ttn.jl
b-kloss: This has been simplified and moved into `ttn_svd` function body.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -6,38 +6,30 @@ using ITensors.NDTensors: Block, maxdim, nblocks, nnzblocks
...
-# determine whether an operator string crosses a given graph vertex
-function crosses_vertex(t::Scaled{C,Prod{Op}}, g::AbstractGraph, v) where {C}
-  return v ∈ span(t, g)
+function split_at_vertex(g::AbstractGraph, v)
+  _g = copy(g)
+  rem_vertex!(_g, v)
+  return Set.(connected_components(_g))
 end
 
+split_at_vertex(g::AbstractDataGraph, v) = split_at_vertex(underlying_graph(g), v)
```
mtfishman, line 31: Can you check if this can be removed? `connected_components(::AbstractDataGraph)` was fixed in https://github.com/mtfishman/DataGraphs.jl/pull/23.

mtfishman: If so, please also change the `[compat]` entry for `DataGraphs` to `v0.1.13` in the `Project.toml` file to make it clear we are relying on that version, which is the one with the fix for `connected_components(::AbstractDataGraph)`.

mtfishman:
@b-kloss looks good to me, thanks for the quick fix. I'll merge once that final comment is addressed.

emstoudenmire:
Nice to see this. Thanks, Benedikt

Project.toml
```diff
@@ -46,7 +46,7 @@ ITensorNetworksEinExprsExt = "EinExprs"
 AbstractTrees = "0.4.4"
 Combinatorics = "1"
 Compat = "3, 4"
-DataGraphs = "0.1.7"
+DataGraphs = "v0.1.13"
```
mtfishman, line 49:
```suggestion
DataGraphs = "0.1.13"
```

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -6,36 +6,25 @@ using ITensors.NDTensors: Block, maxdim, nblocks, nnzblocks
...
+  return intersect(reference_edges, align_edges(edges, reference_edges))
 end
 
-# determine whether an operator string crosses a given graph vertex
-function crosses_vertex(t::Scaled{C,Prod{Op}}, g::AbstractGraph, v) where {C}
-  return v ∈ span(t, g)
+function split_at_vertex(g::AbstractGraph, v)
+  _g = copy(g)
+  rem_vertex!(_g, v)
+  return Set.(connected_components(_g))
```
mtfishman, line 27:
```suggestion
  g = copy(g)
  rem_vertex!(g, v)
  return Set.(connected_components(g))
```
Just a style change, I prefer not prepending variable names with `_`, and there isn't a need for a new binding here.

PR #158, opened 2024-04-12
Refactor `sqrt_inv_sqrt`

JoeyT1994:
This PR removes the `sqrt_inv_sqrt` function in `apply.jl` in favor of a `map_itensor(f::Function, t::ITensors, args...)` added in the file `itensorsextensions.jl` in the new internal module `ITensorsExtensions` (where `src/ITensorsExt/utils.jl` is also now moved).

`map_itensor(f::Function, A::ITensor, ...)` uses the SVD to take a decomposition `U, S, V` of `A` and applies `f` over the diagonal elements of `S`. This uses the `sqrt_decomp` function from `ITensors.jl` (introduced in the PR https://github.com/ITensor/ITensors.jl/pull/1197/files) which correctly handles quantum numbers. The indices of the returned `ITensor` match those of the input `A`.

The `symmetric_factorize` function has been removed entirely from `apply.jl` as we can just call `factorize_svd(A::ITensor; ortho = "none") from `ITensors.jl` to do the same thing (again this was introduced in PR https://github.com/ITensor/ITensors.jl/pull/1197/files). 

Tests are included. @mtfishman these appear to show the square root / inverse being correctly taken by the SVD backend and I don't see any phase issues cropping up? 

A few other changes in this PR:
`src/tensornetworkoperators.jl` and `test/test_tnos.jl` have been removed and archived as they will be significantly refactored down the line.


src/ITensorsExtensions/ITensorsExtensions.jl
```diff
@@ -57,7 +59,13 @@ function map_itensor(
 )
```
mtfishman, line 59: It seems like this function should be called `map_singular_values`. Also I don't like using `first(inds(A))` as a default, generally ITensor functions should not depend on the ordering of the indices and that is a pretty arbitrary choice.

src/ITensorsExtensions/ITensorsExtensions.jl
```diff
@@ -57,7 +59,13 @@ function map_itensor(
...
   USV = svd(A, lind; kwargs...)
   U, S, V, spec, u, v = USV
-  S = map_diag(s -> f(s + regularization), S)
+
+  if !isnothing(regularization)
+    S = map_diag(s -> f(s + regularization), S)
+  else
+    S = map_diag(s -> f(s), S)
+  end
```
mtfishman, line 67: I think a nicer design would be:
```julia
f = isnothing(regularization) ? f : f ∘ Base.Fix2(+, regularization)
S = map_diag(f, S)
```
though honestly I think it would be better to remove `regularization` from this function and just expect users to change the mapping function themselves as needed.

src/ITensorsExtensions/ITensorsExtensions.jl
```diff
@@ -57,7 +59,13 @@ function map_itensor(
...
+
   sqrtDL, δᵤᵥ, sqrtDR = sqrt_decomp(S, u, v)
   sqrtDR = denseblocks(sqrtDR) * denseblocks(δᵤᵥ)
   L, R = U * sqrtDL, V * sqrtDR
```
mtfishman, line 71: Is all of this logic needed or can it just return `U * S * V`? Perhaps I am missing something about what this code is doing.

mtfishman:
I have to say that it doesn't sit well with me to use the SVD to perform square roots and inverses of positive semi-definite Hermitian matrices, I think it is safer and more direct to use a Hermitian eigendecomposition to ensure the gauge is fixed in the correct way.

src/ITensorsExtensions/ITensorsExtensions.jl
JoeyT1994: Yeah `map_eigenvalues` (assuming we switch to `Eigen` based on below) seems reasonable. 

Agreed there is some arbitraryness in the choice of `linds` but tbh I don't really like the current default of looking for prime and unprimed pairs. I feel like if I send this function a tensor with `2` indices it should automatically work by treating it as a matrix which in the prime case it won't unless the tensor has indices `i` and `i`.

In our case we are most often going to be passing message tensors to this function and assuming that we construct the norm network `ψψ` by `simming` the link indices `ψ` then we presumably want a default that handles that.

src/ITensorsExtensions/ITensorsExtensions.jl
JoeyT1994: Yeah that seems sensible to say that the `regularization` should be encoded in `f` already.

src/ITensorsExtensions/ITensorsExtensions.jl
mtfishman: But it is also arbitrary to have special behavior for order 2 tensors, and in general we don't like to do that since it makes it more difficult to write generic ITensor code. It is safer to expect users to input the indices explicitly and not have a default, this is meant to be a general function and not be tailored to the use case of BP (where in general message tensors won't be order 2, in which case the default you are suggesting won't work and have surprising behavior).

src/ITensorsExtensions/ITensorsExtensions.jl
```diff
@@ -52,16 +54,26 @@ invsqrt_diag(it::ITensor) = map_diag(inv ∘ sqrt, it)
 pinv_diag(it::ITensor) = map_diag(pinv, it)
 pinvsqrt_diag(it::ITensor) = map_diag(pinv ∘ sqrt, it)
 
-function map_itensor(
-  f::Function, A::ITensor, lind=first(inds(A)); regularization=nothing, kwargs...
+function map_eigenvalues(
```
mtfishman, line 57: Let's call this `map_eigvals` since Julia's `LinearAlgebra` library uses `eigvals(...)` to get the eigenvalues of a matrix.

src/ITensorsExtensions/ITensorsExtensions.jl
```diff
@@ -52,16 +54,26 @@ invsqrt_diag(it::ITensor) = map_diag(inv ∘ sqrt, it)
...
+  rinds = setdiff(inds(A), linds)
+
+  D, U = eigen(A, linds, rinds; ishermitian, kwargs...)
+  ul, ur = noncommonind(D, U), commonind(D, U)
+  ulnew = sim(ul)
+
+  Ul = replaceinds(U, (rinds..., ur), (linds..., ulnew))
+
+  D_mapped = map_diag(f, D)
+  D_mapped = replaceind(D_mapped, ul, ulnew)
```
mtfishman, line 74: We could factor out some of this indexing logic into an `eigendecomp` function as suggested here: https://github.com/ITensor/ITensors.jl/issues/341. I.e. if we define `eigendecomp`, you could then just rewrite this as:
```julia
function eigendecomp(a::ITensor, linds, rinds; kwargs...)
  d, vr = eigen(a, linds, rinds; kwargs...)
  # Reindexing logic
  return vr, d, vl
end

function map_eigvals(a::ITensor, inds...; kwargs...)
  vr, d, vl = eigendecomp(a, inds...; kwargs...)
  return vr * map_diag(f, d) * vl
end
```

mtfishman: Note that the non-Hermitian case is a bit more involved, as discussed in #341, since you have to compute the left and right eigenvectors with two calls to `eigen` and then normalize them against each other. We could leave that for future work, up to you.

src/ITensorsExtensions/ITensorsExtensions.jl
JoeyT1994: Yeah okay I can make it forward to `eigen` and inherit that behavior

src/ITensorsExtensions/ITensorsExtensions.jl
JoeyT1994: Sure thing.

src/ITensorsExtensions/ITensorsExtensions.jl
JoeyT1994: Yeah this seems like a good separation of logic to follow

JoeyT1994: I will also add a `#TODO` for the non-Hermitian case

src/ITensorsExtensions/ITensorsExtensions.jl
```diff
@@ -52,16 +54,26 @@ invsqrt_diag(it::ITensor) = map_diag(inv ∘ sqrt, it)
...
-  sqrtDL, δᵤᵥ, sqrtDR = sqrt_decomp(S, u, v)
-  sqrtDR = denseblocks(sqrtDR) * denseblocks(δᵤᵥ)
-  L, R = U * sqrtDL, V * sqrtDR
-  return L * R
+#TODO: Make this work for non-hermitian A
+function eigendecomp(A::ITensor, linds, rinds; ishermitian=false, kwargs...)
+  @assert ishermitian
+  D, U = eigen(A, linds, rinds; ishermitian, kwargs...)
+  ul, ur = noncommonind(D, U), commonind(D, U)
+  ulnew = sim(ul)
```
mtfishman, line 62: Do we need to make a new index here or can we just use `ul`? Ideally users would be able to make choices about `ul` after the fact or through `eigen`.

JoeyT1994: Yeah we can just use `ul` this was only necessary when I was introducing `sqrt_decomp` but should be redundant now

test/test_itensorsextensions.jl
```diff
@@ -0,0 +1,73 @@
...
+  delta,
+  inds,
+  noprime,
+  prime,
+  randomITensor,
+  replaceind,
+  replaceinds,
+  sim
+using ITensorNetworks.ITensorsExtensions: map_eigvals
+using Random
```
mtfishman, line 16:
```suggestion
using Random: Random
```

test/test_itensorsextensions.jl
```diff
@@ -0,0 +1,73 @@
+using ITensors:
```
mtfishman, line 1:
```suggestion
@eval module $(gensym())
using ITensors:
```

test/test_itensorsextensions.jl
```diff
@@ -0,0 +1,73 @@
...
+        I = replaceind(inv_P * P, new_ind, i)
+        @test I ≈ op("I", i) atol = 1e-12
+
+        inv_sqrtP = replaceind(inv_sqrtP, i', new_ind)
+        I = replaceind(inv_sqrtP * sqrtP, new_ind, i)
+        @test I ≈ op("I", i) atol = 1e-12
+      end
+    end
+  end
+end
```
mtfishman, line 73:
```suggestion
end
end
```

test/test_itensorsextensions.jl
```diff
@@ -0,0 +1,73 @@
...
+
+        inv_sqrtP = replaceind(inv_sqrtP, i, i')
+        I = inv_sqrtP * sqrtP
+        @test I ≈ delta(eltype, inds(I)) atol = 1e-12
+      end
+    end
+  end
+
+  @testset "Test map eigvals with QNS" begin
+    for eltype in [Float64, ComplexF64]
```
mtfishman, line 49:
```suggestion
    for eltype in (Float64, ComplexF64)
```

test/test_itensorsextensions.jl
```diff
@@ -0,0 +1,73 @@
...
+  replaceind,
+  replaceinds,
+  sim
+using ITensorNetworks.ITensorsExtensions: map_eigvals
+using Random
+using Test: @test, @testset
+
+Random.seed!(1234)
+@testset "ITensorsExtensions" begin
+  @testset "Test map eigvals without QNS" begin
```
mtfishman, line 21:
```suggestion
  @testset "Test map eigvals without QNS (eltype=$elt, dim=$n)" for elt in (Float32, Float64, Complex{Float32}, Complex{Float64}), n in [2, 3, 5, 10]
    ...
  end
```
This is usually how I write tests that loop over parameters, that way you can see what is being run in the test set log. Also good to run over single precision as well since it is easy to accidentally get that wrong and promote to double precision inside of source code.

src/apply.jl
```diff
@@ -11,73 +11,23 @@ using ITensors:
...
   dag,
   denseblocks,
+  factorize,
   hasqns,
   isdiag,
   noprime,
   prime,
   replaceind,
   replaceinds,
+  factorize_svd,
```
mtfishman, line 21:
```suggestion
  factorize,
  factorize_svd,
  hasqns,
  isdiag,
  noprime,
  prime,
  replaceind,
  replaceinds,
```
(alphabetized).

src/apply.jl
```diff
@@ -117,28 +67,44 @@ function full_update_bp(
...
-    sqrt_and_inv_sqrt.(envs_v1; ishermitian=true, cutoff, regularization)
-  sqrt_and_inv_sqrt_envs_v2 =
-    sqrt_and_inv_sqrt.(envs_v2; ishermitian=true, cutoff, regularization)
-  sqrt_envs_v1 = first.(sqrt_and_inv_sqrt_envs_v1)
-  inv_sqrt_envs_v1 = last.(sqrt_and_inv_sqrt_envs_v1)
-  sqrt_envs_v2 = first.(sqrt_and_inv_sqrt_envs_v2)
-  inv_sqrt_envs_v2 = last.(sqrt_and_inv_sqrt_envs_v2)
+  sqrt_envs_v1 = [
+    ITensorsExtensions.map_eigvals(
+      sqrt, env, first(inds(env)), last(inds(env)); cutoff, ishermitian=true
```
mtfishman, line 90: Instead of `first(inds(env)), last(inds(env))` I think it is better to have a check that `ndims(env) == 2` and then you can just use calls like `ITensorsExtensions.map_eigvals(sqrt, env, inds(env)...; ...)`.

src/ITensorsExtensions/ITensorsExtensions.jl
```diff
@@ -1,5 +1,5 @@
 module ITensorsExtensions
-using LinearAlgebra: LinearAlgebra, eigen, pinv
+using LinearAlgebra: LinearAlgebra, eigen, ishermitian, pinv
```
mtfishman, line 2:
```suggestion
using LinearAlgebra: LinearAlgebra, eigen, pinv
```
From what I can tell this isn't needed since the function `ishermitian(...)` isn't being used anywhere.

JoeyT1994: Good spot.

mtfishman:
Looks good, thanks! Good to go back and start cleaning up some of this code.

PR #160, opened 2024-04-22
Subspace expansion

wladiKrin:
This PR introduces a general and flexible subspace expansion interface. 

This allows users to perform a 2-site expansion alongside other (typically 1-site) routines, which has better scaling both in the bond dimension and the local Hilbertspace dimension than their 2-site analogs. 

The expansion works on general tree structures and can be seamlessly incorporated into existing algorithms by composing and chaining different local updaters via the `compose_updaters` function. 
Keyword arguments like `maxdim` and `cutoff` inside of `expander_kwargs` can be used to modify parameters specific to the expansion. Furthermore, a general, randomized svd-routine has been implemented, which can be used for the expansion and allows for the favorable, aforementioned scaling.

An example script can be found in the test folder.


src/solvers/subspace_expansions/subspace_expansion.jl
```diff
@@ -0,0 +1,149 @@
+#=
```
b-kloss, line 1: This file can be removed if we agree to focus this PR on the 2s subspace expansion.

mtfishman: Is this for global expansion? I think it would be better to narrow the scope of this PR to 2-site expansion.

src/ITensorNetworks.jl
```diff
@@ -49,6 +50,9 @@ include("solvers/solver_utils.jl")
 include("solvers/defaults.jl")
 include("solvers/insert/insert.jl")
 include("solvers/extract/extract.jl")
+include("solvers/subspace_expansions/linalg/standard_svd.jl")
+#include("solvers/subspace_expansions/linalg/rsvd_aux.jl")
+#include("solvers/subspace_expansions/linalg/rsvd_linalg.jl")
```
mtfishman, line 55: Why is this commented out?

mtfishman: To be deleted, as discussed in https://github.com/mtfishman/ITensorNetworks.jl/pull/137/#issuecomment-2080151198.

src/solvers/subspace_expansions/linalg/rsvd_aux.jl
mtfishman, line 1: To be deleted, as discussed in https://github.com/mtfishman/ITensorNetworks.jl/pull/137/#issuecomment-2080151198.

src/solvers/subspace_expansions/linalg/rsvd_linalg.jl
mtfishman, line 1: To be deleted, as discussed in https://github.com/mtfishman/ITensorNetworks.jl/pull/137/#issuecomment-2080151198.

PR #162, opened 2024-04-24
Refactor expect (single site)

JoeyT1994:
This PR refactors the `expect()` interface.

`expect(psi::AbstractITensorNetwork, op::Op)` now passes to the `environment(f::AbstractFormNetwork, ...)` function introduced in PR https://github.com/mtfishman/ITensorNetworks.jl/pull/145 and thus allows the `expect()` function to have a choice of algorithmic backend, with `"bp"` and `"exact"` currently supported.

Testing is added with `test_expect.jl` with some tests are removed from `test_beliefpropagation.jl` to avoid added.
Currently single-site expectation values are supported (so some tests which were computing two-site ones in `test_tebd.jl` have been temporarily disabled) but multi_site will be added in the near future.


src/expect.jl
```diff
@@ -1,57 +1,67 @@
...
-  for v in vertices
-    O = ITensor(Op(op, v), s)
-    Oψ = apply(O, ψ; cutoff, maxdim, ortho)
-    res[v] = inner(ψ, Oψ; alg="exact", sequence) / normψ²
-  end
-  return res
+  return expect(Algorithm(alg), ψ, args...; kwargs...)
+end
+
+function expect_internal(ψIψ::AbstractFormNetwork, op::Op; contract_kwargs=(;), kwargs...)
```
mtfishman, line 13: From what I can tell we can just call this `expect`, then users can call `expect` on a form network and `Op` if they want to.

test/test_expect.jl
```diff
@@ -0,0 +1,49 @@
+using Test
+using ITensorNetworks
```
mtfishman, line 2: Please wrap in a module and explicit using statements.

src/expect.jl
JoeyT1994: Makes sense

test/test_expect.jl
mtfishman: I see you wrapped it in a module, but there are still implicit using statements `using Test` and `using ITensorNetworks` rather than explicit using statements `using Test: @test, @testset` and `using ITensorNetworks: ...` (whatever additional types/functions being used by `ITensorNetworks` should be included with the list already started below).

JoeyT1994: Ah yeah! Should be resolved?

test/test_belief_propagation.jl
```diff
@@ -133,42 +70,8 @@ using Test: @test, @testset
   rdm = array((rdm * combiner(inds(rdm; plev=0)...)) * combiner(inds(rdm; plev=1)...))
   rdm /= tr(rdm)
 
-  eigs = eigvals(rdm)
+  eigs = Float64.(eigvals(rdm))
```
mtfishman, line 73: What's this conversion for?

src/expect.jl
```diff
@@ -1,57 +1,63 @@
-using ITensors.ITensorMPS: ITensorMPS, expect, promote_itensor_eltype, OpSum
+using Dictionaries: Dictionary, set!
+using ITensors: Op, op, contract, siteinds, which_op
+using ITensors.ITensorMPS: ITensorMPS, expect
+
+default_expect_alg() = "bp"
+
+function ITensorMPS.expect(ψIψ::AbstractFormNetwork, op::Op; contract_kwargs=(;), kwargs...)
```
mtfishman, line 7: I've lost track, does this function ever run BP or does it just use the message tensors as they are?

JoeyT1994: Depends. If you call it yourself and set `alg = "bp"` in the `kwargs` it will.

But when it gets called from `ITensorMPS.expect(ψ::AbstractITensorNetwork, op::Op; kwargs...)` to doesn't as that functions constructs the form network and does some cache logic before calling `ITensorMPS.expect(ψIψ::AbstractFormNetwork, op::Op; update_cache = false, kwargs...)`  internally

test/test_belief_propagation.jl
JoeyT1994: Was having some failing tests on Windows because of eigs that were `~-1E-17` so I refactored the test and this now essentially checks that all values in `eigs` are real up to machine precision.
But I should change this back to check the more explicit way.

src/expect.jl
mtfishman: I see, thanks, I missed the logic around `update_cache`, I'll take a closer look at that logic.

test/test_belief_propagation.jl
JoeyT1994: Have changed that now

src/formnetworks/abstractformnetwork.jl
```diff
@@ -69,11 +74,12 @@ end
...
 function environment(
   f::AbstractFormNetwork,
-  original_state_vertices::Vector;
+  original_vertices::Vector;
+  vertex_mapping_function=state_vertices,
   alg=default_environment_algorithm(),
   kwargs...,
 )
-  form_vertices = state_vertices(f, original_state_vertices)
+  form_vertices = vertex_mapping_function(f, original_vertices)
```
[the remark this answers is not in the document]
mtfishman: I see that is almost how it is designed, so I think my primary concern is using an if-statement.

Is it harmful to pass a `partitioned_vertices` keyword argument to every backend? Other contraction backends can ignore it if they want to, then this function would be generic to any `alg`. So my proposal would be:
```julia
function environment(
  f::AbstractFormNetwork,
  subvertices::Vector;
  alg=default_environment_algorithm(),
  kwargs...,
)
  partitioned_vertices = group(v -> original_state_vertex(f, v), vertices(f))
  return environment(
      tensornetwork(f), subvertices; alg, partitioned_vertices, kwargs...
    )
else
```
and it is up to the different backends to process `partitioned_vertices` as they see fit (and maybe ignore them if they want to).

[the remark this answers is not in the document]
mtfishman: We don't necessarily have to address the redesign of the form types in this PR. We took a best guess at the design of the form networks a bit in a vacuum and now have to see how those design choices play out as we try to write more sophisticated code with it. I think this PR is already exposing some warts of the current design.

src/expect.jl
```diff
@@ -1,57 +1,63 @@
...
+  v = only(op.sites)
+  ψIψ_v = ψIψ[operator_vertex(ψIψ, v)]
+  s = commonind(ψIψ[ket_vertex(ψIψ, v)], ψIψ_v)
+  operator = ITensors.op(op.which_op, s)
+  ∂ψIψ_∂v = environment(ψIψ, [v]; vertex_mapping_function=operator_vertices, kwargs...)
```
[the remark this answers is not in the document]
JoeyT1994: Yeah I felt like when I added in the `vertex_mapping_function` kwarg it was more just to try to make it "fit". I can refactor `environment(f::AbstractFormNetwork, ...)` to just take vertices of the form and then the expectation is to map state / op vertices before you pass them to the function

src/formnetworks/abstractformnetwork.jl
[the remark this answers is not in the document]
JoeyT1994: Yeah we can remove the `if` statement. I was probably trying to avoid sending a "dud" `kwarg` down to `environment` when `alg = "exact"` but you're right that it looks a bit messy and isn't really necessary and there's nothing so bad about `partitioned_vertices` being passed down if its not needed.

We could always define `default_partitioned_vertices(alg, tn)` or something and have it return `nothing` if the `alg` is not a partitioning-based `alg`.

JoeyT1994:
@mtfishman I refactored `environment(alg::Algorithm, ...)` in `src/environment.jl` so that the central BP version takes a `PartitionedGraph` as input and then `environment(alg::Algorithm"bp", psi::AbstractITensorNetwork, ...)` will wrap to a PartitionedGraph and send to the central version.

As `AbstractFormNetwork <: AbstractITensorNetwork` this means that those functions will automatically work on an `AbstractFormNetwork` and so we can now remove any re-definition of `environment` in the `src/formnetworks`

mtfishman:
Looks good, thanks!

Issue #117, opened 2024-01-09
Improvements in `OpSum` to `TTN` conversion


mtfishman:
Followup to #116:

- [ ] Replace `MatElem` and `QNArrElem` with [FillArrays.OneElement](https://juliaarrays.github.io/FillArrays.jl/stable/#FillArrays.OneElement).
- [ ] Rename `determine_val_type` to `coefficient_type`.
- [ ] Default `OpSum` coefficient type to `Float64`, require users to specify `OpSum{ComplexF64}` if they want that.
- [ ] Check / Improve compatibility with feature set of `OpSum` to `MPO` conversion in ITensors: support multi-site operators, ensure sorting comparisons work and are implemented consistently with ITensors implementation, implement all relevant sorting w.r.t to traversal order of tree instead of site-labels to ensure compatibility with arbitrary `vertextype`.
- [ ] Copy `ITensors` functions being used in `ttn_svd` like `ITensors.determineValType`, `ITensors.posInLink!`, `ITensors.MatElem`, etc. to `ITensorNetworks.jl` and update their style. Functions like `ITensors.which_op`, `ITensors.params`, `ITensors.site`, `ITensors.argument`, etc. that come from the `Ops` module related to `OpSum` shouldn't be copied over.
- [ ] Split off logic for building symbolic representation of TTNO into a separate function.
- [ ] Move `calc_qn` outside of `ttn_svd`.
- [ ] Use sparse matrix/array data structures or metagraphs for symbolic representation of TTNO (for example [NDTensors.SparseArrayDOKs](https://github.com/ITensor/ITensors.jl/tree/main/NDTensors/src/lib/SparseArrayDOKs) may be useful for that).
- [ ] Split off logic of grouping terms by QNs.
- [ ] Factor out logic for building link indices, make use of `IndsNetwork`.
- [ ] Refactor code logic to first work without merged blocks/QNs and then optionally merge and compress as needed.
- [ ] Support other compression schemes, like rank-revealing sparse QR.
- [ ] Implement sequential compression as opposed to the current method which uses parallel compression (i.e. right now it compresses each link index effectively independently) to improve performance.
- [ ] Allow compression to take into account operator information (perhaps by preprocessing by expanding in an orthonormal operator basis), not just coefficients.
- [ ] Handle starting and ending blocks in a more elegant way, for example as part of a sparse matrix.
- [ ] Handle vertices without any site indices (internal vertices, such as for hierarchical TTN).
[...]

PR #166, opened 2024-05-03
OpSum to TTN Refactor

emstoudenmire:
This PR refactors the `ttn_svd` function into three functions, `make_sparse_ttn`, `svd_bond_coefs`, and `compress_ttn`. 

Also the `calc_qn` function and associated cache has been turned into a "function object" called `TermQN`.

Other changes include:
* remove some unnecessary helper data structures in `opsum_to_ttn.jl`
* move code for `MatElem` and `QNArrElem` types into separate files
* add more `ITensors.Ops` functions to `using` statement in `opsum_to_ttn.jl`
* fix a missing keyword argument in `src/apply.jl`


src/ITensorNetworks.jl
```diff
@@ -40,6 +40,8 @@ include("solvers/local_solvers/contract.jl")
 include("solvers/local_solvers/linsolve.jl")
 include("treetensornetworks/abstracttreetensornetwork.jl")
 include("treetensornetworks/treetensornetwork.jl")
+include("treetensornetworks/matelem.jl")
+include("treetensornetworks/qnarrelem.jl")
 include("treetensornetworks/opsum_to_ttn.jl")
```
mtfishman, line 45: Probably we should make a `treetensornetworks/opsum_to_ttn` directory where we put these three files to make it clearer they are meant to be used with `opsum_to_ttn.jl`.

src/treetensornetworks/matelem.jl
```diff
@@ -0,0 +1,41 @@
+
+#
+# MatElem
+#
+
+struct MatElem{T}
```
mtfishman, line 6: We could just use `FillArrays.OneElement` for this can get rid of this type: https://juliaarrays.github.io/FillArrays.jl/stable/#FillArrays.OneElement. Not needed for this PR, I added a note about it to #117.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -40,92 +44,79 @@ Hamiltonian, compressing shared interaction channels.
...
-  coefficient_type::Type{<:Number},
-  os::OpSum,
-  sites0::IndsNetwork,
-  root_vertex;
-  mindim::Int=1,
-  maxdim::Int=typemax(Int),
-  cutoff=eps(real(coefficient_type)) * 10,
-)
-  linkdir_ref = ITensors.In   # safe to always use autofermion default here
+function remove_dups!(v::Vector)
```
mtfishman, line 51: Is this the same as https://docs.julialang.org/en/v1/base/collections/#Base.unique!?

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -40,92 +44,79 @@ Hamiltonian, compressing shared interaction channels.
...
   site_coef_done = Prod{Op}[]
-  # temporary symbolic representation of TTN Hamiltonian
-  tempTTN = Dict(v => QNArrElem{Scaled{coefficient_type,Prod{Op}},degrees[v]}[] for v in vs)
 
-  # build compressed finite state machine representation
-  for v in vs
-    # for every vertex, find all edges that contain this vertex
-    edges = align_and_reorder_edges(incident_edges(sites, v), es)
+  # Temporary symbolic representation of TTN Hamiltonian
+  tempTTN = Dict(
```
mtfishman, line 102:
```suggestion
  temp_ttn = Dict(
```

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -32,6 +29,13 @@ end
 # Tree adaptations of functionalities in ITensors.jl/src/physics/autompo/opsum_to_mpo.jl
 # 
 
+function determine_val_type(terms)
```
mtfishman, line 32:
```suggestion
function promote_coefficient_type(terms)
```
Also I think we can write this logic in a better way, for example:
```julia
function promote_coefficient_type(terms)
  return mapreduce(t -> typeof(coefficient(t)), promote_type, terms)
end
```

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -40,92 +44,79 @@ Hamiltonian, compressing shared interaction channels.
...
+  if pos == -1
+    pos = length(linkmap) + 1
+    linkmap[k] = pos
+  end
+  return pos
+end
 
-  sites = copy(sites0)  # copy because of modification to handle internal indices 
+function make_sparse_ttn(
+  os::OpSum, sites::IndsNetwork, root_vertex, coefficient_type, calc_qn
```
mtfishman, line 82:
```suggestion
  coefficient_type::Type, os::OpSum, sites::IndsNetwork; root_vertex, calc_qn
```
Conventionally types are input as the first arguments. Also I think it is clearer if we make `root_vertex` and `calc_qn` keyword arguments.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -236,34 +223,61 @@ function ttn_svd(
 
       push!(tempTTN[v], el)
     end
+    remove_dups!(tempTTN[v])
```
mtfishman, line 226:
```suggestion
    unique!(tempTTN[v])
```

src/treetensornetworks/opsum_to_ttn.jl
mtfishman: Seems like this data structure could be a `DataGraph` with a graph structure matching the `IndsNetwork`/`TTN` graph structure and a `SparseArrayDOK` stored on the vertices, where the number of dimensions is the degree of the graph and the elements are `Scaled{coefficient_type,Prod{Op}}`. Does that sound right to you?

I suppose one thing that needs to be stored is the meaning of each dimension of the `SparseArrayDOK` on the vertices since you want to know which dimension corresponds to which neighbor. So interestingly the best representation may be an `ITensor`, or maybe a [NamedDimsArray](https://github.com/ITensor/ITensors.jl/tree/main/NDTensors/src/lib/NamedDimsArrays) wrapping a `SparseArrayDOK`, where the dimension names are the edges of the graph.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -40,92 +44,79 @@ Hamiltonian, compressing shared interaction channels.
...
+  pos = get(linkmap, k, -1)
+  if pos == -1
+    pos = length(linkmap) + 1
+    linkmap[k] = pos
+  end
+  return pos
+end
 
-  sites = copy(sites0)  # copy because of modification to handle internal indices 
+function make_sparse_ttn(
```
mtfishman, line 81: Maybe `symbolic_ttn`?

mtfishman: Or `ttn_symbolic`, matching the convention that the one that compresses is called `ttn_svd`.

src/treetensornetworks/matelem.jl
emstoudenmire: Yes, this is for sure. Good to know.

src/treetensornetworks/qnarrelem.jl
```diff
@@ -0,0 +1,30 @@
+using StaticArrays: MVector
+
+#####################################
+# QNArrElem (sparse array with QNs) #
+#####################################
+
+struct QNArrElem{T,N}
```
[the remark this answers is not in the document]
emstoudenmire: I will have to think about what happens to the QN data but I hope it can be a OneElement also

src/treetensornetworks/opsum_to_ttn.jl
emstoudenmire: I think it is... will replace and see.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -236,34 +223,61 @@ function ttn_svd(
...
+
+  return tempTTN, inbond_coefs
+end
 
-    ITensorMPS.remove_dups!(tempTTN[v])
-    # manual truncation: isometry on incoming edge
+function svd_bond_coefs(
+  coefficient_type, sites, ordered_verts, ordered_edges, inbond_coefs; kws...
+)
+  Vs = Dict(e => Dict{QN,Matrix{coefficient_type}}() for e in ordered_edges)
```
mtfishman, line 235: This could be a `DataGraph` with that data on the edges of the graph.

I also wonder if `Dict{QN,Matrix{coefficient_type}}` could be a block diagonal `BlockSparseMatrix` where those matrices are the diagonal blocks and the QNs are the sector labels of the graded axes.

src/treetensornetworks/opsum_to_ttn.jl
```diff
@@ -413,10 +428,61 @@ function ttn_svd(
...
+
+  Vs = svd_bond_coefs(
+    coefficient_type, sites, ordered_verts, ordered_edges, inbond_coefs; kws...
+  )
+
+  Hflux = -calc_qn(terms(first(terms(os))))
+
+  T = compress_ttn(
+    coefficient_type, sites, ordered_verts, ordered_edges, Hflux, tempTTN, Vs
+  )
```
mtfishman, line 481: I'm not a fan of this kinds of functions that take many positional arguments, to me it is a warning sign that maybe the number of arguments could be cut down, the functions could be split up into smaller parts or refactored in some other way, or arguments could be moved to keyword arguments.

We could leave that for future work, some of it may be improved naturally by refactoring in terms of better data structures.

src/treetensornetworks/opsum_to_ttn.jl
emstoudenmire: I found something interesting here:
* when I used `sort!(v); unique!(v)` instead of `remove_dups!(v)` the tests failed even though `remove_dups!(v)` is doing exactly what the documentation for `unique!(v)` says (plus an initial sort)
* sure enough, I am finding that there are some bugs in `Base.unique!`
* one bug (in my opinion) is that `Base.unique!` calls `Base.isless` on objects in the vector even though the documentation says that `Base.unique!` relies on `isequal` for comparisons. When I input types that do not have `Base.isless` defines, `unique!` errors and does not even run
* the more serious bug is that `unique!` just gives clearly wrong results, and I tracked it down to the `Prod` type from `LazyApply`. I can't quite work out what the issue is yet, and it might be that `Prod` (or `Applied`) has a subtle bug in one of its comparison operators such as `Base.isequal` or `Base.isless` but I haven't found it yet.

So let's keep `remove_dups!` in this PR and treat removing it as a Issue so I can investigate more of what the actual bug is with `unique!` or whether it's a bug with `Prod` etc.

src/treetensornetworks/opsum_to_ttn.jl
emstoudenmire: Totally agree. I was thinking that this 'entanglement' between the different functions could be left for a future PR.

src/treetensornetworks/opsum_to_ttn.jl
emstoudenmire: Interestingly, this is a code improvement that reveals another possible bug, this time in `OpSum` or `Ops`. Benedikt noted this bug in a comment in the code too. Here's the bug:
```julia
using ITensors: OpSum
let
  terms = OpSum()
  terms .+= 1.0, "Sz",1,"Sz",2
  @show typeof(coefficient(first(terms)))
  #
  # Prints:
  #   typeof(coefficient(first(terms))) = ComplexF64
  #
  return
end
```
The `typeof(coefficient(t))` for any term in an `OpSum` is `ComplexFloat64` regardless of whether the original coefficient is real or not.

mtfishman: That's by design, `OpSum()` is defined as `OpSum{ComplexF64}()`.

That was to allow backwards compatibility with the old `OpSum`/`AutoMPO` type, otherwise:
```julia
  terms = OpSum()
  terms .+= 1.0 + 2.0im, "Sz",1,"Sz",2
```
would error.

If you want to check if the coefficients are numerically real (i.e. `Float64`, or `ComplexF64` with zero imaginary part) you would have to use a runtime check like `isreal` rather than check the type, or constrain the `OpSum` to be real with `OpSum{Float64}()`.

emstoudenmire: Good to know – I had forgotten if it was by design or not. I'll come up with a similar function to `promote_coefficient_type` above, then, that uses a more generic and modern style.

src/treetensornetworks/opsum_to_ttn.jl
emstoudenmire: Using `DataGraph`s could be really helpful. There are a lot of complicated dictionaries where the keys are things like edge=>QN pairs, which makes logical sense and is a carryover from the MPO code, but still looks too complicated. So maybe a `DataGraph` with a dictionary having QN keys on each edge could work and would be much cleaner.

src/treetensornetworks/opsum_to_ttn.jl
mtfishman: Interesting, too bad it doesn't "just work".

Could you rename `remove_dups!` to `_unique!` and put a comment about how we plan to use `Base.unique!` instead but it doesn't work right now for collections of `Prod`?

src/treetensornetworks/opsum_to_ttn.jl
emstoudenmire: Will rename to `_sort_unique!` or similar for now

emstoudenmire:
Ok so the status of this PR is that I've made all the changes you recommended.

I have *not* done the following:
* replacing `MatElem` with `FillArrays.OneElem` (turned out to be a little complicated to do for reasons I can explain)
* have not started using `SparseArraysDOK` yet (but that was for later anyway)
* have not started using `DataGraphs` yet either (also pending more planning and design considerations for future PR's)

mtfishman:
Sounds like a reasonable place to leave things, I'll merge once tests pass.

emstoudenmire:
Great. Yes this PR was mostly just about splitting the functions. The excessive arguments ought to come down once the data structures and graph analysis patterns are better (in future PR's).

PR #168, opened 2024-05-06
Bp optimal sequence finding

JoeyT1994:
This is short PR to update the `inner(alg::Algorithm"bp", ...)` and `expect(alg::Algorithm"bp", ...)` so that the calls to `contract()` inside them use `optimal_contraction_sequence` finding by default.

This speeds them up drastically and avoid memory issues that were appearing for larger networks due to non-optimal sequences being contracted.


mtfishman:
Looks good, thanks!

PR #169, opened 2024-05-06
Fix bug in QuadraticFormNetwork constructor when no operator present

JoeyT1994:
This PR fixes a bug in the `QuadraticFormNetwork(ket::AbstractITensorNetwork; kwarg...)` constructor where a `bra` variable was being passed that doesn't exist.

This PR should fix that and adds a quick test for the above constructor so that it would get caught in the future.


mtfishman:
Thanks!

Issue #164, opened 2024-04-26
Delete `flatten_networks`


mtfishman:
I don't think `flatten_networks` is needed now that we have some better concepts like forms, partitioned graphs, etc.

@JoeyT1994 seems like you already started removing it's usage in the tests in #162?

JoeyT1994:
Yeah I have been moving away from it.

I think, with the changes in #162, the only use I can see of it is in `test_contraction_sequence_to_graph` which was some code from a while back converting `contraction_sequences` to a graphical representation.

PR #170, opened 2024-05-07
Remove flatten networks and refactor test

JoeyT1994:
PR to remove `flatten_networks(...)` as it is not being used for anything meaningful anymore in the library.

The test where it was being used in `test_contraction_sequence_to_graph` now directly constructs a `flatened` network by just calling `ITensorNetwork(g; link_space =2)` avoiding any use of `siteinds`.

This should address Issue #164

### DIFF
--- a/src/abstractitensornetwork.jl
+++ b/src/abstractitensornetwork.jl
@@ -663,38 +663,6 @@ function split_index(
   return tn
 end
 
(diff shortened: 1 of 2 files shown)


mtfishman:
Great, thanks! Always nice to see code get deleted.

PR #172, opened 2024-05-09
Start using ITensorMPS.jl

mtfishman:


mtfishman:
Test failures related to `sortmergeterms` will be fixed by https://github.com/ITensor/ITensorMPS.jl/pull/4.

mtfishman:
More test failures caused by `ITensorMPS.determineValType` not be accessible. However, I don't think we should be using that function anyway, and indeed it is already being replaced in #166.

@emstoudenmire what's the status of #166? I'll what for that to get merged before merging this.

PR #175, opened 2024-05-12
Use DataGraphs in ttn_svd and other changes

emstoudenmire:
This PR contains mostly minor but helpful improvements to the `ttn_svd` function and its sub functions.

Here is a list of changes, from most to least significant:
* Use DataGraphs to hold the bond coefficients and sparse/symbolic vertex tensors returned from `make_symbolic_ttn`
* Reduce the number of arguments going into `make_symbolic_ttn`, `svd_bond_coefs`, and `compress_ttn`. (They weren't really sharing as much state as the code made it look, even compared to the current code following the last PR.)
* Significantly simplify `svd_bond_coefs` logic to just loop over all edges (it does a trivially parallel task of just forming and SVD'ing each edge coefficient matrix) rather than previous more complicated logic involving vertices and graph analysis.
* Removed other unneeded dictionaries.
* Use better graph analysis (`vertex_path`) to check which terms cross a given vertex. Could use `steiner_tree` but it seemed to have an issue with some disconnected components being left in the output.
* Remove `coefficient_type` function barrier in `ttn_svd` since now that function just calls out to other functions anyway.
* Simplify `MatElem` type keeping only code used for `OpSum` to `TTN` conversion. Will probably delete `MatElem` soon anyway in favor of `SparseArray`.

### DIFF
--- a/src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
+++ b/src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
@@ -137,13 +137,11 @@ function make_symbolic_ttn(
 
       # Only consider terms crossing current vertex
       crosses_v((s1, s2),) = (v ∈ vertex_path(sites, s1, s2))
-      v_crossed = any(crosses_v, partition(op_sites, 2, 1))
+      v_crossed = (v ∈ op_sites) || any(crosses_v, partition(op_sites, 2, 1))
       # TODO: seems we could make this just
       # v_crossed = (v ∈ vertices(steiner_tree(sites,op_sites))))
       # but it is not working - steiner_tree seems to have spurious extra vertices?
-
-      # If term doesn't cross vertex, skip it
-      (v_crossed || (v ∈ op_sites)) || continue
+      v_crossed || continue
 
       # Filter out ops that acts on current vertex
       onsite_ops = filter(t -> (site(t) == v), ops)
@@ -151,11 +149,8 @@ function make_symbolic_ttn(
 
       # Filter out ops that come in from the direction of the incoming edge
       incoming_ops = filter(t -> which_incident_edge[site(t)] == edge_in, non_onsite_ops)
-
       # Also store all non-incoming ops in standard order, used for channel merging
-      non_incoming_ops = filter(
-        t -> (site(t) == v) || which_incident_edge[site(t)] != edge_in, ops
-      )
+      non_incoming_ops = setdiff(ops, incoming_ops)
 
       # Compute QNs
       incoming_qn = term_qn_map(incoming_ops)
@@ -190,11 +185,11 @@ function make_symbolic_ttn(
       end
       for dout in dims_out
         out_edge = v_edges[dout]
-        out_op = filter(t -> which_incident_edge[site(t)] == out_edge, non_onsite_ops)
-        qn_outmap = get!(outmaps, out_edge => term_qn_map(out_op), Dict{Vector{Op},Int}())
+        out_ops = filter(t -> which_incident_edge[site(t)] == out_edge, non_onsite_ops)
+        qn_outmap = get!(outmaps, out_edge => term_qn_map(out_ops), Dict{Vector{Op},Int}())
         # Add outgoing channel
-        T_inds[dout] = pos_in_link!(qn_outmap, out_op)
-        T_qns[dout] = term_qn_map(out_op)
+        T_inds[dout] = pos_in_link!(qn_outmap, out_ops)
+        T_qns[dout] = term_qn_map(out_ops)
       end
       # If term starts at this site, add its coefficient as a site factor
       site_coef = one(coefficient_type)
--- a/src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
+++ b/src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
@@ -250,20 +250,18 @@ end
 function svd_bond_coefs(
   coefficient_type, sites, inbond_coefs; ordered_verts, ordered_edges, kws...
 )
-  Vs = Dict(e => Dict{QN,Matrix{coefficient_type}}() for e in ordered_edges)
-  for v in ordered_verts
-    edges = align_and_reorder_edges(incident_edges(sites, v), ordered_edges)
-    dim_in = findfirst(e -> dst(e) == v, edges)
-    if !isnothing(dim_in) && !isempty(inbond_coefs[edges[dim_in]])
-      for (q, mat) in inbond_coefs[edges[dim_in]]
-        M = toMatrix(mat)
-        U, S, V = svd(M)
-        P = S .^ 2
-        truncate!(P; kws...)
-        tdim = length(P)
-        nc = size(M, 2)
-        Vs[edges[dim_in]][q] = Matrix{coefficient_type}(V[1:nc, 1:tdim])
-      end
+  edge_data_eltype = Dict{QN,Matrix{coefficient_type}}
+  Vs = DataGraph(sites; edge_data_eltype)
+  for e in edges(sites)
+    Vs[e] = edge_data_eltype()
+    for (q, coefs) in inbond_coefs[e]
+      M = toMatrix(coefs)
+      U, S, V = svd(M)
+      P = S .^ 2
+      truncate!(P; kws...)
+      tdim = length(P)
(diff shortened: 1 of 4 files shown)


src/treetensornetworks/opsum_to_ttn/qnarrelem.jl
```diff
@@ -1,21 +1,22 @@
...
 #####################################
 # QNArrElem (sparse array with QNs) #
 #####################################
 
-struct QNArrElem{T,N}
-  qn_idxs::MVector{N,QN}
-  idxs::MVector{N,Int}
+struct QNArrElem{T}
+  qn_idxs::Vector{QN}
+  idxs::Vector{Int}
```
emstoudenmire, line 8: This change made it possible to have a dictionary of vectors of `QNArrElem`'s have a concrete eltype (of each vector). Otherwise the value type was just `Vector`.

mtfishman: I don't understand that, could you explain that more?

src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
```diff
@@ -88,130 +77,119 @@ function pos_in_link!(linkmap::Dict, k)
...
-      crosses_vertex || continue
+      ops = terms(term)
+      op_sites = ITensors.sites(term)
+
+      # Only consider terms crossing current vertex
+      crosses_v((s1, s2),) = (v ∈ vertex_path(sites, s1, s2))
+      v_crossed = (v ∈ op_sites) || any(crosses_v, partition(op_sites, 2, 1))
+      # TODO: seems we could make this just
+      # v_crossed = (v ∈ vertices(steiner_tree(sites,op_sites))))
+      # but it is not working - steiner_tree seems to have spurious extra vertices?
```
mtfishman, line 143: Can we define our own `tree_steiner_tree` function that is specific to tree graphs and calls `a_star` between each pair of specified vertices?

mtfishman: I believe @b-kloss also reported the `steiner_tree` didn't always give the minimal Steiner tree, even on trees. Probably `Graphs.jl` should have a special code path if the graph is a tree, that would make for a good PR. Also reporting an issue to `Graphs.jl` of an example where `steiner_tree` fails would be a good start.

b-kloss: Please check whether this is a performance regression. The previous code was an optimization to remove a significant performance bottleneck when using `vertex_path` logic. Note that the test cases in ITensorNetworks/test are not sufficiently hard to verify this. The bottleneck was evident for multi-orbital impurity problems with star-geometry on a fork graph with O(100-1000) sites on each leg of the fork.

mtfishman: I see `NamedGraphs.GraphsExtensions.vertex_path` may not be implemented be implemented in the best way: https://github.com/mtfishman/NamedGraphs.jl/blob/v0.6.0/src/lib/GraphsExtensions/src/abstractgraph.jl#L436-L449. Is it possible that optimizing that (say by using `a_star`) might fix the performance issue you were seeing on large graphs?

Additionally I wonder if we can rewrite the code in a nice way using `vertex_path`/`steiner_tree` and then fix any performance issues with a general solution like memoization to cache paths that have already been computed, i.e. find a way to get the best of both worlds in terms of simple code and good performance.

b-kloss: Concretely, I think the previous code was O(n_vertices^2 + ...) and O(n_terms) * O(average__n_site_per_term)).

b-kloss: I doubt that `a_star` is the way to go for trees. I think there should be an algorithm that takes a set of vertices on a tree and returns the span of said vertices (i.e. union of all possible vertex_paths among the set) that has cost O(vertices_in_span). Basically the vertex_path falls out directly from using a tree datastructure, you pick a root and then walk the tree towards the root from each vertex in the pair until you either hit the other vertex or a joint parent. You put the visited vertices in a dictionary, so that lookup is O(1) and the algorithm should overall scale as O(span). Since we are only interested in whether a given vertex is within the span, we can terminate once the site is found in the span, making the cost O(1)-O(span).
I imagine if this was efficiently implemented, the prefactors become small enough that it shouldn't be noticeably slower than the previous approach for any realistic Hamiltonian.

mtfishman: Thanks. So to summarize your comment, if we had an implementation of `steiner_tree(graph, vertices)` that scaled at worst as `O(nv(steiner_tree(graph, vertices)))` if `is_tree(graph)`, then used that function here, that could help to make the code simpler and also likely get good enough efficiency in the large tree limit.

I agree that my suggestion of repeatedly using `a_star` between the specified vertices may not be the absolute best way to implement it, I was more suggesting that as a simple first thing to try. The algorithm you suggest sounds good.

b-kloss: Yes, that summarizes it.
However, this PR still uses _which_incident_edge, so it already constructed an object that allows to determine crosses_vertex at cost O(n_sites_in_term).

mtfishman: Our goal now is more about simplifying the code as much as possible, while keeping in mind parts that may be bottlenecks and need to be optimized, and then once the code is simpler we will go back and optimize. It's part of a bigger rewrite based on https://github.com/ITensor/ITensorNetworks.jl/issues/117.

mtfishman: But it is helpful to hear about parts you found to be bottlenecks if they are not written properly.

src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
```diff
@@ -88,130 +77,119 @@ function pos_in_link!(linkmap::Dict, k)
...
-      only(boundary_edges(underlying_graph(sites), subgraph)) for subgraph in subgraphs
-    ]
-    _boundary_edges = align_edges(_boundary_edges, edges)
+    _boundary_edges = [only(boundary_edges(sites, subgraph)) for subgraph in subgraphs]
+    _boundary_edges = align_edges(_boundary_edges, v_edges)
     which_incident_edge = Dict(
       Iterators.flatten([
         subgraphs[i] .=> ((_boundary_edges[i]),) for i in eachindex(subgraphs)
       ]),
     )
```
b-kloss, line 133: Given the target to simplify the code structure, I suggest removing this and reverting the dependent implementation below to the state of #116. For an efficient implementation of `vertex_path` the original implementation using 'vertex/edge_path` is unlikely to cause a significant overhead.

src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
emstoudenmire: Thanks, Benedikt. That's really helpful background information as well as feedback about which graph algorithms we might want to use versus avoid. I'm on a learning curve with some of the graph mathematics. Based on this discussion, I'm going to create some performance benchmarks for me to test over time (is there already a collection of these somewhere?) and to check against the version before the previous PR. I definitely believe you that some of these changes are performance regressions so we plan to get that performance back as we iterate the design.

src/treetensornetworks/opsum_to_ttn/opsum_to_ttn.jl
emstoudenmire: Thanks for the link to #116. I will consider that, and a good plan here could be:
* revert the parts using `which_incident_edge` to use the older code, which looks quite nice and readable
* keep this code as a reference for performance, i.e. that we have to get back to at least this performance level in the future to consider the refactor 'done'
* try to find ways to either speed up calls involving `edge_path`, say, by using a different graph algorithm either inside `edge_path` or replacing `edge_path` with another algorithm altogether

We may end up making some new graph functions in the process which might end up repeating the `which_incident_edge` logic in the end, but the upshot is that it would be behind a more "self documenting" interface and properly pulled out as a separate function etc.

PR #178, opened 2024-05-14
Account for BP edge case where network evaluates to 0

JoeyT1994:
This PR fixed the `bp` edge case where the network contracts to the scalar `0` which would previously yield `Nan`.

This is achieved by:

1. Updating the `default_message_update` function to only divide by the `norm` if its not zero -> avoiding the creating of message tensors with `nan` in them.
2. Adding to `logscalar(alg::Algorithm"bp", tn::AbstractITensorNetwork)` a check if any of the denominators terms are zero. If this is the case then `log(tn)` should be `-Inf`. This is because, on a tree, if the product of `message(pe)` and `message(reverse(pe)` on any `partitionedge` is zero then the contraction of the tree must be zero (assuming the messages are converged).

A quick test is added.

@ryanlevy this should fix the issue with `bp` in `ITensorNumericalAnalysis.jl`

### DIFF
--- a/src/caches/beliefpropagationcache.jl
+++ b/src/caches/beliefpropagationcache.jl
@@ -19,9 +19,9 @@ default_message_norm(m::ITensor) = norm(m)
 function default_message_update(contract_list::Vector{ITensor}; kwargs...)
   sequence = optimal_contraction_sequence(contract_list)
   updated_messages = contract(contract_list; sequence, kwargs...)
-  n = norm(updated_messages)
-  if !iszero(n)
-    updated_messages /= norm(updated_messages)
+  message_norm = norm(updated_messages)
+  if !iszero(message_norm)
+    updated_messages /= message_norm
   end
   return ITensor[updated_messages]
 end
--- a/src/caches/beliefpropagationcache.jl
+++ b/src/caches/beliefpropagationcache.jl
@@ -19,7 +19,10 @@ default_message_norm(m::ITensor) = norm(m)
 function default_message_update(contract_list::Vector{ITensor}; kwargs...)
   sequence = optimal_contraction_sequence(contract_list)
   updated_messages = contract(contract_list; sequence, kwargs...)
-  updated_messages /= norm(updated_messages)
+  n = norm(updated_messages)
+  if !iszero(n)
+    updated_messages /= norm(updated_messages)
+  end
   return ITensor[updated_messages]
 end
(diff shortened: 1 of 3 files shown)


src/contract.jl
```diff
@@ -73,7 +73,11 @@ function logscalar(
     denominator_terms
   end
 
-  return sum(log.(numerator_terms)) - sum(log.((denominator_terms)))
+  if any(t -> iszero(t), collect(denominator_terms))
+    return -Inf
+  else
+    return sum(log.(numerator_terms)) - sum(log.((denominator_terms)))
+  end
```
mtfishman, line 80:
```suggestion
  any(iszero, denominator_terms) && return -Inf
  return sum(log.(numerator_terms)) - sum(log.((denominator_terms)))
```
Not sure if `collect(denominator_terms)` can be changed to `denominator_terms`, what's the type of `denominator_terms`?

JoeyT1994: It's a map from `partitionedges` to scalars so yeah I think the `collect()` can be dropped. Have changed that.

src/contract.jl
```diff
@@ -73,7 +73,11 @@ function logscalar(
     denominator_terms
   end
 
-  return sum(log.(numerator_terms)) - sum(log.((denominator_terms)))
+  if any(t -> iszero(t), denominator_terms)
+    return -Inf
+  else
+    return sum(log.(numerator_terms)) - sum(log.((denominator_terms)))
+  end
```
mtfishman, line 80: You removed `collect` but didn't simplify the code in the other ways that I suggested.

src/contract.jl
mtfishman: You removed `collect` but didn't simplify the code in the other ways that I suggested.

JoeyT1994: Ah I see sorry I missed that.

src/caches/beliefpropagationcache.jl
```diff
@@ -19,7 +19,10 @@ default_message_norm(m::ITensor) = norm(m)
 function default_message_update(contract_list::Vector{ITensor}; kwargs...)
   sequence = optimal_contraction_sequence(contract_list)
   updated_messages = contract(contract_list; sequence, kwargs...)
-  updated_messages /= norm(updated_messages)
+  n = norm(updated_messages)
+  if !iszero(n)
+    updated_messages /= norm(updated_messages)
+  end
```
mtfishman, line 25:
```suggestion
  message_norm = norm(updated_messages)
  if !iszero(message_norm)
    updated_messages /= message_norm
  end
```

PR #191, opened 2024-06-12
BP DMRG - Alpha Version

JoeyT1994:
This PR adds some files for doing one-site DMRG with belief propagation. 

First, support is added in `src/beliefpropagationdmrg/tensornetworkoperators.jl` for converting an `OpSum` to a vector of `ITensorNetworks` which are each trees and span the same set of vertices but have different edges. The sum of those `ITensorNetworks` is equal to the Hamiltonian represented by `OpSum`.

Then code is added for doing `bp` style dmg on those `ITensorNetworks` in the same vein as `alternating update` (in the style of using an inserter, extracter and updater)

@mtfishman This is very much a draft for you to look at - it isn't put into the alternating update framework yet because the 
algorithm is too unstable so I think we still need to figure out some details. In the case the graph `g` is a tree the code works fine. When it isn't then the code descends in energy well but then tends to "oscillate" near the true ground state energy but is not able to converge to a solution. Moreover, increasing the bond dimension above `chi = 2` tends to make things worse.

The BP error on the solution in the example `main.jl` (periodic ring of 24 sites) is incredibly small (the exact energy from contracting the result vs BP calculated energy agree to ~ 5 d.p.) but the energy oscillations are often on the order of, say, 3 d.p.


emstoudenmire:
A key question about the stability is does the core eigensolver solve a regular or generalized eigenvalue problem?

JoeyT1994:
The core eigensolver solves a regular eigenvalue problem here by switching to the `BP` gauge and inserting inverse square roots of the message tensors. This essentially sets the environment for the local state to the identity matrix within the BP approximation.

emstoudenmire:
Thanks. So then the eigensolver part sounds fine. I wonder then if it's the gauging procedure. While mathematically correct, the inverse square roots may cause precision issues. I ran into this in the past when implementing MPS in the "Vidal gauge" for parallel DMRG and it was quite difficult to deal with. I wonder if it's possible to work in a different gauge that avoids any inverses, or else implements the inverses by doing linear solves.

PR #192, opened 2024-06-16
Normalize

JoeyT1994:
This PR adds support for normalizing tensor networks with either a BP backend or an exact backend.

Specifically given an `ITensorNetwork` `tn` we can call `tn_normalized = normalize(tn; alg)` to enforce `tn_normalized * dag(tn_normalized) == 1`  within the framework of the desired algorithm.

This is particularly useful in the context of `alg = "bp"` as it stabilizes the fixed point of belief propagation such that the norm of the message tensors is stable when running subsequent bp iterations on `tn_normalized`. 

@mtfishman this is a routine that I am calling frequently in `bp_alternating_update` and so I thought I would add it. I also think it is generally useful when doing things like TEBD to keep the `bp_norm` more stable.


src/caches/beliefpropagationcache.jl
```diff
@@ -16,11 +16,11 @@ using SimpleTraits: SimpleTraits, Not, @traitfn
 default_message(inds_e) = ITensor[denseblocks(delta(i)) for i in inds_e]
 default_messages(ptn::PartitionedGraph) = Dictionary()
 default_message_norm(m::ITensor) = norm(m)
-function default_message_update(contract_list::Vector{ITensor}; kwargs...)
+function default_message_update(contract_list::Vector{ITensor}; normalize=true, kwargs...)
   sequence = optimal_contraction_sequence(contract_list)
   updated_messages = contract(contract_list; sequence, kwargs...)
   message_norm = norm(updated_messages)
-  if !iszero(message_norm)
+  if !iszero(message_norm) && normalize
```
mtfishman, line 23:
```suggestion
  if normalize && !iszero(message_norm)
```
so then it only checks if it is zero if you try to normalize (slightly more efficient).

src/caches/beliefpropagationcache.jl
```diff
@@ -157,6 +157,15 @@ function environment(bp_cache::BeliefPropagationCache, verts::Vector)
   return vcat(messages, central_tensors)
 end
 
+function factors(bp_cache::BeliefPropagationCache, vertices::Vector)
+  tn = tensornetwork(bp_cache)
+  return ITensor[tn[vertex] for vertex in vertices]
+end
```
mtfishman, line 163:
```suggestion
function factors(bp_cache::BeliefPropagationCache, vertices)
  tn = tensornetwork(bp_cache)
  return map(vertex -> tn[vertex], vertices)
end
```
I generally like using `map` better since it is a bit more generic, for example it preserves the type of the collection being input over, for example it outputs a `Dictionary` if `vertices` are stored in a `Dictionary`.

src/normalize.jl
```diff
@@ -0,0 +1,56 @@
...
+  tn = copy(tn)
+  cache![] = normalize_messages(cache![])
+  norm_tn = tensornetwork(cache![])
+
+  vertices_states = Dictionary()
+  for v in vertices(tn)
+    v_ket, v_bra = ket_vertex(norm_tn, v), bra_vertex(norm_tn, v)
+    pv = only(partitionvertices(cache![], [v_ket]))
+    vn = region_scalar(cache![], pv)
+    state = (1.0 / sqrt(vn)) * tn[v]
```
mtfishman, line 43:
```suggestion
    state = tn[v] / sqrt(vn)
```
`1.0` is implicitly of type `Float64` in Julia, so what you wrote would promote from single precision to double precision (say if you want to use `Float32`, for example on GPU).

src/caches/beliefpropagationcache.jl
```diff
@@ -309,3 +318,23 @@ end
 function scalar_factors_quotient(bp_cache::BeliefPropagationCache)
   return vertex_scalars(bp_cache), edge_scalars(bp_cache)
 end
+
+function normalize_messages(bp_cache::BeliefPropagationCache, pes::Vector{<:PartitionEdge})
+  bp_cache = copy(bp_cache)
+  mts = messages(bp_cache)
+  for pe in pes
+    n = region_scalar(bp_cache, pe)
+    me, mer = only(mts[pe]), only(mts[reverse(pe)])
```
mtfishman, line 327: I think it would be good to first normalize the messages against themselves and then normalize them against each other. That could help to make sure certain messages don't end up with very large or small norms. I realize in general during message tensor updates you would generally normalize but that might not be the case and large norms could accumulate (say when computing expectation values of extensive operators like Hamiltonians).

src/normalize.jl
```diff
@@ -0,0 +1,56 @@
...
+  return tn
+end
+
+function LinearAlgebra.normalize(
+  alg::Algorithm"bp",
+  tn::AbstractITensorNetwork;
+  (cache!)=nothing,
+  update_cache=isnothing(cache!),
+  cache_update_kwargs=default_cache_update_kwargs(cache!),
+)
```
mtfishman, line 25: So it seems like a basic design question here is if normalizing should refer to treating the `tn` as a state that should be normalized to 1 or as something such that you want the result of `contract(tn)` to be 1.

It seems reasonable to define it such that `tn` is a state where you want `contract(norm_network(tn))` to be 1 as you do here, however it may be good to write it in terms of an inner function that takes a tensor network and returns a new one where the tensors are scaled such that contracting it is 1. I can't think of a good name for that right now, but for the time being I'll refer to it as `rescale(tn::AbstractITensorNetwork)`, so `scalar(rescale(tn)) == 1` for any input `tn`, and the input has to be a closed network that evaluates to a scalar. Then we can just define `normalize(tn) = ket_network(rescale(norm_network(tn)))` or something like that.

The current implementation feels a bit too "in the weeds" dealing with quadratic forms, bras, kets, etc. and seems like something that could be abstracted and generalized.

mtfishman: Also defining a function like `rescale` then would be relevant for other kinds of networks like partition functions, where if you track the normalization factors then that gives the evaluation of the partition function.

mtfishman: Relatedly, `rescale(tn::AbstractITensorNetwork)` could be defined in two steps, one where it computes the local scale factors (I think there is already a function for that?) and then a next step where it just divides the factors by those scale factors, so the implementation could be a bit simpler by dividing it into multiple generic steps.

JoeyT1994: Yeah I see what you mean, that's a nice idea to split it apart like that. Will change it to do that

src/caches/beliefpropagationcache.jl
JoeyT1994: Good idea. I have made it so the messages get normalized first.

src/normalize.jl
JoeyT1994: Okay I split it apart. I couldn't quite use `rescale` as ubiquitously for the `alg = "bp"` case because the way that gets renormalized is on a partition vertex by partition vertex basis (so that all the free energy terms are 1). But I was still able to simplify to a reasonable degree.

JoeyT1994:
@mtfishman I have come back to this PR and updated it as it is a precursor to loop corrections (being able to normalize a BP cache). I have made it in-line with the latest `ITensorNetworks.jl` and based it around functionality that lets you `rescale` a `AbstractBeliefPropagationCache` both in terms of its messages and in terms of its partitions and utilizes whatever the definition of `region_scalar` is for that specific cache.

Once the `BoundaryMPSCache` type is supported this should automatically work on that too where the rescaling is done on the columns /rows of the tensor network based on the local `MPS-MPO-MPS` contractions. 

By the way I am noticing that the `test/test_contraction_sequence_to_graph.jl` is sometimes causing the tests to break, potentially due to a memory overflow. I am not sure of the origin but I do notice a lot of the code is based on the external `AbstractTrees.jl` package which hasn't been updated in a long time. Also the `contraction_sequence_to_graph` functions doenot seem to actually be used anywhere in the `ITensorNetworks.jl` library so I am wondering if it is needed or if we should consider it outdated and delete it?

mtfishman:
> @mtfishman I have come back to this PR and updated it as it is a precursor to loop corrections (being able to normalize a BP cache). I have made it in-line with the latest `ITensorNetworks.jl` and based it around functionality that lets you `rescale` a `AbstractBeliefPropagationCache` both in terms of its messages and in terms of its partitions and utilizes whatever the definition of `region_scalar` is for that specific cache.
> 
> Once the `BoundaryMPSCache` type is supported this should automatically work on that too where the rescaling is done on the columns /rows of the tensor network based on the local `MPS-MPO-MPS` contractions.

Sounds great, I'll take a look.

> By the way I am noticing that the `test/test_contraction_sequence_to_graph.jl` is sometimes causing the tests to break (it is the cause of the failing tests here), potentially due to a memory overflow. I am not sure of the origin but I do notice a lot of the code is based on the external `AbstractTrees.jl` package which hasn't been updated in a long time. Also the `contraction_sequence_to_graph` functions doenot seem to actually be used anywhere in the `ITensorNetworks.jl` library so I am wondering if it is needed or if we should consider it outdated and delete it?

I can't remember exactly what that was being used for, let's get rid of it for now if it is causing problems.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,69 @@ function update(
 )
   return update(Algorithm(alg), bpc; kwargs...)
 end
+
+function rescale_message(bp_cache::AbstractBeliefPropagationCache, partitionpair)
```
mtfishman, line 287: Is `partitionpair` a partition edge? If so we should call it `partitionedge`.

Also, do these `rescale` functions always normalize? Maybe we should use the terminology `normalize` instead of `rescale`.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,69 @@ function update(
...
+
+function logscalar(bpc::AbstractBeliefPropagationCache)
+  numerator_terms, denominator_terms = scalar_factors_quotient(bpc)
+  numerator_terms =
+    any(t -> real(t) < 0, numerator_terms) ? complex.(numerator_terms) : numerator_terms
+  denominator_terms = if any(t -> real(t) < 0, denominator_terms)
+    complex.(denominator_terms)
+  else
+    denominator_terms
+  end
```
mtfishman, line 343:
```suggestion
  if any(t -> real(t) < 0, numerator_terms)
    numerator_terms = complex.(numerator_terms)
  end
  if any(t -> real(t) < 0, denominator_terms)
    denominator_terms = complex.(denominator_terms)
  end
```
Same thing but I find it a bit easier to read.

src/caches/beliefpropagationcache.jl
```diff
@@ -106,3 +106,21 @@ function region_scalar(bp_cache::BeliefPropagationCache, pe::PartitionEdge)
   sequence = contraction_sequence(ts; alg="optimal")
   return contract(ts; sequence)[]
 end
+
+function rescale_messages(bp_cache::BeliefPropagationCache, pes::Vector{<:PartitionEdge})
+  bp_cache = copy(bp_cache)
+  mts = messages(bp_cache)
+  for pe in pes
+    me, mer = only(mts[pe]), only(mts[reverse(pe)])
```
mtfishman, line 114: Maybe I've lost track but when you call `only(mts[pe]), only(mts[reverse(pe)])`, is that assuming something about the messages, such as that the messages are just a single ITensor?

src/normalize.jl
```diff
@@ -0,0 +1,77 @@
...
+end
+
+function rescale(
+  alg::Algorithm"exact",
+  tn::AbstractITensorNetwork;
+  vs_to_rescale=collect(vertices(tn)),
+  kwargs...,
+)
+  logn = logscalar(alg, tn; kwargs...)
+  c = 1.0 / (exp(logn / length(vs_to_rescale)))
```
mtfishman, line 14:
```suggestion
  c = inv(exp(logn / length(vs_to_rescale)))
```
That is friendlier for other number types like `Float32` that are common on GPU, `1.0` is implicitly `Float64` and will promote `c` to `Float64`.

src/normalize.jl
```diff
@@ -0,0 +1,77 @@
+using LinearAlgebra
+
+function rescale(tn::AbstractITensorNetwork; alg="exact", kwargs...)
```
mtfishman, line 3: Is the rescaling being done here a normalization? If so I would call this `normalize`.

src/normalize.jl
```diff
@@ -0,0 +1,77 @@
...
+  return rescale(Algorithm(alg), tn; kwargs...)
+end
+
+function rescale(
+  alg::Algorithm"exact",
+  tn::AbstractITensorNetwork;
+  vs_to_rescale=collect(vertices(tn)),
```
mtfishman, line 10: It feels a little bit "off" to me to have this. I guess the idea is that you might only want to normalize a region of the network?

In that case maybe I would say that you should slice out that subgraph of the network, i.e. `tn_sub = supgraph(tn, subverts)`, and then slice it back into the original network.

I'm kind of on the fence about that but I would bias towards making functions simpler and composable with each other.

src/caches/abstractbeliefpropagationcache.jl
mtfishman: Finally, it is a bit funny to me for this function to be singular, i.e. `rescale_message` instead of `rescale_messages`, if it is normalizing the messages in both direction along the partition edge.

The fact that you aren't specifying the type of the input makes me think maybe the function name should be more specific, i.e. `rescale_partition_edge_messages` or something like that.

src/normalize.jl
mtfishman: Ignore this, I now remember that we deliberately had functions `rescale` and `normalize` to distinguish between closed networks and states.

src/normalize.jl
mtfishman: I see one reason for this is that `vs_to_rescale` is then passed along to the cache versions of these functions, and updating a subcache may be a bit trickier, but I think we should discuss if we can avoid this keyword argument at both levels in a nice way.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,69 @@ function update(
 )
   return update(Algorithm(alg), bpc; kwargs...)
 end
+
+function rescale_message(bp_cache::AbstractBeliefPropagationCache, partitionpair)
+  return rescale_messages(bp_cache, typeof(partitionpair)[partitionpair])
+end
+
+function rescale_messages(bp_cache::AbstractBeliefPropagationCache)
+  return rescale_messages(bp_cache, partitionpairs(bp_cache))
```
mtfishman, line 292: Should we rename `partitionpairs` to `partitionedges`? That name is throwing me off, I was confused by what it meant.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,69 @@ function update(
...
+
+    vn = region_scalar(bpc, pv)
+    if isreal(vn)
+      tn[first(pv_vs)] *= sign(vn)
+      vn *= sign(vn)
+    end
+
+    vn = vn^(1 / length(pv_vs))
+    for v in pv_vs
+      tn[v] /= vn
```
mtfishman, line 314: Similar to the design of `rescale_messages`, maybe the individual tensors should be normalized first so the scaling is distributed more evenly across tensors in the partition.

src/normalize.jl
mtfishman: Also it may be a bit tricky since the normalization would be based on messages/environment tensors coming into the region. But anyway, I think this is a subtle thing that we should discuss the design of, maybe there is another way to go with this.

src/caches/abstractbeliefpropagationcache.jl
mtfishman: Maybe that isn't always a good idea though since the scaling of the individual tensors may have some meaning, curious to hear your thoughts on that.

src/caches/beliefpropagationcache.jl
```diff
@@ -106,3 +106,21 @@ function region_scalar(bp_cache::BeliefPropagationCache, pe::PartitionEdge)
...
+    me, mer = normalize(me), normalize(mer)
+    n = dot(me, mer)
```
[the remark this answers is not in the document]
JoeyT1994: Good question actually. I believe want it to be `dot` here because when we do loop corrections we want to treat them as a "left" and a corresponding "right" eigenvector living in the virtual space of that bond and thus they are "duals" of each other.

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: Yeah I believe that was the thinking we came up with from a while back: `normalize` is for an open tensor network and `rescale` for a closed one

src/caches/beliefpropagationcache.jl
JoeyT1994: Yeah it's assuming they are a single ITensor but stored as a `Vector{<:ITensor}` (which is the typical storage for for the `BeliefPropagationCache` )

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: Yeah good point, I will rename it

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: Yeah at the moment these two operations are kept separate and then when `rescaling` the cache we do this
```
function rescale(bpc::AbstractBeliefPropagationCache; kwargs...)
  bpc = rescale_messages(bpc)
  bpc = rescale_partitions(bpc; kwargs...)
  return bpc
end
```
which will do what you said.

I can also see the argument for doing the `rescale_messages` on the relevant incoming messages inside `rescale_partition` too. That does, however, lead to the issue that messages may get rescaled multiple times as they are each incoming to two partitions so we would either waste compute or have to keep track of which ones are already rescaled.

So far the only time the scaling on the tensors has meaning is in the context of a tree tensor network when they have not been `rescaled` and represent the exact contraction of the given branch of the tree. But if you `rescale` the partitions then the `scale` on the message tensors will need to change too to reflect that so in some sense they are never really independent operations.

mtfishman: I meant maybe doing something like `tn[v] /= norm(tn[v])` as an initial step of this function, I didn't mean that we should rescale any messages in this function, I like that rescaling messages and factors are done separately.

src/caches/beliefpropagationcache.jl
mtfishman: Thanks for the reminder. It seems like we should abstract that away to some extent, for example maybe we could define `message_overlap(bp_cache, edge::PartitionEdge)`. That would hide the detail of how `Vector{<:ITensor}` is handled, and then also generalize to things like MPS/TNS messages.

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: Ah I see --- I misunderstood. That makes sense, I will make it do that first.

src/normalize.jl
mtfishman: Ok that's an interesting example, let's think about how to handle that kind of situation. More generally, there definitely is some freedom in how the rescaling of a factor/partition might get distributed across tensors in the factor/partition and the code should be pretty agnostic about that.

One thought would be to introduce an interface function/callback function `scale_factor(bp_cache::BeliefPropagationCache, pv::PartitionVertex, scale::Number)` where `scale` is `inv(region_scalar(bp_cache, pv))` that defaults to distributing the scale factor `scale` across each tensor in `pv` equally, but then could be specialized in certain cases to distribute it unequally. That is kind of related to the discussion in https://github.com/ITensor/ITensorNetworks.jl/pull/192#discussion_r2052417633 in that there are different choices for how tensors might be scaled and it isn't obvious what the best choice is in general.

src/caches/beliefpropagationcache.jl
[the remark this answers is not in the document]
mtfishman: I see, I think to some extent that is a matter of convention, but I think it would be better to store the left and right eigenvectors such that the can directly contract with each other, i.e. store one of them as the dual of what you are storing now.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,68 @@ function update(
 )
   return update(Algorithm(alg), bpc; kwargs...)
 end
+
+function rescale_message(bp_cache::AbstractBeliefPropagationCache, partitionedge)
+  return rescale_messages(bp_cache, typeof(partitionedge)[partitionedge])
```
mtfishman, line 288:
```suggestion
  return rescale_messages(bp_cache, [partitionedge])
```
The compiler should be able to figure out the element type, were you seeing issues with that?

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,68 @@ function update(
...
+
+  return bpc
+end
+
+function rescale_partitions(bpc::AbstractBeliefPropagationCache; kwargs...)
+  return rescale_partitions(bpc, collect(partitions(bpc)); kwargs...)
+end
+
+function rescale_partition(bpc::AbstractBeliefPropagationCache, partition; kwargs...)
+  return rescale_partitions(bpc, typeof(partition)[partition]; kwargs...)
```
mtfishman, line 326:
```suggestion
  return rescale_partitions(bpc, [partition]; kwargs...)
```

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: No I wasn't - will simplify it to let the compiler figure it out

src/caches/beliefpropagationcache.jl
JoeyT1994: Gotcha, I have now done it that way

JoeyT1994:
@mtfishman Yeah I'm not sure what's going on. I thought it was `contraction_sequence_to_graph` and so I removed that code (which is functionality not being used anywhere so I think we can get rid of it for now anyway) but now it fails on the next one.
I have commented out the `contraction_sequence` tests so maybe that will fix it but I wonder if its a memory issue caused by something not being cleaned up.
Tests pass locally for me on a Macbook so its hard to track down.

src/normalize.jl
JoeyT1994: Yeah a basic version of that behavior is also happening for the `bpc::BeliefPropagationCache` in that `rescale_partitions(bpc, ...)` takes a keyword argument telling it which tensors to do the rescaling on (and then it is done equally across all of those). 

These could be made more sophisticated by doing it based on weights etc but I guess that comes with more code complexity?

mtfishman: Right, though I'm wondering if there is just a different way to think about the design where we get both more flexibility but also a simple code design. I have a feeling there is if we step back from the problem a bit.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,76 @@ function update(
 )
   return update(Algorithm(alg), bpc; kwargs...)
 end
+
+function rescale_message(bp_cache::AbstractBeliefPropagationCache, partitionedge)
```
mtfishman, line 287: I still find it confusing that this is called `rescale_message` but it rescales multiple messages, one in each direction along the partition edge. An alternative could be:
```julia
function rescale_messages(bp_cache::AbstractBeliefPropagationCache, partitionedge::PartitionEdge)
```

JoeyT1994: Yeah that's a good point. I have done that rename.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +283,76 @@ function update(
 )
   return update(Algorithm(alg), bpc; kwargs...)
 end
+
+function rescale_messages(bp_cache::AbstractBeliefPropagationCache, partitionedge)
```
mtfishman, line 287:
```suggestion
function rescale_messages(bp_cache::AbstractBeliefPropagationCache, partitionedge::PartitionEdge)
```
I don't think it is a good idea to leave this argument unconstrained since it is really expecting a partition edge.

src/abstractitensornetwork.jl
```diff
@@ -935,6 +935,26 @@ function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
...
+""" Scale each tensor of the network by a scale factor on each vertex"""
+function scale!(tn::AbstractITensorNetwork, vertices_weights::Dictionary)
+  for v in keys(vertices_weights)
+    setindex_preserve_graph!(tn, vertices_weights[v] * tn[v], v)
+  end
+  return tn
+end
+
+""" Scale each tensor of the network via a function (vertex, ITensor) -> Number"""
+function scale!(tn::AbstractITensorNetwork, weight_function::Function)
```
mtfishman, line 947: We should put the function first since that is conventional for Julia functions that accept other functions. Also, it would allow using the syntax:
```julia
function scale!(tn) do v
  # ...
end
```

src/abstractitensornetwork.jl
```diff
@@ -935,6 +935,26 @@ function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
   return tn12
 end
 
+""" Scale each tensor of the network by a scale factor on each vertex"""
+function scale!(tn::AbstractITensorNetwork, vertices_weights::Dictionary)
+  for v in keys(vertices_weights)
+    setindex_preserve_graph!(tn, vertices_weights[v] * tn[v], v)
+  end
+  return tn
+end
```
mtfishman, line 944: I would design this the opposite way, where this version taking a Dictionary calls the version below taking a function, and the implementation is in the one taking a function.

src/abstractitensornetwork.jl
mtfishman: Also, it would make more sense to me for the function to just accept a vertex, not a vertex and a tensor (the tensor could be implicitly determined from the vertex if needed).

src/abstractitensornetwork.jl
```diff
@@ -935,6 +935,26 @@ function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
...
+function scale!(tn::AbstractITensorNetwork, weight_function::Function)
+  vs = collect(vertices(tn))
+  vertices_weights = Dictionary(vs, [weight_function(v, tn[v]) for v in vs])
+  return scale!(tn, vertices_weights)
+end
+
+function scale(tn, args...)
+  tn = copy(tn)
+  return scale!(tn, args...)
+end
```
mtfishman, line 956: This will have to be updated for the changes above, I would make the signature more specific (i.e. as specific as the in-place versions).

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +287,79 @@ function update(
 )
   return update(Algorithm(alg), bpc; kwargs...)
 end
+
+function scale!(bp_cache::AbstractBeliefPropagationCache, args...)
+  return scale!(tensornetwork(bp_cache), args...)
+end
```
mtfishman, line 293: See above.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -283,3 +287,79 @@ function update(
...
+
+function rescale_messages(bp_cache::AbstractBeliefPropagationCache)
+  return rescale_messages(bp_cache, partitionedges(bp_cache))
+end
+
+function rescale_partitions(
+  bpc::AbstractBeliefPropagationCache,
+  partitions::Vector;
+  verts_to_rescale::Vector=vertices(bpc, partitions),
```
mtfishman, line 308:
```suggestion
  vertices=vertices(bpc, partitions),
```
I think it will be clear those vertices are meant to be the ones being rescaled.

src/normalize.jl
```diff
@@ -0,0 +1,76 @@
+using LinearAlgebra
```
mtfishman, line 1: Can you make this explicit, i.e. `using LinearAlgebra: normalize, ...`?

src/abstractitensornetwork.jl
```diff
@@ -935,6 +935,31 @@ function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
   return tn12
 end
 
+""" Scale each tensor of the network via a function vertex -> Number"""
+function scale!(
+  weight_function::Function, tn::AbstractITensorNetwork; verts=collect(vertices(tn))
```
mtfishman, line 940:
```suggestion
  weight_function::Function, tn::AbstractITensorNetwork; vertices=collect(vertices(tn))
```
Just biasing towards not using shorthands (and it matches the name of the function from Graphs.jl so will be easier to remember).

JoeyT1994:
@mtfishman Okay I defined a `map_vertices_preserve_graph!` function as an alternative to `map_vertex_data_preserve_graph` and had the `scale!` function forward to that. 

A `map_vertex_data` function exists in `DataGraphs.jl` and works immediately on the `AbstractITensorNetwork` type so I removed that definition here.  However, `map_vertex_data_preserve_graph` is not yet defined in `DataGraphs.jl` . 

We should make a PR to `DataGraphs.jl` and define `map_vertex_data_preserve_graph` and  `map_vertices_preserve_graph!` and `map_vertices_preserve_graph` etc. Up to you whether you want me to do that before merging this PR or just leave as a TODO.

src/abstractitensornetwork.jl
```diff
@@ -935,6 +935,30 @@ function add(tn1::AbstractITensorNetwork, tn2::AbstractITensorNetwork)
...
+function scale!(tn::AbstractITensorNetwork, vertices_weights::Dictionary)
+  return scale!(v -> vertices_weights[v], tn; vertices=keys(vertices_weights))
+end
+
+function scale(weight_function::Function, tn; kwargs...)
+  tn = copy(tn)
+  return scale!(weight_function, tn; kwargs...)
+end
+
+function scale(tn, vertices_weights::Dictionary; kwargs...)
```
mtfishman, line 957:
```suggestion
function scale(tn::AbstractITensorNetwork, vertices_weights::Dictionary; kwargs...)
```

mtfishman:
> We should make a PR to `DataGraphs.jl` and define `map_vertex_data_preserve_graph` and `map_vertices_preserve_graph!` and `map_vertices_preserve_graph` etc. Up to you whether you want me to do that before merging this PR or just leave as a TODO.

I think the reason it wasn't defined in DataGraphs.jl is because ITensorNetworks are quite special, since the vertex metadata (the ITensors) implicitly determine the graph structure. That's not the case for most other data graphs I can think of. For that reason, DataGraphs.jl just assumes mapping/modifying the vertex data doesn't change the graph structure. So, I think it is fine to keep this in ITensorNetworks.jl for now.

PR #193, opened 2024-06-25
Fix BP convergence metric for complex networks

JoeyT1994:
This PR alters the `message_diff` function in BP to do a better convergence test.

Specifically, it defines it as `1 - |<m_{e, t+1}|m_{e, t}>|^{2}` where `|m_{e, t}>` is the message tensor (reshaped as a vector) on edge e at iteration `t`  and `|m_{e, t}>` is the message tensor (reshaped as a vector) on edge e at the subsequent iteration.

It is analogous to the infidelity measure for a quantum wavefunction and a strict lower bound on the trace distance,
i.e. `1 - |<m_{e, t+1}|m_{e, t}>|^{2} <= 0.5*Tr(|m_{e, t}><m_{e, t}| - |m_{e, t+1}><m_{e, t+1}|)`

Importantly, it is independent of a "phase" factor `e^i\theta` difference between the two messages, something which I believe the Euclidean distance isn't and therefore would run into trouble with complex numbers.

### DIFF
--- a/src/caches/beliefpropagationcache.jl
+++ b/src/caches/beliefpropagationcache.jl
@@ -45,7 +45,8 @@ function message_diff(message_a::Vector{ITensor}, message_b::Vector{ITensor})
   rhs *= c
   lhs /= norm(lhs)
   rhs /= norm(rhs)
-  return 1 - sqrt(abs(dot(lhs, rhs)))
+  f = abs(dot(lhs, rhs))^2
+  return 1 - f
 end
 
 struct BeliefPropagationCache{PTN,MTS,DM}
--- a/src/caches/beliefpropagationcache.jl
+++ b/src/caches/beliefpropagationcache.jl
@@ -15,7 +15,6 @@ using SimpleTraits: SimpleTraits, Not, @traitfn
 
 default_message(inds_e) = ITensor[denseblocks(delta(i)) for i in inds_e]
 default_messages(ptn::PartitionedGraph) = Dictionary()
-euclidean_dist(x, y) = sqrt(abs(norm(x)^2 + norm(y)^2 - 2 * real(dot(x, y))))
 function default_message_update(contract_list::Vector{ITensor}; kwargs...)
   sequence = optimal_contraction_sequence(contract_list)
   updated_messages = contract(contract_list; sequence, kwargs...)
@@ -44,7 +43,9 @@ function message_diff(message_a::Vector{ITensor}, message_b::Vector{ITensor})
   c = combiner(inds(lhs))
   lhs *= c
   rhs *= c
-  return euclidean_dist(lhs / norm(lhs), rhs / norm(rhs))
+  lhs /= norm(lhs)
+  rhs /= norm(rhs)
+  return 1 - sqrt(abs(dot(lhs, rhs)))
 end
 
 struct BeliefPropagationCache{PTN,MTS,DM}
--- a/src/caches/beliefpropagationcache.jl
+++ b/src/caches/beliefpropagationcache.jl
@@ -1,6 +1,6 @@
 using Graphs: IsDirected
 using SplitApplyCombine: group
-using LinearAlgebra: diag
+using LinearAlgebra: diag, dot
 using ITensors: dir
 using ITensorMPS: ITensorMPS
 using NamedGraphs.PartitionedGraphs:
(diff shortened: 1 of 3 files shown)


src/caches/beliefpropagationcache.jl
```diff
@@ -38,12 +37,16 @@ function default_cache_construction_kwargs(alg::Algorithm"bp", ψ::AbstractITens
...
-  message_a::Vector{ITensor}, message_b::Vector{ITensor}; message_norm=default_message_norm
-)
+function message_diff(message_a::Vector{ITensor}, message_b::Vector{ITensor})
   lhs, rhs = contract(message_a), contract(message_b)
-  norm_lhs, norm_rhs = message_norm(lhs), message_norm(rhs)
-  return 0.5 * norm((denseblocks(lhs) / norm_lhs) - (denseblocks(rhs) / norm_rhs))
+  @assert issetequal(inds(lhs), inds(rhs))
+  c = combiner(inds(lhs))
+  lhs *= c
+  rhs *= c
```
mtfishman, line 45: The combiner seems unnecessary, `norm` and `dot` should be defined for ITensors already. Also checking the indices seems unnecessary, since `dot(a::ITensor, b::ITensor)` is defined as `(dag(a) * b)[]` so will check the result is a scalar, which can only happy if `a` and `b` share indices.

JoeyT1994: Oh that's neat, I will remove those lines

src/caches/beliefpropagationcache.jl
```diff
@@ -33,17 +32,15 @@ default_partitioned_vertices(ψ::AbstractITensorNetwork) = group(v -> v, vertice
...
 end
 
-function message_diff(
-  message_a::Vector{ITensor}, message_b::Vector{ITensor}; message_norm=default_message_norm
-)
+function message_diff(message_a::Vector{ITensor}, message_b::Vector{ITensor})
   lhs, rhs = contract(message_a), contract(message_b)
-  norm_lhs, norm_rhs = message_norm(lhs), message_norm(rhs)
-  return 0.5 * norm((denseblocks(lhs) / norm_lhs) - (denseblocks(rhs) / norm_rhs))
+  f = abs(dot(lhs / norm(lhs), rhs / norm(rhs)))^2
```
mtfishman, line 42:
```suggestion
  f = abs2(dot(lhs / norm(lhs), rhs / norm(rhs)))
```
A slight optimization.

mtfishman:
Good point about the Euclidean distance, I think the Euclidean distance would be something like `1 - real(dot(lhs / norm(lhs), rhs / norm(rhs)))` which as you say depends on the phase factor.

mtfishman:
For some terminology, this measure is related to the [cosine similarity](https://en.wikipedia.org/wiki/Cosine_similarity), and the cosine distance which is defined in [Distances.jl](https://github.com/JuliaStats/Distances.jl), but only for real numbers so far.

src/caches/beliefpropagationcache.jl
```diff
@@ -33,17 +32,15 @@ default_partitioned_vertices(ψ::AbstractITensorNetwork) = group(v -> v, vertice
...
-default_cache_update_kwargs(cache) = (; maxiter=20, tol=1e-5)
+default_cache_update_kwargs(cache) = (; maxiter=25, tol=1e-8)
 function default_cache_construction_kwargs(alg::Algorithm"bp", ψ::AbstractITensorNetwork)
   return (; partitioned_vertices=default_partitioned_vertices(ψ))
 end
 
-function message_diff(
-  message_a::Vector{ITensor}, message_b::Vector{ITensor}; message_norm=default_message_norm
-)
+function message_diff(message_a::Vector{ITensor}, message_b::Vector{ITensor})
```
mtfishman, line 40: An optimization here would be to implement computing the overlaps and norms as network contractions, but that can be left for future work. Could you add a comment about that as a reminder? The current code will probably be slow for messages that are networks. For example if the messages are themselves MPS, the current code would contract the MPS separately and then overlap the result instead of doing a proper inner product of the MPS.

test/test_belief_propagation.jl
```diff
@@ -40,13 +41,13 @@ using Test: @test, @testset
   s = siteinds("S=1/2", g)
   χ = 2
   rng = StableRNG(1234)
-  ψ = random_tensornetwork(rng, s; link_space=χ)
+  ψ = random_tensornetwork(rng, ComplexF64, s; link_space=χ)
```
mtfishman, line 44: A code pattern I use in other tests is to make the entire testset loop over a few element types, for example:
```julia
@testset "belief_propagation (eltype=$elt)" for elt in (Float32, Float64, Complex{Float32}, Complex{Float64})
```
and then use `elt` in the test to construct tensors/tensor networks with different element types. It's also a good way to check that the code preserves the element type properly, by checking that the outputs have element type `elt`, since it is easy to accidentally write code that doesn't preserve single precision (`Float32`), which can cause problems on GPUs.

src/caches/beliefpropagationcache.jl
JoeyT1994: Sure I am happy to add a comment about dealing with overlaps / norms as network contractions.

test/test_belief_propagation.jl
JoeyT1994: Good idea, I added `eltype` support for the message tensor construction and then added that to the test

mtfishman:
Thanks! Nice to see the BP code getting more robust.

PR #194, opened 2024-07-11
BP Efficiency Improvement

JoeyT1994:
The use of `get` in the function `message(bp_cache::BeliefPropagationCache, edge::PartitionEdge)` was causing roughly a factor of 2 slowdown in the BP code and the allocation of ~30% more memory than necessary.

The code has been changed to avoid this (I am not sure why `get` is so inefficient here).

Additionally the `factor(bp_cache::BeliefPropagationCache)` function has been changed to be more efficient and general.


src/caches/beliefpropagationcache.jl
```diff
@@ -106,7 +106,8 @@ end
 
 function message(bp_cache::BeliefPropagationCache, edge::PartitionEdge)
   mts = messages(bp_cache)
-  return get(mts, edge, default_message(bp_cache, edge))
+  haskey(mts, edge) && return mts[edge]
+  return default_message(bp_cache, edge)
```
mtfishman, line 110: Is the issue basically that `default_message(bp_cache, edge)` always gets evaluated, even if the message already exists? There is a version of `get` that accepts a function as the first argument, which makes it so that the default is only evaluated if needed:
```julia
julia> get(Dict("x" => 2), "x", println("test"))
test
2

julia> get(Dict("x" => 2), "x") do
         println("test")
       end
2
```

JoeyT1994: Yeah it seems to be evaluating `default` always which is a lot of unnecessary allocation and causes a factor 2/3 slowdown in some of my tests. Didn't realize that if you made the default the first argument and a function it would fix it. I will do that.

JoeyT1994:
@mtfishman is this able to be merged?

mtfishman:
Yes, thanks for the reminder.

PR #195, opened 2024-08-08
Add normalize flag in default message update

JoeyT1994:
This is quick PR too add support for turning off/ on normalization in the `default_message_update` function for belief propagation.

This avoids having to define a new function when wishing to avoid normalization. Having `normalize = false` is very useful when wishing to contract sub-trees inside a tensor network exactly.


mtfishman:
Looks good, thanks!

PR #201, opened 2024-09-22
Fixes to alternating_update default printing

emstoudenmire:
This PR fixes an issue where the `outputlevel` keyword to the `dmrg` function does not produce any output. (This was noticed when calling it from the TNS solver code, just for some background context.) The changes include:
- setting `region_printer` and `sweep_printer` to use the definitions in defaults.jl for both interfaces to `alternating_update`
- including the `inserter_kwargs` into the `all_kwargs` named tuple in `region_update` so that parameters like `cutoff` will be available to the printers
- making sure `default_region_printer` has defaults (of `nothing`) for the values `cutoff`, `maxdim`, and `mindim` to make that function robust (i.e. not to assume all of these will be passed, since the current pattern was crashing in cases where only, say, `cutoff` was passed into the `inserter_kwargs`)


src/solvers/alternating_update/region_update.jl
```diff
@@ -112,6 +112,7 @@ function region_update(
     outputlevel,
     info...,
     region_kwargs...,
+    inserter_kwargs...,
```
mtfishman, line 115: An alternative design would be to pass `inserter_kwargs` into `all_kwargs` as a NamedTuple instead of splatting them. That would keep those keyword arguments more self contained, i.e. make it clearer which keyword arguments are meant for the inserter. Do you think that would work with the rest of this PR?

emstoudenmire: I do think that change keeps things more organized & it was easy to implement.
I just updated this PR with that change.

mtfishman: I'm confused, I still see `inserter_kwargs` being splatted on this line, did you mean to update that?

src/solvers/defaults.jl
```diff
@@ -23,9 +21,9 @@ function default_region_printer(;
     region = first(sweep_plan[which_region_update])
     @printf("Sweep %d, region=%s \n", which_sweep, region)
     print("  Truncated using")
-    @printf(" cutoff=%.1E", cutoff)
-    @printf(" maxdim=%d", maxdim)
-    @printf(" mindim=%d", mindim)
+    haskey(inserter_kwargs, :cutoff) && @printf(" cutoff=%.1E", inserter_kwargs[:cutoff])
+    haskey(inserter_kwargs, :maxdim) && @printf(" maxdim=%d", inserter_kwargs[:maxdim])
+    haskey(inserter_kwargs, :mindim) && @printf(" mindim=%d", inserter_kwargs[:mindim])
```
mtfishman, line 26:
```suggestion
    haskey(inserter_kwargs, :cutoff) && @printf(" cutoff=%.1E", inserter_kwargs.cutoff)
    haskey(inserter_kwargs, :maxdim) && @printf(" maxdim=%d", inserter_kwargs.maxdim)
    haskey(inserter_kwargs, :mindim) && @printf(" mindim=%d", inserter_kwargs.mindim)
```
Just a style preference, I think it looks cleaner.

emstoudenmire: Agreed – I forgot about that syntax which is much nicer

src/solvers/alternating_update/region_update.jl
emstoudenmire: Yes, I did mean to update it. Actually it was a lucky mistake because I just noticed the code worked anyway. So I did some more digging and `region_kwargs` which is already being splatted into `all_kwargs` already contains `inserter_kwargs`. So the real solution (apart from possibly larger changes) is to just remove that line about `inserter_kwargs` altogether since they are already, subtly, being passed.

emstoudenmire:
I tested the current version and it works well. I would have liked to make a unit test, and while I would know how for a custom printer, here it's the default printer so I couldn't think of a good way except for a kind of test that explicitly checks the stdout output.

mtfishman:
Yeah, testing printing is tricky, since we don't want them to be too specific or else they are constantly breaking, and then we need to update tests any time we change how something prints.

Maybe you could add a test that checks that at least a certain number of lines are printed by default? That shouldn't have to get updated very often.

src/solvers/defaults.jl
```diff
@@ -23,9 +21,9 @@ function default_region_printer(;
     region = first(sweep_plan[which_region_update])
     @printf("Sweep %d, region=%s \n", which_sweep, region)
     print("  Truncated using")
-    @printf(" cutoff=%.1E", cutoff)
-    @printf(" maxdim=%d", maxdim)
-    @printf(" mindim=%d", mindim)
+    haskey(inserter_kwargs, :cutoff) && @printf(" cutoff=%.1E", inserter_kwargs.cutoff)
+    haskey(inserter_kwargs, :maxdim) && @printf(" maxdim=%d", inserter_kwargs.maxdim)
+    haskey(inserter_kwargs, :mindim) && @printf(" mindim=%d", inserter_kwargs.mindim)
```
mtfishman, line 26: This kind of code doesn't feel quite right to me, since it is the same expression repeated multiple times.

Instead, we could turn this into a function and a loop, say:
```julia
print_formatted(x) = print(x)
print_formatted(x::AbstractFloat) = @printf("%.1E", x)
```
and then:
```julia
for key in [:cutoff, :maxdim, :mindim]
  if haskey(inserter_kwargs, key)
    print(key, "=")
    print_formatted(inserter_kwargs[key])
  end
end
```
(I'm not sure if I'm getting the new lines/spacing correct but I think you get the idea.)

emstoudenmire: I don't mind making that change. The only thing about it is that we were making a kind of 'custom' choice for how many digits of the cutoff to print there, just for that one function (the printer, I mean), which is why I was ok with just having it hard coded. But I do agree this is more elegant code and maybe we will find we used `print_formatted` elsewhere like in another printer that we define in that file.

emstoudenmire: (I just mistakenly pushed but didn't make all the changes. Please don't review yet until I do another push & I'll let you know.)

emstoudenmire: Ok now I made all the changes

mtfishman: Yes, that's true that what I suggested hardcodes the number of digits, but we can extend it in the future to allow customization, say by passing the number of digits as a keyword argument and using `Val` dispatch to create definitions like: `ndigits(::Val{:cutoff}) = 1`.

emstoudenmire: I like that idea. See also if you like the tweak I made to it, having it return a string instead of directly printing.

test/test_treetensornetworks/test_solvers/test_dmrg.jl
```diff
@@ -139,7 +165,7 @@ end
     nsweeps,
     maxdim,
     cutoff,
-    outputlevel=2,
+    outputlevel=0,
```
emstoudenmire, line 168: This was previously 2, so was generating a lot of output, and I don't think it was essential to this test so I turned it down to 0.

emstoudenmire:
The tests were working yesterday, except for the documentation, but now are failing.
Interestingly, I can reproduce the testing failure on my laptop, but only when I do `activate .` and `> test` within the REPL.
If I run the tests just by running `runtests.jl` things go better. So it's a bit unclear to me right now what's going on there.

One possible issue, though it may be a red herring, is that on my machine I can't install a testing dependency, "OrdinaryDiffEq" because of a version compatibility issue. (The Pkg error message is quite long and I don't know how to read it, unfortunately.)

src/solvers/defaults.jl
mtfishman: Looks good to me.

mtfishman:
I remember seeing `OrdinaryDiffEq` cause subtle compatibility issues that I wasn't able to understand. I'm trying to run the tests again.

mtfishman:
@emstoudenmire the Windows tests passed, maybe the tests are just timing out or failing for some other non-Julia-specific issue.

emstoudenmire:
I'm glad those passed. So then we can be pretty sure the code is ok. (The relevant tests were passing locally for me too, I was just having issues running the entire testing suite for some reason.)

Ready to merge?

PR #208, opened 2024-10-17
Current ortho fix

JoeyT1994:
This PR fixes an issue with the `current_ortho` function which causes the `alternating_update` interface to break on more generic trees (current tests appeared to only check more standard cases like the comb tree so missed the bug). It also changes a test in `test_dmrg` to use a tree for which the code was previously breaking.

The issue appeared to be in the `overlapping_vertex` clause when at the end of a sweep which was assuming that the last vertex in the sweep was a neighbor of the previous vertex --- which may not always be true for the 'default_region_plan'. Removing the clause fixes the issue as the `else` clause always works anyway.

@mtfishman I feel like the "current_ortho" should either **i)** be a trait carried around by the state (by having, e.g., an ```orthogonal_tensornetwork``` type that wraps an `ITensorNetwork` and a `v_orthocentre`) or **ii)** something that the `alternating update` interface keeps track of as a variable and just re-assigns it each time a la
```
  psi = orthogonalize(psi, v, v_orthocentre)
  v_orthocentre = v
 ```
 where `orthogonalize` can utilize the idea that if `psi` is a tree it just finds the shortest path between `v` and `v_centre` and does a sequence of QRs along that path for efficiency. If `psi` is not a tree it could run BP to put it in the Vidal gauge and then absorb square root `lambdas` everywhere except at `v` where it absorbs the neighboring `lambda` (making the BP environment identity at that site).

### CODE REVIEW

**Comment by JoeyT1994** [comment]
@mtfishman once you merge https://github.com/ITensor/ITensorNetworks.jl/pull/207 I will update the formatting to be in line with `main`. I won't until then just to keep this PR easier to read.

**Comment by mtfishman** [comment]
@JoeyT1994 regarding the formatting, see https://github.com/ITensor/ITensors.jl/pull/1536#issuecomment-2418350000.

**Comment by mtfishman** [comment]
> @mtfishman I feel like the "current_ortho" should either **i)** be a trait carried around by the state (by having, e.g., an `orthogonal_tensornetwork` type that wraps an `ITensorNetwork` and a `v_orthocentre`) or **ii)** something that the `alternating update` interface keeps track of as a variable and just re-assigns it each time a la

The TTN type is designed like **i)**: https://github.com/ITensor/ITensorNetworks.jl/blob/v0.11.21/src/treetensornetworks/treetensornetwork.jl#L11-L12, do you mean you also want that for non-tree networks?

**Comment by JoeyT1994** [comment]
Ah I see okay I will leave the formatting as is then and avoid v2.0 of `JuliaFormatter`

**Comment by JoeyT1994** [comment]
> > @mtfishman I feel like the "current_ortho" should either **i)** be a trait carried around by the state (by having, e.g., an `orthogonal_tensornetwork` type that wraps an `ITensorNetwork` and a `v_orthocentre`) or **ii)** something that the `alternating update` interface keeps track of as a variable and just re-assigns it each time a la
> 
> The TTN type is designed like **i)**: https://github.com/ITensor/ITensorNetworks.jl/blob/v0.11.21/src/treetensornetworks/treetensornetwork.jl#L11-L12, do you mean you also want that for non-tree networks?

I see I didn't know the TTN type had that. I guess I was just thinking more generally, we could have an `OrthogonalITensornetwork <: ITensorNetwork` type instead which carries a `v_centre` around (and if it's a tree it means the environment for that region is exactly identity, and if it's loopy it means the environment on that site/region is, under the BP approximation, identity). That would be in-line with the `VidalITensorNetwork` type we already have and the `ITensorNetwork` type which is just in an arbitrary gauge.

If the `TTN` type carries the centre-site information what is the purpose of the `current_ortho` function here? Can't it just be pulled from the state the `alternating_update` routine is working with?

**Comment by mtfishman** [comment]
> If the `TTN` type carries the centre-site information what is the purpose of the `current_ortho` function here? Can't it just be pulled from the state the `alternating_update` routine is working with instead of the sweep plan?

That's a good question, I don't remember why it was designed that way. It looks like `current_ortho` was introduced by @b-kloss in #143, @b-kloss do you remember the logic behind that design vs. just using the orthogonality center information stored in the `TTN` struct?

**Comment by mtfishman** [comment]
@JoeyT1994 if it seems easier to just use the orthogonality center stored in the TTN, feel free to try to refactor it in that way.

**Comment by b-kloss** [comment]
> > If the `TTN` type carries the centre-site information what is the purpose of the `current_ortho` function here? Can't it just be pulled from the state the `alternating_update` routine is working with instead of the sweep plan?
> 
> That's a good question, I don't remember why it was designed that way. It looks like `current_ortho` was introduced by @b-kloss in #143, @b-kloss do you remember the logic behind that design vs. just using the orthogonality center information stored in the `TTN` struct?

The function  `current_ortho` always returns a single vertex, while the orthogonality center of a TTN is a (contiguous on the graph) set of vertices.  They are not equivalent, and the TTN by itself does not provide enough information to determine what the return of `current_ortho` would be inside `alternating_update`.

For `nsite=1`, i.e. single-site alternating update, `current_ortho` will return the orthogonality center.
For `nsite=2`, i.e. two-site alternating update, `current_ortho` will return the site that overlaps with the previous 2-site region in the sweep plan (which is the one which won't overlap with next 2-site region). For example, the `default_inserter` moves the orthogonality center to the vertex that's not `current_ortho` in the 2-site region. The logic inside `current_ortho` is convoluted because it becomes more complicated than what I described on general trees.

This logic may not generalize to `nsite>2`, and the whole mechanism can certainly be improved.

**Comment by JoeyT1994** [comment]
Thanks for the explanation @b-kloss !

So in the code `current_ortho` always returns a single vertex `ortho_centre` which is not actually the current orthogonality centre but is what you want the orthogonality centre to be? Then that `current_ortho` is used in a call to `orthogonalize(ttn, ortho_centre)` via the `default_extracter`? 

Why does `current_ortho` need to be a single vertex anyway? Could we not just keep track of the `ortho_region` as the region on which `current_ortho` was last called (and thus it can be obtained as `ortho_region(ttn)`) and then during the `extractor` stage of the `region_update` we call
```
ttn = orthogonalize(ttn, region)
``` 
to orthogonalize onto the new region? 

I'll give a refactor a go with this plan in mind as it should be possible to simplify the code along these lines, unless I'm missing something.

**Comment by JoeyT1994** [comment]
@mtfishman @b-kloss I have done a refactor of the code now. The changes done are as follows:

1. I refactored `orthogonalize` so that the functionality is almost entirely written in `abstractitensornetwork.jl` and is wrapped around a base function with takes a vector of edges and performs a sequence of QR decompositions. 
2. The `orthogonalize` functionality in `abstracttreetensornetwork.jl`  is now just a function `orthogonalize(ttn::AbstractTreeTensorNetwork, region::Vector)` which works out the sequence of edges needed to perform `orthogonalize` on to move the orthogonality region of a `ttn` from where it currently is to `region`. Crucially, this is done via `namedgraph_astar` and so _should_ be `O(1)` time complexity assuming `nsites` is finite and the region plan always involves a continuous path through the graph (versus `edge_path` before which I think is `O(L)` time complexity with `L` the size of the network). The path finding logic here could probably be improved further to reduce some overheads but I wrote up a rudimentary one-line version for now.
3. `region_update` now no longer needs to work out any orthogonality information via `ortho_centre`, it can just be pulled from the `ttn` directly when needed and `orthogonalize(ttn, region)` can just be called in the `extracter`. The `ortho_centre` function is now no longer needed and the bug this PR aimed to fix is gone.

I think this simplifies the logic of things quite a bit and should, in principle, actually speed the code up (because it was `O(L^{2})` before due to use of `edge_path` in `orthogonalize`

File: src/abstractitensornetwork.jl [hunk 1]
```diff
@@ -585,36 +585,58 @@ end
...
+  return _orthogonalize_edges(tn, edges; kwargs...)
+end
+
 function ITensorMPS.orthogonalize(tn::AbstractITensorNetwork, edge::Pair; kwargs...)
   return orthogonalize(tn, edgetype(tn)(edge); kwargs...)
 end
 
-# Orthogonalize an ITensorNetwork towards a source vertex, treating
+function ITensorMPS.orthogonalize(
+  tn::AbstractITensorNetwork, edges::Vector{Pair}; kwargs...
```
mtfishman, line 625:
```suggestion
  tn::AbstractITensorNetwork, edges::Vector{<:Pair}; kwargs...
```

src/abstractitensornetwork.jl
```diff
@@ -585,36 +585,58 @@ end
...
+    tn[dst(edge)] *= Y
+  end
   return tn
 end
 
 function ITensorMPS.orthogonalize(tn::AbstractITensorNetwork, edge::AbstractEdge; kwargs...)
   return _orthogonalize_edge(tn, edge; kwargs...)
 end
 
+function ITensorMPS.orthogonalize(
```
mtfishman, line 614: I think it would be better to rename the versions that orthogonalize along an edge or sequence of edges to something else (I know I made the name originally so that was my fault).

There are two distinct concepts, orthogonalizing _along_ a path/trail/walk (https://en.wikipedia.org/wiki/Path_(graph_theory)), and orthogonalizing _towards_ a region. My concern with the current naming scheme is that either one could use vertices or edges, but are really distinct functionality, so it isn't great from a generic coding perspective. What if we want to define the region we orthogonalize towards by a set of edges, or the path we want to orthogonalize along by a set of vertices?

We use `orthogonalize` to mean orthogonalizing towards a region in `ITensorMPS.jl`, and I think that will be the more common operation that external users will use, so I think we should stick with that name. So then I think we should rename this function that defines a path/trail/walk to orthogonalize along as something else. It could be `orthogonalize_walk` based on the terminology here: https://en.wikipedia.org/wiki/Path_(graph_theory). An alternative is `orthogonalize_trail` but I could imagine cases where you want to repeat edges, say on a loopy graph, so maybe it is better to be more general. We could also call it `orthogonalize_along` but I slightly prefer using graph terminology.

src/edge_sequences.jl
```diff
@@ -44,3 +44,8 @@ end
 @traitfn function edge_sequence(::Algorithm"parallel", g::::(!IsDirected))
   return [[e] for e in vcat(edges(g), reverse.(edges(g)))]
 end
+
+function post_order_dfs_edges_region(g::AbstractGraph, region)
+  es = post_order_dfs_edges(g, first(region))
+  return filter(e -> !((src(e) ∈ region) && (dst(e) ∈ region)), es)
+end
```
mtfishman, line 51: So if I understand correctly, this returns the edges directed towards the region, excluding any edges in the region?

mtfishman:
@JoeyT1994 this definitely looks like a nice simplification of the code, though I'm having trouble picturing how the old code and new code are working diagrammatically. Let's meet to discuss just so I make sure I understand what's going on and can be convinced that the new code and old code are doing the same thing.

This could help with something I was trying in the tensor network impurity solver. I was trying to automate finding the tree structure but for tree structures I was generating I was hitting a bug in the solver code, hopefully it was this bug that you're fixing here!

src/edge_sequences.jl
JoeyT1994: Yes exactly, which is exactly the sequence of edges one will need to orthogonalize along to make `region` the `ortho_region` of a `ttn`. 
Maybe we could be a bit tighter with the logic here and demand `region` is a contiguous group of vertices. I suppose it should also probably end up being a `NamedGraphs` function at some point.

src/abstractitensornetwork.jl
JoeyT1994: Yeah that's a good point that we are using the same terminology for two somewhat distinct concepts. What about `orthogonalize_sequence`?

JoeyT1994:
Thanks @mtfishman . I incorporated some of your comments and yes let's meet to discuss how the logic works here.

One slight flag here (which was also in the previous version) is that the `edge_sequence` to move the `ortho_region` is computed via an `O(L)` time complexity function with `L` the number of tensors in the network. 
Hence this actually makes the `alternating_update` sweeping on trees have an `O(L^{2})` time complexity term (the sequence is computed every `region_update`). Although this is likely much smaller than the `O(L Poly(chi))` scaling it is not really optimal and we should be able to write a more sophisticated function to move the `ortho_region` that is `O(1)` if the current `ortho_region` and new `ortho_region` are not extensive and are not separated by an extensive path.

src/abstractitensornetwork.jl
```diff
@@ -584,37 +584,63 @@ function LinearAlgebra.factorize(tn::AbstractITensorNetwork, edge::Pair; kwargs.
...
 # the network as a tree spanned by a spanning tree.
 # TODO: Rename `tree_orthogonalize`.
-function ITensorMPS.orthogonalize(ψ::AbstractITensorNetwork, source_vertex)
-  spanning_tree_edges = post_order_dfs_edges(bfs_tree(ψ, source_vertex), source_vertex)
-  for e in spanning_tree_edges
-    ψ = orthogonalize(ψ, e)
-  end
-  return ψ
+function ITensorMPS.orthogonalize(ψ::AbstractITensorNetwork, region::Vector)
+  spanning_tree_edges = post_order_dfs_edges_region(bfs_tree(ψ, first(region)), region)
```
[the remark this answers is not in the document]
JoeyT1994: Also rename to `tree_orthogonalize`

src/abstractitensornetwork.jl
```diff
@@ -584,37 +584,63 @@ function LinearAlgebra.factorize(tn::AbstractITensorNetwork, edge::Pair; kwargs.
...
+  return orthogonalize_path(ψ, spanning_tree_edges)
+end
+
+function ITensorMPS.orthogonalize(ψ::AbstractITensorNetwork, region)
+  return orthogonalize(ψ, [region])
+end
+
+function ITensorMPS.orthogonalize(ψ::AbstractITensorNetwork, edges::Vector{<:AbstractEdge})
```
JoeyT1994, line 630: Remove `edges` functionality for now as it is not entirely clear whether we are orthogonalizing onto the bond or the region defined by the vertices in `edges`

src/edge_sequences.jl
JoeyT1994: Change the logic to fill in the region if it is not contiguous

src/solvers/extract/extract.jl
```diff
@@ -7,18 +7,19 @@
 # insert_local_tensors takes that tensor and factorizes it back
 # apart and puts it back into the network.
 #
-function default_extracter(state, projected_operator, region, ortho; internal_kwargs)
-  state = orthogonalize(state, ortho)
+function default_extracter(state, projected_operator, region; internal_kwargs)
   if isa(region, AbstractEdge)
```
JoeyT1994, line 11: Add a Todo here about `orthogonalize` supporting bond orthogonalization

src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -29,30 +35,14 @@ function set_ortho_region(tn::AbstractTTN, new_region)
...
+    ttn = typeof(ttn)(orthogonalize_path(ITensorNetwork(ttn), path; kwargs...))
   end
-  return set_ortho_region(tn, typeof(ortho_region(tn))([ortho_center]))
-end
-
-# For ambiguity error
-
-function ITensorMPS.orthogonalize(tn::AbstractTTN, edge::AbstractEdge; kwargs...)
-  return typeof(tn)(orthogonalize(ITensorNetwork(tn), edge; kwargs...))
+  return set_ortho_region(ttn, region)
```
JoeyT1994, line 45: `region` here should always end up being contiguous.

src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -29,30 +35,14 @@ function set_ortho_region(tn::AbstractTTN, new_region)
...
-  if isone(length(ortho_region(tn)))
-    edge_list = edge_path(tn, only(ortho_region(tn)), ortho_center)
-  else
-    edge_list = post_order_dfs_edges(tn, ortho_center)
-  end
-  for e in edge_list
-    tn = orthogonalize(tn, e)
+function ITensorMPS.orthogonalize(ttn::AbstractTTN, region::Vector; kwargs...)
+  new_path = post_order_dfs_edges_region(ttn, region)
+  existing_path = post_order_dfs_edges_region(ttn, ortho_region(ttn))
```
JoeyT1994, line 40: Change this logic. Create a contiguous super region containing `ortho_region` and at least one vertex `v` in `region`. Do `post_order_dfs_edges(ttn[super_region], v)`.

src/abstractitensornetwork.jl
```diff
@@ -584,37 +585,49 @@ function LinearAlgebra.factorize(tn::AbstractITensorNetwork, edge::Pair; kwargs.
...
+function orthogonalize_path(tn::AbstractITensorNetwork, edges::Vector{<:Pair}; kwargs...)
+  return orthogonalize_path(tn, edgetype(tn).(edges); kwargs...)
 end
 
-function ITensorMPS.orthogonalize(tn::AbstractITensorNetwork, edge::Pair; kwargs...)
-  return orthogonalize(tn, edgetype(tn)(edge); kwargs...)
+# Orthogonalize an ITensorNetwork towards a region, treating
+# the network as a tree spanned by a spanning tree.
+function tree_orthogonalize(ψ::AbstractITensorNetwork, region::Vector)
+  region_centre =
```
mtfishman, line 622:
```suggestion
  region_center =
```
(sorry, we're using American English here!)

src/abstractitensornetwork.jl
```diff
@@ -584,37 +585,49 @@ function LinearAlgebra.factorize(tn::AbstractITensorNetwork, edge::Pair; kwargs.
...
-  return _orthogonalize_edge(tn, edge; kwargs...)
+function orthogonalize_path(tn::AbstractITensorNetwork, edges::Vector{<:Pair}; kwargs...)
+  return orthogonalize_path(tn, edgetype(tn).(edges); kwargs...)
 end
 
-function ITensorMPS.orthogonalize(tn::AbstractITensorNetwork, edge::Pair; kwargs...)
-  return orthogonalize(tn, edgetype(tn)(edge); kwargs...)
+# Orthogonalize an ITensorNetwork towards a region, treating
+# the network as a tree spanned by a spanning tree.
+function tree_orthogonalize(ψ::AbstractITensorNetwork, region::Vector)
```
mtfishman, line 621: I'm confused by the current code, does this only orthogonalize `region`, or does it orthogonalize everything outside of `region`?

mtfishman: Oh sorry, I missed the line:
```julia
path = filter(e -> !((src(e) ∈ region) && (dst(e) ∈ region)), path)
```
so I see that it orthogonalizes the complement of the (filled in) region.

src/abstractitensornetwork.jl
```diff
@@ -584,37 +585,49 @@ function LinearAlgebra.factorize(tn::AbstractITensorNetwork, edge::Pair; kwargs.
...
+function orthogonalize_path(tn::AbstractITensorNetwork, edge::AbstractEdge; kwargs...)
+  return orthogonalize_path(tn, [edge]; kwargs...)
+end
+
+function orthogonalize_path(tn::AbstractITensorNetwork, edge::Pair; kwargs...)
+  return orthogonalize_path(tn, edgetype(tn)(edge); kwargs...)
+end
+
+# For ambiguity error; TODO: decide whether to use graph mutating methods when resulting graph is unchanged?
+function orthogonalize_path(
```
mtfishman, line 597: I would prefer calling this `orthogonalize_walk` since the set of edges may in general define a walk rather than a path (https://en.wikipedia.org/wiki/Path_(graph_theory)#Walk,_trail,_and_path).

src/solvers/sweep_plans/sweep_plans.jl
```diff
@@ -71,12 +73,13 @@ function forward_sweep(
...
   if reverse_step
     reverse_regions = collect(
       flatten(
         map(
-          i -> reverse_region(edges, i; region_kwargs=reverse_kwargs, kwargs...),
+          i -> reverse_region(
+            edges, i; reverse_edge=reverse_edges, region_kwargs=reverse_kwargs, kwargs...
+          ),
           eachindex(edges),
         ),
```
mtfishman, line 84: I would prefer using do-block syntax here (https://docs.julialang.org/en/v1/manual/functions/#Do-Block-Syntax-for-Function-Arguments-1) since the function being passed to `map` is pretty long.

src/solvers/sweep_plans/sweep_plans.jl
```diff
@@ -71,12 +73,13 @@ function forward_sweep(
...
     flatten(map(i -> forward_region(edges, i; region_kwargs, kwargs...), eachindex(edges)))
   )
-
   if reverse_step
     reverse_regions = collect(
       flatten(
         map(
-          i -> reverse_region(edges, i; region_kwargs=reverse_kwargs, kwargs...),
+          i -> reverse_region(
+            edges, i; reverse_edge=reverse_edges, region_kwargs=reverse_kwargs, kwargs...
```
mtfishman, line 81:
```suggestion
            edges, i; reverse_edge=reverse_edges, region_kwargs, kwargs...
```

JoeyT1994: How come you wan't to not pass `reverse_kwargs` here? They seem like they are being passed (although they are made identical to `region_kwargs` by the look of the code in `tdvp_sweep_plan`)

mtfishman: Oh, I see, I misread the code and thought it was `region_kwargs=region_kwargs` as opposed to `region_kwargs=reverse_kwargs`, never mind.

mtfishman:
Looks good, thanks!

PR #211, opened 2024-11-26
Gauge walks

JoeyT1994:
This PR generalizes the `tree_orthogonalize` code to `tree_gauge` code which is based around a generic algorithm which `walks` through a series of edges of the `ITensorNetwork` and performs some factorization which can be specified with an `alg::Algorithm` argument. 

This means that we can support other `gauges` such as `interpolative` with very minimal additional code (just one function which tells you how to do the edge factorizations).


src/abstractitensornetwork.jl
```diff
@@ -585,17 +586,22 @@ function LinearAlgebra.factorize(tn::AbstractITensorNetwork, edge::Pair; kwargs.
...
-function orthogonalize_walk(tn::AbstractITensorNetwork, edge::Pair; kwargs...)
-  return orthogonalize_walk(tn, edgetype(tn)(edge); kwargs...)
+function gauge_walk(alg::Algorithm, tn::AbstractITensorNetwork, edge::Pair; kwargs...)
+  return gauge_walk(alg::Algorithm, tn, edgetype(tn)(edge); kwargs...)
 end
 
 # For ambiguity error; TODO: decide whether to use graph mutating methods when resulting graph is unchanged?
-function orthogonalize_walk(
-  tn::AbstractITensorNetwork, edges::Vector{<:AbstractEdge}; kwargs...
+function gauge_walk(
```
mtfishman, line 600: Can we make this function generic to any gauge, and have it loop over a lower level function `gauge_edge(alg::Algorithm, tn::AbstractITensorNetwork, edge::AbstractEdge)` that is then specialized to the gauge? It seems like most gauges will reduce to looping over edges.

JoeyT1994: Makes sense, I changed it to that!

emstoudenmire:
Looks like a nice improvement. @JoeyT1994 will you be updating the ITensorTCI.jl PR to use a similar gauge_walk / gauge_edge / Algorithm design?

JoeyT1994:
Yes, the idea will be that I can then simplify that PR by removing most functions from the `interpolative_gauge.jl` file other than the function that tells you how to do the `interpolative_decomposition` on an edge of a tensor network.

src/abstractitensornetwork.jl
```diff
@@ -44,7 +45,7 @@ using MacroTools: @capture
 using NamedGraphs: NamedGraphs, NamedGraph, not_implemented, steiner_tree
 using NamedGraphs.GraphsExtensions:
   ⊔, directed_graph, incident_edges, rename_vertices, vertextype
-using NDTensors: NDTensors, dim
+using NDTensors: NDTensors, dim, Algorithm
```
mtfishman, line 48: Let's load `Algorithm` from `ITensor` instead, since that is where you are loading `@Algorithm_str`.

mtfishman:
Looks great, thanks! This will be very nice to have.

PR #212, opened 2025-01-02
Generic Boundary Matrix Product States Contraction Code

JoeyT1994:
This PR adds code for doing boundary matrix product state contraction techniques in ITensorNetworks using a 1-site MPS fitting procedure. The result is the ability to do efficient, controlled approximate contractions of 2D and quasi-2D tensor networks.

The `BeliefPropagationCache` has been separated into an `AbstractBeliefPropagationCache` and the concrete subtypes `BeliefPropagationCache` and `BoundaryMPSCache` in order to achieve this in a streamlined way.

 The `bmpsc::BoundaryMPSCache` is an `AbstractBeliefPropagationCache`  which wraps a `bpc::BeliefPropagationCache`, an added layer of partitioning around the `partitioned_tensornetwork` in `BeliefPropagationCache` and the `maximum_virtual_dimension` which indicates the maximum permissible size of the virtual links between message tensors.

The underlying `partitioned_tensornetwork` in the `BeliefPropagationCache` is assumed to form a grid with only horizontal and vertical edges. Typically vertices are expected to be labeled as integer tuples `(column, row)` so that edges are of the form `(column, row) => (column +- 1, row)` or `(row, column) => (column, row +-1)`. Either the rows or the columns can then be partitioned up using the `grouping_function` kwarg for the constructor (default is the columns). As an 'expert mode', Arbitrary vertex names are actually supported provided the user tells the code how to partition them corresponding planar graph up into a line / ring and then how to sort the vertices within each partition so they go in ascending / descending order.

The message tensors in the `BeliefProgpationCache` which live between the row/column partitions carry an additional virtual index and thus connect up to form an implicit `MPS` between partitions.

Functionality for moving the biorthogonal and orthogonal gauge within this implicit `MPS` is supported and the 1-site boundary MPS routine follows naturally by appropriate ordering of gauge moves and calls to `message_update(bp_cache(bmpsc, pe::PartitionEdge))`.

Tests are added which show some of the generality of this code (works immediately for complex numbers and real numbers,  norm networks and flat tensor networks, on a range of lattices such as hexagonal lattices, square lattices, comb trees etc). The code also immediately works for half-periodic systems where the doubly partitioned network forms a ring and effectively makes the BP approximation on the ring (along with the approximation from the rank of the MPS).

Note this code can be  leveraged to do powerful things things like approximately contract "quasi-3D" tensor networks simply by passing a belief propagation cache which partitions up such a network into a flat 2D structure.

Currently the `orthogonal` gauge is used as default for the MPS fitting as it is stable and robust. The `biorthogonal` gauge is working but more sensitive to things like the initial message tensors used, the details of the SVD and the ordering of `partitionedge` updates. Work will need to be done to figure out how to stabilize it.

A number of further improvements are possible. 

1. Support for `n-site` MPS fitting which might help the biorthogonal stability.
2. Support for keeping track of the biorthogonal gauge centre and utilizing it to form `channels` and efficiently compute more complicated derivatives of the tensor network (currently `environment(bmpsc::BoundaryMPSCache, verts::Vector)`) is only supported for contiguous vertices which lie in the same column/ row partition.

@mtfishman @emstoudenmire


src/caches/boundarympscache.jl
```diff
@@ -0,0 +1,506 @@
...
+
+function region_scalar(bmpsc::BoundaryMPSCache, partition)
+  partition_vs = planargraph_vertices(bmpsc, partition)
+  bmpsc = partition_update(bmpsc, first(partition_vs), last(partition_vs))
+  return region_scalar(bp_cache(bmpsc), PartitionVertex(last(partition_vs)))
+end
+
+function region_scalar(bmpsc::BoundaryMPSCache, partitionpair::Pair)
+  pes = planargraph_partitionpair_partitionedges(bmpsc, partitionpair)
+  out = ITensor(1.0)
```
mtfishman, line 501:
```suggestion
  out = ITensor(one(Bool))
```
to be more agnostic about the element type, i.e. `1.0` is implicitly `Float64` which would promote everything to double precision, which isn't good to do on GPUs.

test/test_boundarymps.jl
```diff
@@ -0,0 +1,154 @@
...
+    #Do Orthogonal Boundary MPS
+    message_update_kwargs = (; niters=25, tolerance=1e-10)
+    ψIψ_boundaryMPS = BoundaryMPSCache(QuadraticFormNetwork(ψ_symm); message_rank=1)
+    ψIψ_boundaryMPS = update(ψIψ_boundaryMPS; message_update_kwargs)
+
+    for pe in keys(messages(bp_cache))
+      m_boundarymps = only(message(ψIψ_boundaryMPS, pe))
+      #Prune the dimension 1 virtual index from boundary MPS message tensor
+      m_boundarymps =
+        m_boundarymps * ITensor(1.0, filter(i -> dim(i) == 1, inds(m_boundarymps)))
```
mtfishman, line 131:
```suggestion
        m_boundarymps * ITensor(one(Bool), filter(i -> dim(i) == 1, inds(m_boundarymps)))
```

test/test_boundarymps.jl
```diff
@@ -0,0 +1,154 @@
...
+
+    #Do Biorthogonal Boundary MPS
+    ψIψ_boundaryMPS = BoundaryMPSCache(QuadraticFormNetwork(ψ_symm); message_rank=1)
+    ψIψ_boundaryMPS = update(ψIψ_boundaryMPS; alg="biorthogonal")
+
+    for pe in keys(messages(bp_cache))
+      m_boundarymps = only(message(ψIψ_boundaryMPS, pe))
+      #Prune the dimension 1 virtual index from boundary MPS message tensor
+      m_boundarymps =
+        m_boundarymps * ITensor(1.0, filter(i -> dim(i) == 1, inds(m_boundarymps)))
```
mtfishman, line 146:
```suggestion
        m_boundarymps * ITensor(one(Bool), filter(i -> dim(i) == 1, inds(m_boundarymps)))
```

PR #216, opened 2025-02-13
Pass eager flag to KrylovKit.exponentiate

emstoudenmire:
This PR fixes a typo in the exponentiate solver where the `eager` flag was intended to be passed, but was not being. In practice I have found that `eager=true` is indeed a good default and makes a big difference to performance.


mtfishman:
Great, thanks! Can you bump the package version?

emstoudenmire:
Ok, just bumped it in Project.toml. Please let me know if I need to do anything else e.g. with registering the new version.

mtfishman:
> Ok, just bumped it in Project.toml. Please let me know if I need to do anything else e.g. with registering the new version.

Nothing else to do from your end, I'll merge and then register.

PR #217, opened 2025-03-11
AbstractBeliefPropagationCache

JoeyT1994:
This PR separates the `BeliefPropagationCache` structure into an abstract and a concrete type in order to support future, more general, bp caches such as the `BoundaryMPSCache` which can be viewed as a `BeliefPropagationCache` with the message tensors endowed with extra virtual indices and more complex update rules.

Some functions have been renamed in order to reflect their usage better and the `inner`, `environment` and `scalar` functionality has been cleaned up so as to immediately support more general `abstractbeliefpropagationcache` based Algorithms without the need to hard-code them, provided certain functionality is supported on the underlying cache. 

@mtfishman this is the PR: https://github.com/ITensor/ITensorNetworks.jl/pull/212 but without the `BoundaryMPSCache` parts implemented (which we discussed some changes to and I will work on). This should make for an easier review. Essentially the changes here are a clean-up of the BP code in order to make the way for this future PR (and other caches) where one wants to use `BeliefPropagationCache` structures with different `message_update` functions etc.


mtfishman:
Thanks @JoeyT1994, I'll take a look. I haven't looked at this code in a while so probably it would be helpful to meet to discuss it.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -0,0 +1,292 @@
...
+  incoming_ms = incoming_messages(bpc, vertex; ignore_edges=PartitionEdge[reverse(edge)])
+  state = factors(bpc, vertex)
+
+  return message_update_function(
+    ITensor[incoming_ms; state]; message_update_function_kwargs...
+  )
+end
+
+function update(
+  alg::Algorithm"simplebp",
```
mtfishman, line 207: What's the difference between `"simplebp"` and `"bp"`?

JoeyT1994: Nothing actually, can just call it `"bp"` as that's what it is. Then for a `boundaryMPSCache` the default update would be called `"boundarymps"` or something similar

mtfishman: Sounds good. I would wonder if we could make `"boundarymps"` into just a special "mode" of `"bp"`, i.e. only have the `"bp"` algorithm. Maybe how it could work is that simple BP and boundary MPS both use `"bp"` as the algorithm, but they can be specialized on the cache type (i.e. `BeliefPropagationCache` vs. `BoundaryMPSCache`). Conceptually that makes sense to me but I'm not sure how that would work in practice (or if it is worth it in terms of simplifying the code).

mtfishman: Anyway, that can be left for the next PR, it is independent of this PR.

mtfishman:
@JoeyT1994 this looks good to me, can you bump the package version so I can register it after it is merged?

PR #218, opened 2025-03-20
Contraction sequence finding

JoeyT1994:
This PR makes the default contraction sequence backend for the `tn::AbstractITensorNetwork` type the `optimaltree` function from `TensorOperations.jl`.

This is achieved by extracting the tensors from the `tn` and passing them to the overwritten function `ITensors.optimal_contraction_sequence(tensors::Vector{<:ITensor})` which strips the tensors down to their indices and passes to `optimaltree` which happily recognizes the `ITensor` `Index` type. 

This sequence finder is noticeably faster (well over an order of magnitude) than the current `optimal_contraction_sequence` function in `ITensors.jl` for a range of different networks tested - especially when a large number of tensors is involved.

`TensorOperations` is therefore now a dependency for `ITensorNetworks`.

A similar PR will be made to `ITensors.jl` in order to change the default there.


mtfishman:
Thanks for taking a look at this @JoeyT1994. Some suggestions:
1. I would prefer to turn this into an actual package extension. That would mean requiring loading TensorOperations.jl to use contraction sequence optimization, but I think that is ok. Even though it is slightly less convenient, it is also more explicit for users, potentially improves load times if someone doesn't need contraction sequence optimization, etc.
2. Let's try to not use `ContractionSequenceOptimization` at all, we should probably just get rid of that module entirely if we are moving to using `TensorOperations.optimaltree` here and in ITensors.jl. I think as an alternative we can overload `ITensorNetworks.contraction_sequence(::Algorithm"optimal", tn::Vector{ITensor})` to convert and call out to `TensorOperations.optimaltree` in the package extension.

1. is technically breaking so we should bump to v0.12.

Project.toml
```diff
@@ -32,6 +32,7 @@ SplitApplyCombine = "03a91e81-4c3e-53e1-a0a4-9c0c8f19dd66"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StructWalk = "31cdf514-beb7-4750-89db-dda9d2eb8d3d"
 Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
+TensorOperations = "6aa20fa7-93e2-5fca-9bc0-fbd0db3c71a2"
```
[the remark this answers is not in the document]
mtfishman: Should be moved to `weakdeps`.

JoeyT1994:
> Thanks for taking a look at this @JoeyT1994. Some suggestions:
> 
> 1. I would prefer to turn this into an actual package extension. That would mean requiring loading TensorOperations.jl to use optimal contraction sequence optimization, but I think that is ok. Even though it is slightly less convenient, it is also more explicit for users, potentially improves load times if someone doesn't need contraction sequence optimization, etc.
> 2. Let's try to not use `ContractionSequenceOptimization` at all (including the function `optimal_contraction_sequence`), we should probably just get rid of that module entirely if we are moving to using `TensorOperations.optimaltree` here and in ITensors.jl. I think as an alternative we can overload `ITensorNetworks.contraction_sequence(::Algorithm"optimal", tn::Vector{ITensor})` to convert and call out to `TensorOperations.optimaltree` in the package extension. That would also mean defining `deepmap` in ITensorNetworks.jl, which I think is reasonable anyway.
> 
> `1.` is technically breaking so we should bump to v0.12.

Thanks Matt. I followed your suggestions and made `TensorOperations` a weak dependency. Presumably this means that if `contraction_sequence()` is called and `TensorOperations` is not loaded the code will error? I think optimal sequence finding is used in a lot of places in the code so is there a way to make it error so that it tells the user they should load `TensorOperations` ?

ext/ITensorNetworksTensorOperationsExt/ITensorNetworksTensorOperationsExt.jl
```diff
@@ -0,0 +1,19 @@
...
+
+function ITensorNetworks.contraction_sequence(::Algorithm"optimal", tn::Vector{ITensor})
+  return optimal_contraction_sequence(tn)
+end
+
+function optimal_contraction_sequence(tensors::Vector{<:ITensor})
+  network = collect.(inds.(tensors))
+  inds_to_dims = Dict(i => dim(i) for i in unique(reduce(vcat, network)))
+  seq, _ = optimaltree(network, inds_to_dims)
+  return seq
```
mtfishman, line 16:
```suggestion
  network = collect.(inds.(tn))
  inds_to_dims = Dict(i => dim(i) for i in unique(reduce(vcat, network)))
  seq, _ = optimaltree(network, inds_to_dims)
  return seq
```
Just a bit simpler.

test/test_contraction_sequence.jl
```diff
@@ -5,6 +5,7 @@ using ITensorNetworks:
 using ITensors: ITensors, contract
 using NamedGraphs.NamedGraphGenerators: named_grid
 using OMEinsumContractionOrders: OMEinsumContractionOrders
+using TensorOperations
```
mtfishman, line 8:
```suggestion
using TensorOperations: TensorOperations
```
To minimize what we bring into the namespace.

test/test_expect.jl
```diff
@@ -12,6 +12,7 @@ using ITensorNetworks:
 using SplitApplyCombine: group
 using StableRNGs: StableRNG
 using Test: @test, @testset
+using TensorOperations
```
mtfishman, line 15:
```suggestion
using TensorOperations: TensorOperations
```
To minimize what we bring into the namespace.

mtfishman:
> Thanks Matt. I followed your suggestions and made `TensorOperations` a weak dependency. Presumably this means that if `contraction_sequence()` is called and `TensorOperations` is not loaded the code will error? I think optimal sequence finding is used in a lot of places in the code so is there a way to make it error so that it tells the user they should load `TensorOperations` ?

Yes, that's a downside of this. We could make a backup definition:
```julia
function ITensorNetworks.contraction_sequence(alg::Algorithm, tn::Vector{ITensor})
  return throw(ArgumentError(#= Error message saying `alg` isn't defined, suggest loading a backend like TensorOperations.jl or OMEinsumContractionOrders.jl =#))
end
```

ext/ITensorNetworksTensorOperationsExt/ITensorNetworksTensorOperationsExt.jl
```diff
@@ -0,0 +1,15 @@
+module ITensorNetworksTensorOperationsExt
+
+using ITensors: ITensors, ITensor, dim, inds
+using ITensorNetworks: ITensorNetworks
+using TensorOperations: optimaltree, TensorOperations
+using NDTensors.AlgorithmSelection: @Algorithm_str
```
mtfishman, line 6:
```suggestion
using NDTensors.AlgorithmSelection: @Algorithm_str
using TensorOperations: optimaltree, TensorOperations
```
Alphabetical order.

test/test_additensornetworks.jl
```diff
@@ -7,6 +7,7 @@ using ITensors: ITensors, apply, op, scalar, inner
 using LinearAlgebra: norm_sqr
 using StableRNGs: StableRNG
 using Test: @test, @testset
+using TensorOperations: TensorOperations
```
mtfishman, line 10:
```suggestion
using TensorOperations: TensorOperations
using Test: @test, @testset
```
Alphabetical order.

mtfishman:
@JoeyT1994 looks like you need to add TensorOperations as a test dependency by adding it to `test/Project.toml`.

JoeyT1994:
@mtfishman ! Thanks, I forgot about that and wondered what has happening

mtfishman:
Looks good, thanks! One step closer to getting rid of more code from ITensors.jl, which is always a good thing.

PR #221, opened 2025-03-27
Make compatible with ITensors 0.9

JoeyT1994:
The latest version of ITensors (v0.9) contained some breaking changes due to the use of `TensorOperations` in `optimal_contraction_sequence`. This PR updates `ITensorNetworks.jl` to be compatible with those changes, with belief propagation and apply directly making use of `optimal_sequence` finding.


src/apply.jl
```diff
@@ -447,37 +446,13 @@ function optimise_p_q(
...
-
-  function b(
-    p::ITensor,
-    q::ITensor,
-    o::ITensor,
-    envs::Vector{ITensor},
-    r::ITensor;
-    opt_sequence=nothing,
-  )
+  function b(p::ITensor, q::ITensor, o::ITensor, envs::Vector{ITensor}, r::ITensor;)
```
mtfishman, line 449:
```suggestion
  function b(p::ITensor, q::ITensor, o::ITensor, envs::Vector{ITensor}, r::ITensor)
```

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -17,8 +17,7 @@ using NDTensors: NDTensors
 abstract type AbstractBeliefPropagationCache end
 
 function default_message_update(contract_list::Vector{ITensor}; normalize=true, kwargs...)
-  sequence = optimal_contraction_sequence(contract_list)
-  updated_messages = contract(contract_list; sequence, kwargs...)
+  updated_messages = contract(contract_list; sequence="automatic", kwargs...)
```
JoeyT1994, line 20: Revert this. So that it never passes to `ITensors.jl`. Use `contraction_sequence(contract_list; alg = "optimal")`.

src/apply.jl
mtfishman: Looks like the extra `;` is still there.

src/apply.jl
```diff
@@ -447,61 +449,32 @@ function optimise_p_q(
...
-          ITensor[
-            p_q_tensor,
-            replaceinds(prime(dag(p_q_tensor)), prime(s_ind), s_ind),
-            apply_tensor,
-          ],
-          envs,
-        );
-        sequence=opt_sequence,
-      ),
+  function M_p(envs::Vector{ITensor}, p_q_tensor::ITensor, s_ind, apply_tensor::ITensor;)
```
mtfishman, line 458:
```suggestion
  function M_p(envs::Vector{ITensor}, p_q_tensor::ITensor, s_ind, apply_tensor::ITensor)
```

src/caches/beliefpropagationcache.jl
```diff
@@ -94,22 +94,16 @@ function environment(bpc::BeliefPropagationCache, verts::Vector; kwargs...)
   return vcat(messages, central_tensors)
 end
 
-function region_scalar(
-  bp_cache::BeliefPropagationCache,
-  pv::PartitionVertex;
-  contract_kwargs=(; sequence="automatic"),
-)
+function region_scalar(bp_cache::BeliefPropagationCache, pv::PartitionVertex;)
```
mtfishman, line 97:
```suggestion
function region_scalar(bp_cache::BeliefPropagationCache, pv::PartitionVertex)
```

src/caches/beliefpropagationcache.jl
```diff
@@ -94,22 +94,16 @@ function environment(bpc::BeliefPropagationCache, verts::Vector; kwargs...)
...
 
-function region_scalar(
-  bp_cache::BeliefPropagationCache,
-  pe::PartitionEdge;
-  contract_kwargs=(; sequence="automatic"),
-)
-  return contract(
-    vcat(message(bp_cache, pe), message(bp_cache, reverse(pe))); contract_kwargs...
-  )[]
+function region_scalar(bp_cache::BeliefPropagationCache, pe::PartitionEdge;)
```
mtfishman, line 105:
```suggestion
function region_scalar(bp_cache::BeliefPropagationCache, pe::PartitionEdge)
```

ext/ITensorNetworksTensorOperationsExt/ITensorNetworksTensorOperationsExt.jl
```diff
@@ -7,7 +7,7 @@ using TensorOperations: TensorOperations, optimaltree
 
 function ITensorNetworks.contraction_sequence(::Algorithm"optimal", tn::ITensorList)
   network = collect.(inds.(tn))
-  inds_to_dims = Dict(i => dim(i) for i in unique(reduce(vcat, network)))
+  inds_to_dims = Dict(i => Float64(dim(i)) for i in unique(reduce(vcat, network)))
```
mtfishman, line 10: What's this conversion for?

mtfishman: I would have thought this might be bad since the contraction sequence optimizer has to repeatedly compute the contractions cost of pairwise contractions and working with integers should be faster.

mtfishman:
@JoeyT1994 it looks like you'll need to resolve conflicts caused by the removal of ITensorMPS.jl as a dependency.

ext/ITensorNetworksTensorOperationsExt/ITensorNetworksTensorOperationsExt.jl
JoeyT1994: It's because `TensorOperations` has some overflow of contraction costs if you pass integers. Converting to float allows it to avoid that and, Lukas says, shouldn't affect performance

mtfishman: I see, thanks. @lkdvos if that's the case can this conversion be handled inside of TensorOperations?

lkdvos: The problem is that this is quite an opinionated change: as you mentioned in principle you really do want to use `Int` for the computations, especially when you know that most of the time you'll never hit that overflow (and if you do, you probably can't contract it anyways). Apparently there are however some pathological cases in your test suite, which then leads to failures. Whenever this is the case, the next best thing is using `Float64`, which now leads to some loss of precision (cannot distinguish elements 1e-16 apart), but you can go up to 1e302 without overflow IIRC. True, floating point is a bit slower than Int, but this should really be minor, I think it's just a couple extra cycles. Finally, if you need the precision you could also switch to `BigInt`, but that would obviously not be great for performance.

From the TensorOperations side of things, it seems opinionated to make that choice for you, since we really don't want to try and analyze if the thing you provided might overflow or not, since that would cost about as much as running the algorithm. So, long story short, we leave that up to the users to figure out if they want precision, speed, accuracy, or a middle ground (which I'd say Float64 provides).
You could also consider to just keep it at `Int` and check if the test-case is just not very realistic

mtfishman: That makes sense, I agree with your assessment that automating this would be challenging and involves subtle tradeoffs between speed and safety. The only thing I could think of is having a `check_overflow=true/false` flag, but it is reasonable to just give users more control by inputting a different number type (say, a checked number type).

mtfishman: @JoeyT1994 can you add a comment explaining the reason for the conversion?

JoeyT1994: Sure, I added a comment

lkdvos:
It looks like something is blocking the update of `StridedViews` to v0.4. This doesn't really change functionality, but it (massively) cuts down the compilation times of TensorOperations. It might pay off to see what is holding that back.

mtfishman:
> It looks like something is blocking the update of `StridedViews` to v0.4. This doesn't really change functionality, but it (massively) cuts down the compilation times of TensorOperations. It might pay off to see what is holding that back.

Thanks for the heads up, I need to update the compat entry for StridedViews in NDTensors, looking into that now.

PR #222, opened 2025-03-31
Update gauge_walk functionality and fix bug in `VidalITensorNetwork`

JoeyT1994:
This PR contains the parts of PR #219 which were just quality of life improvements to the code.

Specifically the function for finding the `tree_gauge` to move the gauge of a tree tensor network from one region to another has been separated out so the path finding is done by a separate `gauge_path` function that operates on an abstract graph independent of the concept of an `ITensorNetwork` (which will be utilized in the upcoming `maximize_bilinearform` function).

A bug is also fixed in the `apply` function for the `VidalITensorNetwork` type.


src/abstractitensornetwork.jl
```diff
@@ -623,18 +623,44 @@ function gauge_walk(
   return gauge_walk(alg, tn, edgetype(tn).(edges); kwargs...)
 end
 
+function tree_gauge(alg::Algorithm, ψ::AbstractITensorNetwork, region)
+  return tree_gauge(alg, ψ, [region])
+end
+
+#Get the path that moves the gauge from a to b on a tree
+#TODO: Move to NamedGraphs
+function gauge_path(g::AbstractGraph, region_a::Vector, region_b::Vector)
```
mtfishman, line 632: Can we come up with a name that is more general that `gauge_path` for this, since now it isn't really specific to gauging? Maybe `trail_between_regions`?

Also, is this really a "path" by the definition in https://en.wikipedia.org/wiki/Path_(graph_theory)? It looks like in general this could be a "trail" since it is made from a spanning tree.

src/abstractitensornetwork.jl
```diff
@@ -623,18 +623,44 @@ function gauge_walk(
...
+  alg::Algorithm,
+  ψ::AbstractITensorNetwork,
+  cur_region::Vector,
+  new_region::Vector;
+  kwargs...,
+)
+  path = gauge_path(ψ, cur_region, new_region)
+  if !isempty(path)
+    ψ = gauge_walk(alg, ψ, path; kwargs...)
+  end
```
mtfishman, line 652: It seems like the empty case should be handled by the `gauge_walk` function.

src/abstractitensornetwork.jl
JoeyT1994: I think it may not even be a walk actually as the edges may not be contiguous (for instance it could be like `[3 => 2 , 1 => 2]` to move the gauge center from `[1,2,3]` to `[2]` on a 4-site path). Maybe `edge_sequence_between_regions` ?

mtfishman: Sounds good to me.

PR #228, opened 2025-04-04
Linear form network

JoeyT1994:
# Description

This PR introduces the `LinearFormNetwork` type, an `AbstractITensorNetwork` representing the overlap network `<x|y>`. This will be useful for the upcoming BoundaryMPS PR and  also fleshes out the different `FormNetwork` types supported by `ITensorNetworks`. Tests are included, along with a converter between a `BiLinearFormNetwork` and a `LinearFormNetwork`.


src/formnetworks/linearformnetwork.jl
```diff
@@ -0,0 +1,62 @@
...
+    copy(tensornetwork(lf)), bra_vertex_suffix(lf), ket_vertex_suffix(lf)
+  )
+end
+
+function update(
+  lf::LinearFormNetwork,
+  original_bra_state_vertex,
+  original_ket_state_vertex,
+  bra_state::ITensor,
+  ket_state::ITensor,
```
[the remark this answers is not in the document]
JoeyT1994: Good point, I have changed that now.

src/formnetworks/abstractformnetwork.jl
```diff
@@ -20,6 +20,13 @@ function SimilarType.similar_type(f::AbstractFormNetwork)
   return typeof(tensornetwork(f))
 end
 
+# TODO: Use `NamedGraphs.GraphsExtensions.parent_graph_type`.
+function data_graph_type(G::Type{<:AbstractFormNetwork})
+  return data_graph_type(fieldtype(G, :tensornetwork))
+end
```
mtfishman, line 26: I realize this was just moved in this PR but I noticed there is something off with this definition.

In general, it is best to not define definitions on abstract types that assume subtypes have a certain field (generic functions defined on abstract types should only depend on how objects _behave_, not how their data is stored).

Does this need to be in the type domain? I.e. an alternative could be:
```julia
function data_graph_type(f::AbstractFormNetwork)
  return data_graph_type(tensornetwork(g))
end
```
Other alternatives would be, to make the tensor network type a type parameter of `AbstractFormNetwork`, or expect subtypes to define this function themselves.

JoeyT1994: Good point, yeah this was assuming a field that may or may not exist.

I think moving it to the type domain should work, I just pushed that change.

mtfishman:
Looks good, thanks!

PR #232, opened 2025-04-11
Report Approximate Truncation Error from Apply Function

JoeyT1994:
This small PR allows passing a `truncation_error!` reference to the `apply` function in the case of the simple update routine. This will be filled in with the `truncerr` from the corresponding call to `factorize_svd` allowing one to keep track of the approximate error induced by the simple update routine. 

Note that this is not possible to infer solely from the `singular_values!` reference object due to the fact that the spectrum is not normalized by `factorize_svd!`. 

Tests are updated in `apply_test.jl` to make sure the functionality is working.


mtfishman:
@JoeyT1994 take a look at the callback design in `ITensorMPS.truncate!`: https://github.com/ITensor/ITensorMPS.jl/blob/v0.3.17/src/abstractmps.jl#L1692-L1714 (introduced in https://github.com/ITensor/ITensorMPS.jl/pull/99), I think that could be a good code pattern to use here as well since it is more general (it makes it easier for users to decide how to store the output, output different combinations of values, etc.).

JoeyT1994:
@mtfishman Oh yeah that's quite a nice way to do it. I have updated the code to use a callback for the `ITensorNetworks` interface instead

JoeyT1994:
@mtfishman I can't say I understand the reason for the failing tests. I am not seeing anything locally, it seems to be erroring here when running `test_contraction_sequence_to_graph.jl` but not reporting the reason for the exit code.

JoeyT1994:
Yeah maybe next time we see it I can try to remove the test that causes it

PR #234, opened 2025-04-25
Improve the BilinearForm Constructor

JoeyT1994:
This PR improves the `BilinearFormNetwork` and `QuadraticFormNetwork` constructors in the case an operator isn't explicitly passed and therefore the "identity" operator is assumed.
The constructor can now handle vertices with multiple (or no) site indices per vertex, whereas previously it could only handle strictly one site index per vertex. Tests are added.

Additionally:

1.  An error in the `VidalITensorNetwork` constructor is fixed, where the cache was being updated even when `update_cache= false` is passed.
2. The `EinExprs.jl` HyPar tests are disabled due to the fact they no longer work.
3. The `contraction_tree_to_graph.jl` code and tests has been removed as it is not being used anywhere in the library and I cannot envision any future use for it right now.

@emstoudenmire @anmello this is the bug we discussed yesterday


src/formnetworks/bilinearformnetwork.jl
```diff
@@ -60,9 +60,16 @@ function BilinearFormNetwork(
...
+  s = siteinds(ket)
+  s_mapped = dual_site_index_map(s)
+  operator_inds = union_all_inds(s, s_mapped)
+  constructor_f =
+    v -> inds -> if !isempty(inds)
+      reduce(*, [delta(dag(s), sm) for (s, sm) in zip(s[v], s_mapped[v])])
+    else
+      ITensor(one(Bool))
+    end
+  O = ITensorNetwork(constructor_f, operator_inds; link_space)
```
mtfishman, line 72:
```julia
function itensor_identity_map(i_pairs::Vector{<:Pair})
  return prod(i_pairs; init=ITensor(one(Bool))) do i_pair
    return delta(Bool, dag(first(i_pair)), last(i_pair))
  end
end
```
and then:
```suggestion
  O = ITensorNetwork(operator_inds; link_space) do v
    # Use `itensor_identity_map` on `s .=> s_mapped`.
  end
```
Hopefully that makes sense. There is something a bit funny that `operator_inds` is being passed but not really being used.

test/test_contraction_sequence.jl
```diff
@@ -49,9 +49,9 @@ using Test: @test, @testset
       Pkg.rm("KaHyPar"; io=devnull)
       res_kahypar_bipartite = contract(tn; sequence=seq_kahypar_bipartite)[]
       @test res_optimal ≈ res_kahypar_bipartite
-      seq_einexprs_kahypar = contraction_sequence(tn; alg="einexpr", optimizer=HyPar())
-      res_einexprs_kahypar = contract(tn; sequence=seq_einexprs_kahypar)[]
-      @test res_einexprs_kahypar ≈ res_optimal
+      # seq_einexprs_kahypar = contraction_sequence(tn; alg="einexpr", optimizer=HyPar())
+      # res_einexprs_kahypar = contract(tn; sequence=seq_einexprs_kahypar)[]
+      # @test res_einexprs_kahypar ≈ res_optimal
```
mtfishman, line 54: Could you add a comment saying these were leading to some some issue on CI that needs to be investigated?

JoeyT1994: Sure

src/formnetworks/bilinearformnetwork.jl
JoeyT1994: I changed it to that. Although I had to lower the type restriction on `itensor_identity_map` to avoid it erroring in the case `s[v] = Index[]` and `s_mapped[v] = Index[]`.

src/formnetworks/bilinearformnetwork.jl
```diff
@@ -60,9 +66,13 @@ function BilinearFormNetwork(
...
   link_space = isempty(flatten_siteinds(bra)) ? 1 : nothing
-  operator_inds = union_all_inds(siteinds(ket), dual_site_index_map(siteinds(ket)))
-  # TODO: Define and use `identity_network` here.
-  O = ITensorNetwork(Op("I"), operator_inds; link_space)
+  s = siteinds(ket)
+  s_mapped = dual_site_index_map(s)
+  operator_inds = union_all_inds(s, s_mapped)
+
+  O = ITensorNetwork(operator_inds; link_space) do v
+    inds -> itensor_identity_map(s[v] .=> s_mapped[v])
```
mtfishman, line 74:
```suggestion
    return inds -> itensor_identity_map(s[v] .=> s_mapped[v])
```

PR #236, opened 2025-05-09
Make the `AbstractBeliefPropagationCache` a subtype of the `AbstractITensorNetwork`

JoeyT1994:
This PR sets`AbstractBeliefPropagationCache <: AbstractITensorNetwork` so that functionality defined on that type can be utilized.

Specifically it allows accession of the tensors in the tensor network inside a `bpc::AbstractBeliefPropagationCache` via `bps[v]`  and operations that were previously forwarded onto the network to immediately work (finding indices etc). Extra tests are added in `test/test_beliefpropagation.jl`.

Setting of the tensors inside the cache via `bpc[v] = some_itensor` is disabled just to the fact it may affect partitioning and message tensor information. Currently a `set_index_preserve_graph` version is supported, although maybe this should be changed to `set_index_preserve_graph_and_messages` so it is explicit to the user what they are doing. 

Most importantly, this PR will allow us to more naturally use the `BeliefPropagationCache` representing say `<x|A|y>`  in `Problem` types in `NetworkSolvers` without also requiring a separate copy of the state (`|x>`) to be passed around which is what the current implementation does.


src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -241,11 +286,11 @@ function update(
   edge_groups::Vector{<:Vector{<:PartitionEdge}};
   kwargs...,
 )
-  new_mts = copy(messages(bpc))
+  new_mts = Dictionary()
```
mtfishman, line 289:
```suggestion
  new_mts = empty(messages(bpc))
```
That would help to preserve the type of the message cache.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -40,6 +49,9 @@ default_messages(ptn::PartitionedGraph) = Dictionary()
 end
 default_partitioned_vertices(ψ::AbstractITensorNetwork) = group(v -> v, vertices(ψ))
 
+function Base.setindex!(bpc::AbstractBeliefPropagationCache, factor::ITensor, vertex)
+  not_implemented()
```
mtfishman, line 53:
```suggestion
  return not_implemented()
```

PR #237, opened 2025-05-13
Use @preserve_graph in`gauge_edge(Algorithm::"orthogonalize", tn::AbstractITensorNetwork, edge, ...)` and `truncate(tn::AbstractITensorNetwork, edge, ...)`

JoeyT1994:
This PR changes `gauge_edge(Algorithm::"orthogonalize", tn::AbstractITensorNetwork, edge, ...)` and `truncate(tn::AbstractITensorNetwork, edge, ...)` to use graph preserving updates of the `tn::AbstractITensorNetwork` to improve their efficiency. 

The current use of `orthogonalize(...)` in `alternating_update` (now `sweep_solver` in `NetworkSolvers`) involves rechecking the graph structure every time `gauge_edge` is called and thus invokes an `O(L)` time complexity call in a step that should be `O(1)` where `L` is the system-size. 

It is not clear to me there is a case where these operations should re-check the graph structure as they implicitly conserve it from the looks of things provided the factorization is done correctly...

@emstoudenmire @mtfishman


mtfishman:
Makes sense to me, I think it is fair to assume they preserve the graph.

The only case I could imagine is if the edge isn't actually in the graph originally. It would be strange to gauge along a non-existent edge, so probably we should check for that case in those functions and users will have to manually add edges they want to gauge/truncate along if they don't exist yet. Can you add a check in those functions you modified that the edge exists in the graph?

JoeyT1994:
Agreed it would be strange to gauge on a non-existing edge. I added a check based on the neighbors of `dst(edge)` so the check doesn't scale with system size.

src/abstractitensornetwork.jl
```diff
@@ -603,12 +603,13 @@ function gauge_edge(
   # # TODO: Implement as `only(common_neighbors(tn, src(edge), dst(edge)))`
   # new_vertex = only(neighbors(tn, src(edge)) ∩ neighbors(tn, dst(edge)))
   # return contract(tn, new_vertex => dst(edge))
+  @assert src(edge) ∈ neighbors(tn, dst(edge))
```
mtfishman, line 606: It is best to avoid `@assert`, it is meant to be used for low level checks of things that should never happen, rather than checking bad inputs, and could get removed by certain optimization levels (https://docs.julialang.org/en/v1/base/base/#Base.@assert). I think the right error to throw is `throw(ArgumentError("Edge not in graph."))` since that is the error to throw for bad arguments.

mtfishman:
> I changed it to `has_edge(g, e)` which (from my testing) appears to also be system-size independent in terms of time and memory allocations. `e in edges(g)` does appear to scale with system-size and also is dependent on the order of the edges. I guess it's not calling `has_edge` but is just checking `e` against the whole list of edges?

Thanks for checking, good to know. Graphs.jl can make `e in edges(g)` call `has_edge` for `SimpleGraph` because `edges(::SimpleGraph)` outputs a special type (https://github.com/JuliaGraphs/Graphs.jl/blob/v1.12.1/src/SimpleGraphs/simpleedgeiter.jl#L119), while `edges(::NamedGraph)` outputs an arbitrary collection. We should raise an issue about that in NamedGraphs.jl.

PR #239, opened 2025-05-24
Identity map fix v2

JoeyT1994:
This PR updates the `itensor_identity_map` function so that the `eltype` of the identity tensors can be specified. A default of `Float64` is used if they aren't.

A default `link_space=1` is also provided to the `random_itensornetwork()` function such that `random_itensornetwork(s::IndsNetwork)` now works and produces a product state network if the `IndsNetwork` contains quantum numbers (previously it would error due to fluxes not being correct).


mtfishman:
Looks good, thanks for the quick fix.

PR #242, opened 2025-07-02
Network solvers redesign

emstoudenmire:
This PR is a rewrite of codes for "sweep solver" algorithms such as DMRG, TDVP, etc.
It introduces a simplified designs for the whole system, especially regarding the creation of "region plans", how the double loop comprising what is now called `sweep_solve` is coded, and the handling of keyword arguments.

The PR also adds a subspace expansion system with one currently implementation (based on a "projected density matrix perturbation" idea, basically McCulloch+White hybrid method). Having a subspace expansion is crucial to converge DMRG on certain tree lattices, not to mention other cases like 2D DMRG with QN conservation.

Internally, the core design revolves around two iterators: a sweep iterator and a region iterator. Each of these is in principle wrappable by "iteration adapters" (see for example the tuple adapter in adapters.jl). At each iteration of the RegionIterator type, it calls a `region_iterator_action` function which can be overloaded, but by default calls three "subactions": `extracter`, `updater`, and `inserter`. Currently `extracter` also calls down into a `subspace_expand` function which can be customized through multiple backends. These "action" functions all dispatch on a "problem" type which can hold arbitrary data, making these codes more flexible and future-proof toward cases like optimizing two tensor network states at once, working with sets of operators, etc.

Other improvements that may fit better into future PRs:
* install the "fitting" implementations of `truncate` and `apply` already in the NetworkSolvers repo (there were some issues bringing them over)
* replace the use of ProjTTN with a BP cache
* more friendly interfaces to methods like `dmrg` which automatically propagate truncation arguments into more "expert" keyword argument packs
* helper functions for making keyword parameter packs and propagating defaults
* discuss what is the best strategy to optionally truncate the bonds in 1-site TDVP, perhaps during the orthogonalize/gauge_walk part of the extracter step
* improve the subspace expansion code to expand all bonds around the current region and simplify the current code (the first change may already help to simplify the code somewhat)

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -5,5 +5,5 @@ Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 
 [compat]
 Documenter = "1.10.0"
-ITensorNetworks = "0.14.0"
+ITensorNetworks = "0.15.0"
 Literate = "2.20.1"


emstoudenmire:
Is the version check meaning I should just bump the version number? Which version number should I bump to?

mtfishman:
> Is the version check meaning I should just bump the version number? Which version number should I bump to?

The version check is checking whether or not you've bumped the package version here: https://github.com/ITensor/ITensorNetworks.jl/blob/2d7db49c2f003c6588a1632fd6f0fd802c0ac9ca/Project.toml#L4. The reason for the check is that in general we are trying to enforce that we always bump the package version in every PR and register a new version after the PR is merged, so that versions don't drag out and accumulate too many changes, turnover time to release bug fixes and features for users is faster, etc. I'm sure this PR is breaking in some way so you should bump the version to v0.14.

src/solvers/permute_indices.jl
```diff
@@ -0,0 +1,16 @@
+
+function permute_indices(tn)
```
mtfishman, line 2:
```suggestion
function align_indices(tn)
```
I've started to use the terminology `align` instead of `permute` for ordering the indices of an ITensor/NamedDimsArray in NamedDimsArrays.jl, I think it expresses the intent better.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,76 @@
...
+function updater(
+  E::EigsolveProblem,
+  local_state,
+  region_iterator;
+  outputlevel,
+  solver=eigsolve_solver,
+  kws...,
+)
+  eigval, local_state = solver(ψ->optimal_map(operator(E), ψ), local_state; kws...)
+  E = setproperties(E; eigenvalue=eigval)
```
mtfishman, line 23:
```suggestion
  E = set_eigenvalue(E, eigval)
```

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,76 @@
...
+  operator::Operator
+  eigenvalue::Number = Inf
+end
+
+eigenvalue(E::EigsolveProblem) = E.eigenvalue
+ITensorNetworks.state(E::EigsolveProblem) = E.state
+operator(E::EigsolveProblem) = E.operator
+
+function updater(
+  E::EigsolveProblem,
```
mtfishman, line 15:
```suggestion
  problem::EigsolveProblem,
```
I think it would be helpful to stay consistent with the notation and always call the problem `problem`.

src/solvers/adapters.jl
```diff
@@ -0,0 +1,32 @@
...
+# TupleRegionIterator
+#
+# Adapts outputs to be (region, region_kwargs) tuples
+#
+# More generic design? maybe just assuming RegionIterator
+# or its outputs implement some interface function that
+# generates each tuple?
+#
+
+mutable struct TupleRegionIterator{RegionIter}
```
mtfishman, line 12: Maybe we can come up with a name that is more descriptive, like `RegionIteratorReturnKwargs`. `Tuple` is a bit vague.

src/solvers/applyexp.jl
```diff
@@ -0,0 +1,103 @@
+using Printf: @printf
+import ConstructionBase: setproperties
+
+@kwdef mutable struct ApplyExpProblem{State}
+  state::State
+  operator
+  current_time::Number = 0.0
+end
+
+ITensorNetworks.state(A::ApplyExpProblem) = A.state
```
mtfishman, line 10:
```suggestion
state(A::ApplyExpProblem) = A.state
```
since we are in the `ITensorNetworks` module.

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
...
+    sweep_kws_state = iterate(S.sweep_kws)
+  else
+    sweep_kws_state = iterate(S.sweep_kws, which)
+  end
+  isnothing(sweep_kws_state) && return nothing
+  current_sweep_kws, next = sweep_kws_state
+
+  if !isnothing(which)
+    S.region_iter = region_iterator(
+      problem(S.region_iter); sweep=S.which_sweep, current_sweep_kws...
```
mtfishman, line 26:
```suggestion
      problem(S); sweep=S.which_sweep, current_sweep_kws...
```
I think, based on the definition of `problem(::SweepIterator)` above.

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
...
+  R.which_region==length(R.region_plan) ? nothing : R.region_plan[R.which_region + 1][1]
+end
+is_last_region(R::RegionIterator) = isnothing(next_region(R))
+
+function Base.iterate(R::RegionIterator, which=1)
+  R.which_region = which
+  region_plan_state = iterate(R.region_plan, which)
+  isnothing(region_plan_state) && return nothing
+  (current_region, region_kwargs), next = region_plan_state
+  R.problem = region_iterator_action(problem(R), R; region_kwargs...)
```
mtfishman, line 69: I'm not a big fan of the name `region_iterator_action` but I can't think of anything much better. The only thing I can think of is `region_update` but I don't know how I feel about reusing the term `update`. Maybe `region_step`?

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
+#
+# SweepIterator
+#
+
+mutable struct SweepIterator
+  sweep_kws
+  region_iter
+  which_sweep::Int
+end
```
mtfishman, line 9: What about parameterizing this  struct by the types of the fields `sweep_kws` and `region_iter`? Is it meant to be dynamic, i.e. is it expected that those types might change?

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
...
+
+@kwdef mutable struct RegionIterator{Problem,RegionPlan}
+  problem::Problem
+  region_plan::RegionPlan
+  which_region::Int = 1
+end
+
+problem(R::RegionIterator) = R.problem
+current_region_plan(R::RegionIterator) = R.region_plan[R.which_region]
+current_region(R::RegionIterator) = current_region_plan(R)[1]
```
mtfishman, line 54: What's the data structure for the output of `current_region_plan(R)`? Something doesn't feel write to me that the region is accessed with indexing by `1`, maybe it should be a `NamedTuple` and the region could be accessed as `current_region_plan(R).region` or a struct where it is accessed with a function `get_region(current_region_plan(R))`?

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
...
+region_kwargs(R::RegionIterator) = current_region_plan(R)[2]
+function previous_region(R::RegionIterator)
+  R.which_region==1 ? nothing : R.region_plan[R.which_region - 1][1]
```
mtfishman, line 57: It seems like maybe we should be using function accessors rather than field accessors, i.e. `which_region(R)` instead of `R.which_region`.

src/solvers/sweep_solve.jl
```diff
@@ -0,0 +1,40 @@
+
+default_region_callback(problem; kws...) = nothing
+
+default_sweep_callback(problem; kws...) = nothing
+
+function default_sweep_printer(problem; outputlevel, sweep, nsweeps, kws...)
```
mtfishman, line 6: Similar to my comment above, what about calling these just `region_callback`, `sweep_callback`, and `sweep_printer`, since they would get overloaded on the `problem` type anyway?

src/solvers/sweep_solve.jl
```diff
@@ -0,0 +1,40 @@
...
+  outputlevel=0,
+  region_callback=default_region_callback,
+  sweep_callback=default_sweep_callback,
+  sweep_printer=default_sweep_printer,
+  kwargs...,
+)
+  for (sweep, region_iter) in enumerate(sweep_iterator)
+    for (region, region_kwargs) in region_tuples(region_iter)
+      region_callback(
+        problem(region_iter);
```
mtfishman, line 23: I noticed this gets passed the problem while `sweep_callback` and `sweep_printer` get passed the `region_iter`, is there a logic behind that? I would lean towards all of them getting passed the `region_iter`, and there could be a layer that unwraps the problem as needed.

test/solvers/test_tree_dmrg.jl
```diff
@@ -0,0 +1,64 @@
+using Test: @test, @testset
+using ITensors
+using TensorOperations # Needed to use contraction order finding
```
mtfishman, line 3:
```suggestion
using TensorOperations: TensorOperations # Needed to use contraction order finding
```
That only brings `TensorOperations` itself into the namespace.

test/solvers/utilities/tree_graphs.jl
```diff
@@ -0,0 +1,23 @@
+import Graphs as gr
+import NamedGraphs as ng
```
mtfishman, line 2: For the sake of keeping the style consistent throughout the library I would prefer:
```julia
using Graphs: add_edge!, ...
using NamedGraphs: NamedGraph, ...
```
If we don't stay consistent with the style it leads to chaos where code is written in different styles and it is a lot of work to go back and make it consistent later.

src/solvers/extracter.jl
```diff
@@ -0,0 +1,17 @@
+import ConstructionBase: setproperties
+
+function extracter(problem, region_iterator; sweep, trunc=(;), kws...)
```
mtfishman, line 3: If we don't change this to `extract` as discussed elsewhere, which should change this to `extractor` since `extracter` isn't the correct spelling.

src/solvers/applyexp.jl
```diff
@@ -0,0 +1,103 @@
+using Printf: @printf
+import ConstructionBase: setproperties
+
+@kwdef mutable struct ApplyExpProblem{State}
+  state::State
+  operator
+  current_time::Number = 0.0
+end
```
mtfishman, line 8: Related to my comment above about `SweepIterator`, I think we should parameterize the problem types. Also related to my other comment I think we should switch the argument ordering so the `operator` is first.

Also, I'm suspicious when I see `0.0` since that is a `Float64`. If the time is `Float64` it might then promote the element type of other objects it interacts with to double precision, which would be bad on GPU where single precision is better. We should have a constructor that determines the number type from the inputs, such as: `ApplyExpProblem(operator, state) = ApplyExpProblem(operator, state, zero(promote_type(scalartype(operator), scalartype(state)))`.

src/solvers/truncation_parameters.jl
```diff
@@ -0,0 +1,14 @@
+default_maxdim() = typemax(Int)
+default_mindim() = 1
+default_cutoff() = 0.0
```
mtfishman, line 3:
```suggestion
default_cutoff(elt::Type{<:Real}) = zero(elt)
```
so that we don't assume double precision (`Float64`).

src/solvers/subspace/ortho_subspace.jl
```diff
@@ -0,0 +1,77 @@
...
+  ax_space = expand_space(space(ci), expansion_factor)
+  ax = Index(ax_space, "ax")
+
+  linear_map(w) = (w - A * (dag(A) * w))
+  Y = linear_map(random_itensor(basis_inds, dag(ax)))
+  expand_maxdim = compute_expansion(
+    dim(a), basis_size; expansion_factor, max_expand, maxdim
+  )
+  (norm(Y) <= 1E-15 || expand_maxdim <= 0) && return local_tensor
+  Ux, S, V = svd(Y, basis_inds; cutoff=1E-14, maxdim=expand_maxdim, lefttags="ux,Link")
```
mtfishman, line 66: We should determine an element type/scalar type `elt` from the inputs and replace `1E-15` and `1E-14` with something like `10 * eps(elt)` to be more agnostic about using single precision or double precision, to be friendlier for running on GPU.

test/solvers/test_tree_dmrg.jl
```diff
@@ -0,0 +1,64 @@
...
+using ITensorNetworks: siteinds, ttn, dmrg
+import Graphs: dst, edges, src
+import ITensorMPS: OpSum
+
+include("utilities/simple_ed_methods.jl")
+include("utilities/tree_graphs.jl")
+
+@testset "Tree DMRG" begin
```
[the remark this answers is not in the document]
mtfishman: For an example of how I also test on GPU, you can see: https://github.com/ITensor/BlockSparseArrays.jl/blob/44574b808861a03c789a74620d8308f3b8c1b8d4/test/test_basics.jl#L49-L64.

src/solvers/eigsolve.jl
```diff
@@ -0,0 +1,76 @@
...
+)
+  eigval, local_state = solver(ψ->optimal_map(operator(E), ψ), local_state; kws...)
+  E = setproperties(E; eigenvalue=eigval)
+  if outputlevel >= 2
+    @printf("  Region %s: energy = %.12f\n", current_region(region_iterator), eigenvalue(E))
+  end
+  return E, local_state
+end
+
+function eigsolve_sweep_printer(region_iterator; outputlevel, sweep, nsweeps, kws...)
```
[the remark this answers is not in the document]
emstoudenmire: I see, yes I guess this makes sense in case some one (a developer or user) bypasses the `eigsolve` interface but still uses an `EigSolveProblem` and wants to get this printing behavior. Interesting pattern about the `EigSolveRegionIterator` - I'll see if that helps. (It has definitely been useful to have some way of accessing the region iterator in the printing functions and other implementation functions like subspace expansion.)

src/solvers/applyexp.jl
emstoudenmire: I'll look into what error I was getting when I didn't prepend `ITensorNetworks`. I agree it shouldn't be needed in principle.

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
...
+  isnothing(sweep_kws_state) && return nothing
+  current_sweep_kws, next = sweep_kws_state
+
+  if !isnothing(which)
+    S.region_iter = region_iterator(
+      problem(S.region_iter); sweep=S.which_sweep, current_sweep_kws...
+    )
+  end
+  S.which_sweep += 1
+  return S.region_iter, next
```
mtfishman, line 30: I'm a bit on the fence about this, but I would lean towards a design where `iterate(::SweepIterator)` actually performs the region iteration as opposed to returning the region iterator, so then `for _ in sweep_iterator end` actually runs the calculation. Then, we could have an iteration adapter such as `region_iterators(::SweepIterator)` returns an iterator that returns the region iterator at each iteration.

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
+#
+# SweepIterator
+#
+
+mutable struct SweepIterator
+  sweep_kws
+  region_iter
```
mtfishman, line 7: Maybe this should be called `current_region_iter` or something like that to indicate it is the latest region iterator.

I have to say this part confused me a bit. I see the region iterator is recreated from scratch from the `sweep_kws` at each sweep, is there a reason to store it? It seems like it could just be made and used "on the fly" at each sweep anyway.

src/solvers/iterators.jl
```diff
@@ -0,0 +1,104 @@
+#
+# SweepIterator
+#
+
+mutable struct SweepIterator
+  sweep_kws
```
mtfishman, line 6: Maybe this could be named `each_sweep_kwargs` or `sweep_kwargs_iterator` to indicate that it itself is an iterator that returns the keyword arguments for the sweep at each iteration (rather than just the keyword arguments of the latest sweep).

src/solvers/iterators.jl
mtfishman: For reference, in DifferentialEquations.jl, `iterate` is just defined as:
```julia
function Base.iterate(integrator::DEIntegrator, state = 0)
    done(integrator) && return nothing
    state += 1
    step!(integrator)
    return integrator, state
end
```
which seems like a reasonable goal to aim for, and then the complexity of the implementation is in `done(...)` and `step!(...)`.

test/solvers/utilities/tree_graphs.jl
emstoudenmire: Yes, I meant to change these to the current style of ITensorNetworks. I will do that and in any other files I find.

src/solvers/applyexp.jl
```diff
@@ -0,0 +1,103 @@
...
+  extracter_kwargs=(;),
+  updater_kwargs=(;),
+  inserter_kwargs=(;),
+  outputlevel=0,
+  nsites=1,
+  tdvp_order=4,
+  sweep_printer=applyexp_sweep_printer,
+  kws...,
+)
+  time_steps = diff([0.0, exponents...])[2:end]
```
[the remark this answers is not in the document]
emstoudenmire: Good to know, thanks

src/solvers/iterators.jl
emstoudenmire: Sounds good – changed to `region_step`

emstoudenmire:
Thanks, Matt. I'll make those Project.toml changes. 

(By the way, I may be pushing some commits to this PR but in a way that doesn't signal that all of the review-related changes have been implemented yet. The reason is that I'm sharing this branch with Jason K. as a backend for some TDVP experiments he's helping me with.)

src/solvers/applyexp.jl
```diff
@@ -0,0 +1,121 @@
...
+  operator
+  state::State
+  current_exponent::Number = 0.0
+end
+
+ITensorNetworks.state(A::ApplyExpProblem) = A.state
+operator(A::ApplyExpProblem) = A.operator
+current_exponent(A::ApplyExpProblem) = A.current_exponent
+function current_time(A::ApplyExpProblem)
+  t = im*A.current_exponent
```
mtfishman, line 13:
```suggestion
  t = im * current_exponent(A)
```
Just a style change.

src/solvers/applyexp.jl
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  prob = set_current_exponent(prob, current_exponent(prob)+exponent_step)
+
+  return prob, local_state
+end
+
+function sweep_printer(
+  problem::ApplyExpProblem;
+  exponent_description="exponent",
```
mtfishman, line 67: This input is a bit odd to me, let's discuss this.

src/apply.jl
JoeyT1994, line 1: A similar change to the function ```
function ITensors.apply(
  o,
  ψ::VidalITensorNetwork; ...)``` lower down (lime 322) will be necessary to avoid ambiguity and the current failing test.

src/solvers/iterators.jl
emstoudenmire: Yes, keeping the problem in the sweep iterator could simplify the mechanics quite a lot. I was thinking originally like "better to have fewer fields in SweepIterator, especially since the region iterator will have a copy of the problem" but I think on balance it's best not to have complicated code handling creating each region iterator.

PR #250, opened 2025-08-02
Fix map_eigvals for fermionic case and test

emstoudenmire:
Along similar lines to the recent PR in ITensors.jl, this PR fixes the `map_eigvals` function in ITensorNetworks.ITensorsExtensions to handle the case of fermionic tensors correctly. The PR includes tests of the case where the non-linear map is the function `sqrt`, checking that the resulting tensors square back to the original one.


mtfishman:
@emstoudenmire I noticed in #248 some TEBD tests are failing in this arrow direction check: https://github.com/ITensor/ITensors.jl/blob/c019c65a0f3cb4d7a88e57714bbcfa3986f56de4/src/tensor_operations/matrix_algebra.jl#L79-L84 introduced in https://github.com/ITensor/ITensors.jl/pull/1661.

I think the issue might be that in some of the tests, the auto fermion system is enabled but a non-QN calculation is being run, and those checks are assuming the tensors have QNs.

emstoudenmire:
Latest updates to this PR include changing the code to be more similar to the final version of the recent ITensors `exp` PR, and including a test of `map_eigvals` for a bosonic tensor with the auto fermion system turned on.

You may want to have a look at lines 68 and 69 of the `itensor.jl` file (first two lines of `map_eigvals` where I found that the rest of the code had trouble handing the case where `Linds` and `Rinds` weren't collections (vectors or tuples) but were individual indices. This made later code like calling `reverse(Rinds)...` not work correctly. I thought calling `collect` on `Linds` and `Rinds` might be the solution but for an individual `Index` `i`, I think `collect(i)` tries to iterate the Index so doesn't have the same effect as writing `[i]`.

src/lib/ITensorsExtensions/src/itensor.jl
```diff
@@ -58,18 +61,53 @@ function eigendecomp(A::ITensor, linds, rinds; ishermitian=false, kwargs...)
...
   ul, ur = noncommonind(D, U), commonind(D, U)
   Ul = replaceinds(U, vcat(rinds, ur), vcat(linds, ul))
-
   return Ul, D, dag(U)
 end
 
-function map_eigvals(f::Function, A::ITensor, inds...; ishermitian=false, kwargs...)
+function map_eigvals(f::Function, A::ITensor, Linds, Rinds; kws...)
+  Linds = isa(Linds, Index) ? [Linds] : collect(Linds)
+  Rinds = isa(Rinds, Index) ? [Rinds] : collect(Rinds)
```
mtfishman, line 69: Are there cases where we call this with `Linds::Index` and/or `Rinds::Index`? I would lean towards just not supporting that, and users have to call `map_eigvals(f, A, (i,), (j,))` or `map_eigvals(f, A, [i], [j])`.

emstoudenmire: Yes I kind of would too... Especially if we are setting these functions up to be more like an "expert" back end to a future ITensor operator system. 

There were some tests that triggered this issue. Let me track them down and then the right move could be to just improve those tests.

emstoudenmire: Ok, it was my own fault for writing a test with a line like 
```
sqrtM = map_eigvals(sqrt, M, s1', dag(s1); ishermitian=true)
```
which I've now changed to have brackets around the indices to make collections.

Do you have a feel for how we could get a better error message in that case? Like constraining the arguments or checking if the inputs satisfy the concept of a collection?

src/lib/ITensorsExtensions/src/itensor.jl
```diff
@@ -58,18 +61,53 @@ function eigendecomp(A::ITensor, linds, rinds; ishermitian=false, kwargs...)
...
+    else
+      error(
+        "For fermionic exp, Linds and Rinds must have same directions within each set. Got dir.(Linds)=",
+        dir.(Linds),
+        ", dir.(Rinds)=",
+        dir.(Rinds),
+      )
+    end
+    A = permute(A, ordered_inds)
+    # A^n now sign free, ok to temporarily disable fermion system
```
mtfishman, line 92: I should have suggested this in the `exp` PR, but probably it would be better to split this logic off into a separate function for the sake of readability (and maybe so that the logic can be reused elsewhere).

Maybe this could be split off into a function `A = make_bosonic(A::ITensor, Linds, Rinds)`. In that design, I believe `ordered_inds` used down below could be determined from `inds(A)`.

src/lib/ITensorsExtensions/src/itensor.jl
emstoudenmire: I implemented this. Interestingly, after the check I caught some preexisting tests passing in individual indices. I guess that worked because it didn't interact with the newer `make_bosonic` code and the lower-down `eigen` call can handle that case. `eigen` does it by calling `indices` on its inputs. So we could go that direction too, as a way of "normalizing" the inputs and then allowing individual Index objects to be passed.

emstoudenmire:
Some of the other tests, such as in `test_apply.jl` in some places, are using the `map_eigvals` function with individual `Index` objects passed as the `Linds,Rinds` arguments instead of collections. So this stricter behavior is looking more like a breaking change. Of course we are allowing breaking changes in ITensorNetworks, but in hindsight should I just put lines like:
```
Linds = indices(Linds)
Rinds = indices(Rinds)
```
at the top of `map_eigvals` to allow broader types of inputs like individual indices? Or do we want to move these functions to being more "expert" interfaces with narrower allowed inputs?

mtfishman:
> Some of the other tests, such as in `test_apply.jl` in some places, are using the `map_eigvals` function with individual `Index` objects passed as the `Linds,Rinds` arguments instead of collections. So this stricter behavior is looking more like a breaking change. Of course we are allowing breaking changes in ITensorNetworks, but in hindsight should I just put lines like:
> 
> ```
> Linds = indices(Linds)
> Rinds = indices(Rinds)
> ```
> 
> at the top of `map_eigvals` to allow broader types of inputs like individual indices? Or do we want to move these functions to being more "expert" interfaces with narrower allowed inputs?

I see, thanks for investigating this. I forgot that `eigen` supports individual `Index` inputs and canonicalizes in that way, let's just do that here to match the behavior and design of `eigen`.

PR #253, opened 2025-08-17
Allow BP to work on gpus and various kwarg improvements

JoeyT1994:
This PR adds GPU support for the `BeliefPropagationCache`, allowing the structure to be `adapted` to GPUs via overload of the `adapt_structure` function on both the tensor network and underlying messages. In doing so, `adapt` has been made a core dependency of the library and is no longer a weak dependency.

Moreover a `custom_device` message update alg has been defined which allows the tensors in a message update to be passed to a designated device (such as GPU) and then brought back to the device the tensors were originally on after contraction.

To enable these features, the `kwargs` of the BP code have been improved such that they are expected to be passed as kwargs inside the algorithm `struct` specified for BP. E.g. `update(bp_cache; alg = Algorithm("bp"; maxiter = , tol = , edge_sequence = ....))`. A reasonable default is defined.

Meanwhile the `message_update` method can be passed as `message_update_alg = Algorithm("contract"; kwargs...)` with the relevant kwargs contained within there. 

The following code is now supported

```
using CUDA
ψ = random_tensornetwork(rng, s; link_space=χ)
ψψ = QuadraticFormNetwork(ψ)
bpc = BeliefPropagationCache(ψψ)

#Move the cache to Nvidia GPU and do BP on there
gpu_bpc = CUDA.cu(bpc)
gpu_bpc = update(gpu_bpc)

#Alternatively, keep the cache on CPU and just use a GPU to do updates
bpc = update(bpc; message_update_alg = Algorithm("contract_custom_device"; normalize = true, custom_device_adapt =     CUDA.cu, sequence_alg = "optimal")
```


src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -24,26 +25,48 @@ function data_graph_type(bpc::AbstractBeliefPropagationCache)
...
 data_graph(bpc::AbstractBeliefPropagationCache) = data_graph(tensornetwork(bpc))
 
-function default_message_update(contract_list::Vector{ITensor}; normalize=true, kwargs...)
-  sequence = contraction_sequence(contract_list; alg="optimal")
-  updated_messages = contract(contract_list; sequence, kwargs...)
+function message_update_function(
+  alg::Algorithm"contract",
+  contract_list::Vector{ITensor};
+  normalize=alg.kwargs.normalize,
+  sequence_alg=alg.kwargs.sequence_alg,
```
mtfishman, line 32: I would lean towards not having these keyword arguments and just have the keyword arguments passed through `alg`, I think otherwise it starts getting confusing keeping track of how to pass the keyword arguments, which ones take precedence, etc.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -24,26 +25,48 @@ function data_graph_type(bpc::AbstractBeliefPropagationCache)
...
 function message_diff(message_a::Vector{ITensor}, message_b::Vector{ITensor})
   lhs, rhs = contract(message_a), contract(message_b)
   f = abs2(dot(lhs / norm(lhs), rhs / norm(rhs)))
   return 1 - f
 end
 
-default_message(elt, inds_e) = ITensor[denseblocks(delta(elt, i)) for i in inds_e]
+function default_message(datatype, elt, inds_e)
+  ITensor[adapt(datatype, denseblocks(delta(elt, i))) for i in inds_e]
+end
```
mtfishman, line 67: I think it is confusing passing both the `datatype` and `eltype`, since normally the `eltype` will be specified as part of the `datatype`. Also I think there is a missing `return`, and I think we don't need to specify the type as `ITensor`.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -144,6 +152,28 @@ function incoming_messages(
...
+  end
+  return bpc
+end
+function adapt_factors(to, bpc::AbstractBeliefPropagationCache)
+  bpc = copy(bpc)
+  for v in vertices(bpc)
+    @preserve_graph bpc[v] = adapt(to).(bpc[v])
+  end
+  return bpc
+end
```
mtfishman, line 169: Maybe these could be written more generally as `map_messages` and `map_factors` that take any function, and then `adapt_structure` could be written in terms of `map_messages(adapt(to), bpc)`, etc.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -24,26 +25,48 @@ function data_graph_type(bpc::AbstractBeliefPropagationCache)
...
   return 1 - f
 end
 
-default_message(elt, inds_e) = ITensor[denseblocks(delta(elt, i)) for i in inds_e]
+function default_message(datatype, elt, inds_e)
+  ITensor[adapt(datatype, denseblocks(delta(elt, i))) for i in inds_e]
+end
 default_messages(ptn::PartitionedGraph) = Dictionary()
-@traitfn default_bp_maxiter(g::::(!IsDirected)) = is_tree(g) ? 1 : nothing
+@traitfn default_bp_maxiter(g::::(!IsDirected)) = is_tree(g) ? 1 : 30
```
mtfishman, line 69: Is there something special about 30 max iterations? I feel uncomfortable picking an arbitrary number for that, but also appreciate that you don't want the code to just run forever. Maybe that could be something we require users to specify (that's what we do in `ITensorMPS.dmrg`).

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -24,26 +25,48 @@ function data_graph_type(bpc::AbstractBeliefPropagationCache)
...
+  updated_messages = contract(contract_list; sequence)
   message_norm = norm(updated_messages)
   if normalize && !iszero(message_norm)
     updated_messages /= message_norm
   end
   return ITensor[updated_messages]
 end
 
+function message_update_function(
+  alg::Algorithm"contract_custom_device",
```
mtfishman, line 44: I think we should make this more general and not specific to using "contract" as an update. For example, we could call it `Algorithm"adapt_update"` and expect it to have keyword arguments `alg.kwargs.adapt` and `alg.kwargs.alg`, where `alg.kwargs.alg` gets passed as the algorithm to `message_update_function` once the messages are adapted.

If users are using "contract" a lot I think they could make a more convenient wrapper if they want that for convenience, but I think writing:
```julia
update(bpc; message_update_alg=Algorithm("adapt_update"; adapt=CUDA.cu, alg=Algorithm("contract"; normalize=true, sequence_alg = "optimal")))
```
isn't so bad and makes it clearer to users what is going on and how they could customize the behavior.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -24,26 +25,48 @@ function data_graph_type(bpc::AbstractBeliefPropagationCache)
...
+  normalize=alg.kwargs.normalize,
+  sequence_alg=alg.kwargs.sequence_alg,
+  custom_device_adapt=alg.kwargs.adapt,
+)
+  adapted_contract_list = custom_device_adapt.(contract_list)
+  updated_messages = message_update_function(
+    Algorithm("contract"), adapted_contract_list; normalize, sequence_alg
+  )
+  dtype = datatype(first(contract_list))
+  return ITensor[adapt(dtype, updated_message) for updated_message in updated_messages]
```
mtfishman, line 55:
```suggestion
  return map(adapt(dtype), updated_messages)
```

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -24,26 +25,48 @@ function data_graph_type(bpc::AbstractBeliefPropagationCache)
 end
 data_graph(bpc::AbstractBeliefPropagationCache) = data_graph(tensornetwork(bpc))
 
-function default_message_update(contract_list::Vector{ITensor}; normalize=true, kwargs...)
-  sequence = contraction_sequence(contract_list; alg="optimal")
-  updated_messages = contract(contract_list; sequence, kwargs...)
+function message_update_function(
```
mtfishman, line 28:
```suggestion
function message_update(
```
Just simpler and I don't think `_function` is needed now that we're using `Algorithm` more.

src/caches/abstractbeliefpropagationcache.jl
mtfishman: Note that we could have two versions:
```julia
function default_message(datatype::Type{<:AbstractArray}, inds_e)
  return [adapt(datatype, denseblocks(delta(i))) for i in inds_e]
end

function default_message(elt::Type{<:Number}, inds_e)
  return default_message(Vector{elt}, inds_e)
end
```

src/caches/beliefpropagationcache.jl
```diff
@@ -69,19 +68,40 @@ function Base.copy(bp_cache::BeliefPropagationCache)
...
+    maxiter=default_bp_maxiter(bp_cache),
+    edge_sequence=default_bp_edge_sequence(bp_cache),
+    tol=nothing,
+  )
+end
+function default_message_update_alg(bp_cache::BeliefPropagationCache)
+  Algorithm("contract"; normalize=true, sequence_alg="optimal")
+end
+function set_kwargs(alg::Algorithm"contract")
+  normalize = get(alg.kwargs, :normalize, true)
```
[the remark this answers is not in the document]
JoeyT1994: Sure, I went with the first one

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -144,6 +126,36 @@ function incoming_messages(
...
+  for v in vs
+    @preserve_graph bpc[v] = f(bpc[v])
+  end
+  return bpc
+end
+function adapt_messages(to, bpc::AbstractBeliefPropagationCache, args...)
+  map_messages(adapt(to), bpc, args...)
+end
+function adapt_factors(to, bpc::AbstractBeliefPropagationCache, args...)
+  map_factors(adapt(to), bpc, args...)
```
mtfishman, line 150:
```suggestion
  return map_factors(adapt(to), bpc, args...)
```

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -234,35 +246,58 @@ function delete_message(bpc::AbstractBeliefPropagationCache, pe::PartitionEdge)
...
-  return set_message(bpc, edge, updated_message(bpc, edge; kwargs...))
+  return set_message(bpc, edge, updated_message(message_update_alg, bpc, edge; kwargs...))
 end
 
 """
 Do a sequential update of the message tensors on `edges`
 """
-function update(
-  alg::Algorithm,
+function update_one_iteration(
```
mtfishman, line 299: I would vote for just calling this `update_iteration`, I think `iteration` implies it is singular.

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -303,24 +338,18 @@ end
...
-  bpc::AbstractBeliefPropagationCache;
-  edges=default_edge_sequence(alg, bpc),
-  maxiter=default_bp_maxiter(alg, bpc),
-  message_update_kwargs=default_message_update_kwargs(alg, bpc),
-  tol=nothing,
-  verbose=false,
-)
-  compute_error = !isnothing(tol)
-  if isnothing(maxiter)
+function update(alg::Algorithm, bpc::AbstractBeliefPropagationCache; kwargs...)
```
mtfishman, line 341: Same question here, are there still `kwargs` being passed and forwarded here?

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -303,24 +338,18 @@ end
...
     error("You need to specify a number of iterations for BP!")
   end
-  for i in 1:maxiter
+  for i in 1:alg.kwargs.maxiter
     diff = compute_error ? Ref(0.0) : nothing
-    bpc = update(alg, bpc, edges; (update_diff!)=diff, message_update_kwargs...)
-    if compute_error && (diff.x / length(edges)) <= tol
-      if verbose
+    bpc = update_one_iteration(
+      alg, bpc, alg.kwargs.edge_sequence; (update_diff!)=diff, kwargs...
```
mtfishman, line 349: I see that `update_diff!` is still being passed along, that's slightly unfortunate since it forces every implementation of `update_one_iteration`/`update_iteration` to compute that. Can't we compute it directly in this function by comparing `bpc` before and after calling `update_iteration`?

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -234,44 +246,63 @@ function delete_message(bpc::AbstractBeliefPropagationCache, pe::PartitionEdge)
...
+  bpc::AbstractBeliefPropagationCache,
+  edge::PartitionEdge;
+  alg=default_message_update_alg(bpc),
+  kwargs...,
+)
+  return updated_message(set_default_kwargs(Algorithm(alg; kwargs...)), bpc, edge)
+end
+
+function update_message(
+  message_update_alg::Algorithm, bpc::AbstractBeliefPropagationCache, edge::PartitionEdge;
```
mtfishman, line 288:
```suggestion
  message_update_alg::Algorithm, bpc::AbstractBeliefPropagationCache, edge::PartitionEdge
```

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: Yeah I think that is the only one being passed around anymore. It could be computed inside the function, although I think that forces us to have two copies of all of the message tensors (one for this iteration and one for the one before) which is more memory intensive.

I think currently we compute update_diff `message` by `message` as they are updated, which means we should only ever have the current messages and one of the previous messages if the garbage collection is done right

src/caches/abstractbeliefpropagationcache.jl
```diff
@@ -284,15 +315,14 @@ Do parallel updates between groups of edges of all message tensors
 Currently we send the full message tensor data struct to update for each edge_group. But really we only need the
 mts relevant to that group.
 """
-function update(
+function update_iteration(
   alg::Algorithm,
```
mtfishman, line 319:
```suggestion
  alg::Algorithm"bp",
```
and for the other `update_iteration` functions. Does that make sense to you?

mtfishman:
@JoeyT1994 it looks like you need to update the compat entries for ITensorNetworks to v0.14 in `docs/Project.toml` and `examples/Project.toml`.

src/caches/abstractbeliefpropagationcache.jl
mtfishman: Probably in the cases where keeping two copies of the BP cache in memory is too much, it would be better to write parts of it to disk anyway. Relying on the garbage collector could be finicky.

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: Yeah it does, have done that

src/caches/abstractbeliefpropagationcache.jl
JoeyT1994: Yeah that's a good point. I have left it as is for now and made sure the `kwarg` `update_diff` is specified explicitly  and no `kwargs` are passed further than `update_iteration` but in the future going to disk sounds sensible

PR #254, opened 2025-08-22
Preserve message type upon use of `rescale` 

JoeyT1994:
This PR fixes a bug where the `rescale(bpc::BeliefPropagationCache, ...)` function would automatically convert messages to 32-bit floating point precision because of them being normalized by an inverse power which automatically converts everything to 64-bit precision. 

A test is added that checks calling `normalize` now appropriately conserves the precision.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -5,5 +5,5 @@ Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 
 [compat]
 Documenter = "1.10.0"
-ITensorNetworks = "0.14.1"
+ITensorNetworks = "0.14.0"
 Literate = "2.20.1"
--- a/examples/Project.toml
+++ b/examples/Project.toml
@@ -2,4 +2,4 @@
 ITensorNetworks = "2919e153-833c-4bdc-8836-1ea460a35fc7"
 
 [compat]
-ITensorNetworks = "0.14.1"
+ITensorNetworks = "0.14.0"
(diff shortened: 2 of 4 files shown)


docs/Project.toml
```diff
@@ -5,5 +5,5 @@ Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 
 [compat]
 Documenter = "1.10.0"
-ITensorNetworks = "0.14.0"
+ITensorNetworks = "0.14.1"
```
mtfishman, line 8: Just a note that you only need to update the `Project.toml` compat entries in subdirectories for breaking releases.

JoeyT1994:
@mtfishman Yeah I noticed it when trying to do some sampling from 2D PEPS on GPU (which pre-normalizes for stability). Definitely will be good to have it all working for all sort of different number types (as far as I can tell we have done a pretty good job as this was all I could find).

mtfishman:
> @mtfishman Yeah I noticed it when trying to do some sampling from 2D PEPS on GPU (which pre-normalizes for stability). Definitely will be good to have it all working for all sort of different number types (as far as I can tell we have done a pretty good job as this was all I could find).

Something I do with tests now is set them up so they loop over different element types and array types: https://github.com/ITensor/BlockSparseArrays.jl/blob/8cfd7f9ea56a6cfb7c185903298879c1dfdb9e61/test/test_basics.jl#L49-L61, we should start doing that with ITensorNetworks tests.

PR #255, opened 2025-09-24
Changes to the iterator interface

jack-dunham:
This PR attempts to define an interface for the iteration pattern that is similar to that of DifferentialEquations (as described in the Google doc), however in this case it was not as simple. I have described why and what I did below:

## `AbstractNetworkIterator`

The `iterate` function in Julia really describes how one should transition between states, i.e. from A to B. 
For example, when iterating through `SweepIterator`, one must define the new `RegionIterator` for the next sweep. This is something that is suitable to be wrapped in `iterate` as it essentially is the step required to transition from one sweep to the next. What is less obvious is when to perform the region iteration *itself* such that the code
```julia
for _ in sweep_iterator end
```
performs the actual calculation. The region iteration is not really something that is part of "transitioning from sweep A to B" but rather work that is performed *during* a given sweep. This makes it somewhat difficult to reconcile the state of the data (such as the tensors themselves etc) with the state of the iterator.  
Considering what iteration lowers to:
```julia
next = iterate(iter)
while next !== nothing
    (item, state) = next
    # body (this is where any call to a callback function would occur)
    next = iterate(iter, state)
end
```
One has computed and iterated before any callback function. Each callback essentially acts *before* each computation, which is perhaps little awkward, but not necessarily a problem. What is a problem is that the first step is an exception to this; we don't get to execute a callback at step 1 before any computation happens. 

To avoid this problem, and allow the callback to execute *after* the computation I have decide to separate out the "transition" step and the computation step into two functions `increment!` and `compute!` respectively. This also makes it clear which code is responsible for moving the iterator from A to B, and which code is responsible for performing computation while *in state* A (or B etc). We now also make the first call to `increment!` implicit. That is, the iterator should, when initialized, run the necessary code such that it is ready for computation at the first step. In a way, this is as if we have transition from some abstract 0th state to the 1st state. *In the first call to iterate only* the call to `increment!` is then skipped. 

This is the specific interface of `AbstractNetworkIterator`. An `AbstractNetworkIterator` is a *stateful* iterator and can be summarized by the following `iterate` definition:
```julia
function Base.iterate(NI::AbstractNetworkIterator, init=true)
  done(NI) && return nothing
  init || increment!(NI)
  rv = compute!(NI)
  return rv, false
end
```
The function `done` just checks that the iteration is complete. Note, we increment before we compute, therefore we require
```Julia
done(NI::AbstractNetworkIterator) = state(NI) >= length(NI)
```
with a `>=` rather than `>` to avoid over shooting by 1. Simple subtypes of `AbstractNetworkIterator` can use the above method by defining `state` and `Base.length`, however one can instead choose to dispatch on `done` itself for more complicated cases (see `SweepIterator` for an example). 

The code 
```julia
for _ in iter
    callback(iter)
end
```
when `iter::AbstractNetworkIterator` now executes the callback function after the computation step, but before the work has been done to transition to the next state.

### More details

- The `compute!` function be default just returns the iterator itself (and does nothing else). One can choose to return something else to be used in the body of a loop if desired. 
- The function `increment!` has no default implementation on purpose
- See the `PauseAfterIncrement` adapter iterator as one example of how the interface can be used. This is a general version of the proposed adapter that would allow manual iteration over the region.

### Misc

- `SweepIterator` now has a single type parameter `Problem` that can be used for dispatch (useful for default callbacks, see below).
- I've changed some function calls to be explicit calls to constructors to make it clearer when one is constructing rather than getting.
- `RegionIterator` now has an additional field, `sweep` that is constant and simply stores the current sweep the region iteration is part of. This is to avoid constantly passing around this `sweep` keyword when dispensing the sweep iterator in favor of the simpler region iterator. 

### `sweep_solve`

- The function `sweep_solve` is now just a wrapper over the iterators that allows callbacks to be passed (and therefore requires `PauseAfterIncrement`). 
- The callback functions now take the entire `sweep_iterator` as the only positional argument. 
- I've allowed passing in arbitrary `kwargs` for now. 
- The `sweep_callback`  and `region_callback` functions have has `default_` prepended to their function names to distinguish them from the keyword argument (and to make their purpose clear).

## Notes

- I tested the relevant tests in solvers subdirectories but I couldn't get the full test suite to run for some reason. Might be something up with my environment. 
- I will add some unit tests!

### DIFF
--- a/src/solvers/adapters.jl
+++ b/src/solvers/adapters.jl
@@ -30,3 +30,21 @@ iterator which outputs a tuple of the form (current_region, current_region_kwarg
 at each step.
 """
 region_tuples(R::RegionIterator) = TupleRegionIterator(R)
+
+"""
+  struct PauseAfterIncrement{S<:AbstractNetworkIterator}
+
+Iterator wrapper whos `compute!` function simply returns itself, doing nothing in the 
+process. This allows one to manually call a custom `compute!` or insert their own code it in
+the loop body in place of `compute!`.
+"""
+struct PauseAfterIncrement{S<:AbstractNetworkIterator} <: AbstractNetworkIterator
+  parent::S
+end
+
+done(NC::PauseAfterIncrement) = done(NC.parent)
+state(NC::PauseAfterIncrement) = state(NC.parent)
+increment!(NC::PauseAfterIncrement) = increment!(NC.parent)
+compute!(NC::PauseAfterIncrement) = NC
+
+PauseAfterIncrement(NC::PauseAfterIncrement) = NC
--- a/src/solvers/fitting.jl
+++ b/src/solvers/fitting.jl
@@ -79,7 +79,7 @@ function fit_tensornetwork(
   insert_kwargs = (; insert_kwargs..., normalize, set_orthogonal_region=false)
   common_sweep_kwargs = (; nsites, outputlevel, update_kwargs, insert_kwargs)
   kwargs_array = [(; common_sweep_kwargs..., sweep=s) for s in 1:nsweeps]
-  sweep_iter = sweep_iterator(init_prob, kwargs_array)
+  sweep_iter = SweepIterator(init_prob, kwargs_array)
   converged_prob = sweep_solve(sweep_iter; outputlevel, kws...)
   return rename_vertices(inv_vertex_map(overlap_network), ket(converged_prob))
 end
--- a/src/solvers/iterators.jl
+++ b/src/solvers/iterators.jl
@@ -1,45 +1,4 @@
 #
-# SweepIterator
-#
-
-mutable struct SweepIterator
-  sweep_kws
-  region_iter
-  which_sweep::Int
-end
-
-problem(S::SweepIterator) = problem(S.region_iter)
-
-Base.length(S::SweepIterator) = length(S.sweep_kws)
-
-function Base.iterate(S::SweepIterator, which=nothing)
-  if isnothing(which)
-    sweep_kws_state = iterate(S.sweep_kws)
-  else
-    sweep_kws_state = iterate(S.sweep_kws, which)
-  end
-  isnothing(sweep_kws_state) && return nothing
-  current_sweep_kws, next = sweep_kws_state
-
-  if !isnothing(which)
-    S.region_iter = region_iterator(
-      problem(S.region_iter); sweep=S.which_sweep, current_sweep_kws...
-    )
-  end
-  S.which_sweep += 1
-  return S.region_iter, next
-end
-
-function sweep_iterator(problem, sweep_kws)
-  region_iter = region_iterator(problem; sweep=1, first(sweep_kws)...)
-  return SweepIterator(sweep_kws, region_iter, 1)
-end
-
-function sweep_iterator(problem, nsweeps::Integer; sweep_kws...)
-  return sweep_iterator(problem, Iterators.repeated(sweep_kws, nsweeps))
-end
-
-#
 # RegionIterator
 #
 
@@ -54,10 +13,14 @@ current_region_plan(R::RegionIterator) = R.region_plan[R.which_region]
 current_region(R::RegionIterator) = current_region_plan(R)[1]
 region_kwargs(R::RegionIterator) = current_region_plan(R)[2]
 function previous_region(R::RegionIterator)
-  R.which_region==1 ? nothing : R.region_plan[R.which_region - 1][1]
+  return R.which_region == 1 ? nothing : R.region_plan[R.which_region - 1][1]
 end
 function next_region(R::RegionIterator)
-  R.which_region==length(R.region_plan) ? nothing : R.region_plan[R.which_region + 1][1]
+  return if R.which_region == length(R.region_plan)
+    nothing
+  else
+    R.region_plan[R.which_region + 1][1]
+  end
 end
 is_last_region(R::RegionIterator) = isnothing(next_region(R))
 
@@ -97,4 +60,45 @@ end
 
 function region_plan(problem; kws...)
   return euler_sweep(state(problem); kws...)
+end
+
+#
+# SweepIterator
+#
+
+mutable struct SweepIterator{Problem}
+  sweep_kws
+  region_iter::RegionIterator{Problem}
+  which_sweep::Int
+end
+
+problem(S::SweepIterator) = problem(S.region_iter)
+
+Base.length(S::SweepIterator) = length(S.sweep_kws)
+
+function Base.iterate(S::SweepIterator, which=nothing)
+  if isnothing(which)
+    sweep_kws_state = iterate(S.sweep_kws)
+  else
+    sweep_kws_state = iterate(S.sweep_kws, which)
+  end
+  isnothing(sweep_kws_state) && return nothing
+  current_sweep_kws, next = sweep_kws_state
+
+  if !isnothing(which)
+    S.region_iter = region_iterator(
+      problem(S.region_iter); sweep=S.which_sweep, current_sweep_kws...
+    )
+  end
+  S.which_sweep += 1
+  return S.region_iter, next
+end
+
+function sweep_iterator(problem, sweep_kws)
+  region_iter = region_iterator(problem; sweep=1, first(sweep_kws)...)
+  return SweepIterator(sweep_kws, region_iter, 1)
+end
+
+function sweep_iterator(problem, nsweeps::Integer; sweep_kws...)
+  return sweep_iterator(problem, Iterators.repeated(sweep_kws, nsweeps))
 end
--- a/src/solvers/iterators.jl
+++ b/src/solvers/iterators.jl
@@ -1,61 +1,113 @@
+"""
+  abstract type AbstractNetworkIterator
+
+A stateful iterator with two states: `increment!` and `compute!`. Each iteration begins
+with a call to `increment!` before executing `compute!`, however the initial call to
+`iterate` skips the `increment!` call as it is assumed the iterator is initalized such that 
+this call is implict. Termination of the iterator is controlled by the function `done`.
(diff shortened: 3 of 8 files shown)


src/solvers/iterators.jl
```diff
@@ -1,100 +1,158 @@
...
+#
+# SweepIterator
+#
+
+mutable struct SweepIterator{Problem} <: AbstractNetworkIterator
+  sweep_kws
+  region_iter::RegionIterator{Problem}
+  which_sweep::Int
+  function SweepIterator(problem, sweep_kws)
+    sweep_kws = Iterators.Stateful(sweep_kws)
```
mtfishman, line 126:
```suggestion
    stateful_sweep_kws = Iterators.Stateful(sweep_kws)
```
I would find that a bit clearer (i.e. it is easier to keep track that we are now using the stateful version of `sweep_kws`).

src/solvers/iterators.jl
```diff
@@ -1,100 +1,158 @@
...
 function region_plan(problem; kws...)
   return euler_sweep(state(problem); kws...)
 end
+
+#
+# SweepIterator
+#
+
+mutable struct SweepIterator{Problem} <: AbstractNetworkIterator
+  sweep_kws
```
mtfishman, line 122: Should we parametrize `SweepIterator` by the type of `sweep_kws`? Or are we expecting it could be modified such that it changes type (though if that is the case we could make it explicit by setting the type parameter to `Any`)?

src/solvers/sweep_solve.jl
```diff
@@ -1,40 +1,30 @@
...
-        region_iterator=region_iter,
-        region,
-        region_kwargs,
-        sweep,
-        kwargs...,
-      )
+  # Don't compute the region iteration automatically as we wish to insert a callback.
+  for _ in PauseAfterIncrement(sweep_iterator)
+    for _ in region_iterator(sweep_iterator)
+      region_callback(sweep_iterator; outputlevel=outputlevel)
```
mtfishman, line 19: Passing `outputlevel` is a bit funny to me (I think it is there for historical reasons). I would think that it would be easy enough to pass keyword arguments as part of the callback, so I think it can be removed.

mtfishman: Otherwise, we could have users pass `sweep_callback_kwargs` and `region_callback_kwargs` NamedTuples.

So basically the alternatives would be:
```julia
sweep_solve(sweep_iterator; region_callback=(x -> default_region_callback(x; outputlevel=1))
```
or:
```julia
sweep_solve(sweep_iterator; region_callback_kwargs=(; outputlevel=1))
```
I'm ok with either one, though I'd bias towards the first one for now since we can always add the second one later, for the sake of keeping this level of the code simpler.

src/solvers/sweep_solve.jl
```diff
@@ -1,40 +1,30 @@
...
-    )
+    sweep_callback(sweep_iterator; outputlevel=outputlevel)
   end
   return problem(sweep_iterator)
 end
+
+# I suspect that `sweep_callback` is the more commonly used callback, so allow this to
+# be set using the `do` syntax.
+function sweep_solve(sweep_callback, sweep_iterator; kwargs...)
+  return sweep_solve(sweep_iterator; sweep_callback=sweep_callback, kwargs...)
```
mtfishman, line 29:
```suggestion
  return sweep_solve(sweep_iterator; sweep_callback, kwargs...)
```

src/solvers/sweep_solve.jl
```diff
@@ -1,40 +1,30 @@
...
-  if outputlevel >= 1
-    println("Done with sweep $sweep/$nsweeps")
-  end
+function default_region_callback(sweep_iterator; kwargs...)
+  return sweep_iterator
 end
-
+function default_sweep_callback(sweep_iterator; kwargs...)
+  return sweep_iterator
+end
```
mtfishman, line 7: Maybe it's better to remove the `kwargs...` to make sure we don't accidentally pass unsupported keyword arguments.

src/solvers/adapters.jl
```diff
@@ -30,3 +30,21 @@ iterator which outputs a tuple of the form (current_region, current_region_kwarg
...
+struct PauseAfterIncrement{S<:AbstractNetworkIterator} <: AbstractNetworkIterator
+  parent::S
+end
+
+done(NC::PauseAfterIncrement) = done(NC.parent)
+state(NC::PauseAfterIncrement) = state(NC.parent)
+increment!(NC::PauseAfterIncrement) = increment!(NC.parent)
+compute!(NC::PauseAfterIncrement) = NC
+
+PauseAfterIncrement(NC::PauseAfterIncrement) = NC
```
mtfishman, line 50: I small recommendation that I would prefer a variable name besides `NC`, maybe we could just use `iterator`.

src/solvers/sweep_solve.jl
jack-dunham: I don't agree with passing `outputlevel` as a kwarg, I agree its a bit weird, and I left it in for historical reasons as you mention.

IMO, options like `outputlevel` can be defined as part of a closure in a wrapper function and then exposed to a user, if we (or another user) wishes to do that. We should not assume a notion of `outputlevel` for all callbacks.

src/solvers/iterators.jl
```diff
@@ -1,100 +1,158 @@
...
+
+region_iterator(S::SweepIterator) = S.region_iter
+problem(S::SweepIterator) = problem(region_iterator(S))
+
+state(SR::SweepIterator) = SR.which_sweep
+Base.length(S::SweepIterator) = length(S.sweep_kws)
+function increment!(SR::SweepIterator)
+  SR.which_sweep += 1
+  sweep_kwargs, _ = Iterators.peel(SR.sweep_kws)
+  SR.region_iter = RegionIterator(problem(SR); sweep=state(SR), sweep_kwargs...)
```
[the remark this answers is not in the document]
mtfishman: Also, what does `SR` stand for? Generally I prefer lower case variable names, and also more descriptive ones, say `sweep_iter`.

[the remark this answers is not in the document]
jack-dunham: I think `SR` was an abbreviation of `SweepRegionIterator` which was a previous name for something I was using previously...

On the first point, I feel like `update_region_iterator` is not really descriptive of what the function as implemented there does. It does not update an existing `RegionIterator`, it creates a new one via the constructor. I do like however the following: 
```julia
function update_region_iterator!(iterator::SweepIterator; kwargs...)
    iterator.region_iter = new_region_iterator(iterator.region_iter; kwargs...)
    return iterator
end
```
where `new_region_iterator` is a renaming of the method you describe. Let me know if you agree.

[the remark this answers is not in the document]
mtfishman: > It does not update an existing `RegionIterator`, it creates a new one via the constructor. I do like however the following:

I suppose that's a matter of perspective, i.e. we could think of it as a fancy kind of ["setter" method](https://github.com/JuliaObjects/Accessors.jl?tab=readme-ov-file#usage) that keeps some properties and sets other ones, which for immutable types has to create a new object by definition, but I'm ok with `new_region_iterator`.

jack-dunham:
Some comments:

Bearing in mind the discussions about avoiding the proliferation of methods that don't do a lot (i.e. just pass data from structs as arguments), I have elected to take the opposite approach regarding the interface defined for `default_kwargs`. 
This is, however, for good reason; most of the defaults don't actually use the contents `problem` (or `region_iter` for that matter), i.e. it is just used for dispatch, therefore it makes sense to have users dispatch on the type itself as one can then access these defaults without constructing an instance of said type. 

If one needs some of the runtime data from `problem` or whatever, they can still just define a method dispatching on `AbstractProblem` like normal as now it does not even make sense to have this default accessible without an instance of this type as the default depends on values at runtime... 

For that reason, I think supporting both makes sense.  One can define a method in two ways: 
```Julia
default_kwargs(f, ::MyProblem)) # value domain
default_kwargs(f, ::Type{<:MyProblem)) # type domain
```
If no method in the value domain is set, then it falls back to the one in the type domain. I have (briefly) described the interface in the docstring of `default_kwargs`.

Some more things to discuss:

1. Which functions should support this interface
2. If we should support defaults that are not tied to a function 
3. Avoiding type piracy

On the last point, we should be strict about asking users to only define `default_kwargs` methods on their own types and functions. It would therefore be sensible to have default methods defined by us to have signatures `default_kwargs(f, ::AbstractProblem)` as then one cannot redefine this method without getting a warning (I think/hope??). We then need a system for overriding these defaults which can be done with a `ScopedValue`.

mtfishman:
That sounds reasonable to me.

emstoudenmire:
This PR has a green light from me, after remaining review comments are resolved or answered.

src/solvers/fitting.jl
```diff
@@ -98,14 +109,11 @@ end
 #end
 
 function ITensors.apply(
-  A::ITensorNetwork,
-  x::ITensorNetwork;
-  maxdim=default_maxdim(),
-  cutoff=default_cutoff(),
-  kwargs...,
+  A::ITensorNetwork, x::ITensorNetwork; maxdim=typemax(Int), cutoff=0.0, sweep_kwargs...
```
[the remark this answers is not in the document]
jack-dunham: See https://github.com/ITensor/ITensorNetworks.jl/pull/255#issuecomment-3406768513

src/ITensorNetworks.jl
```diff
@@ -45,6 +45,7 @@ include("treetensornetworks/projttns/projttn.jl")
 include("treetensornetworks/projttns/projttnsum.jl")
 include("treetensornetworks/projttns/projouterprodttn.jl")
 
+include("solvers/default_kwargs.jl")
```
mtfishman:
```suggestion
```

PR #262, opened 2025-10-28
Fix bug with length 1 `AbstractNetworkIterator` types not executing their only `compute!` call. 

jack-dunham:
# Description

Checking of `islaststep` is now skipped for the first iteration. This fixes a bug where  length 1 iterators don't iterate at all and thus don't execute their only call to `compute!`. 
Skipping `islaststep` check for the initial iterations makes sense as the interface of `AbstractNetworkIterator` assumes an implicit "move to first state" at construction. 

This also means that constructing a `AbstractNetworkIterator` is basically undefined as there is in that case there is no first state to move to. As such, constructing `SweepIterator` and `RegionIterator` now throws an error if if empty iterators are passed. 

I have added some tests for these edge cases; all other tests pass.


jack-dunham:
Closing this as these changes to be included in DataGraphs and NamedGraphs upgrade patch.

PR #267, opened 2025-12-09
Improve iterators and restore outputlevel

emstoudenmire:
This PR introduces a handful of improvements to the updated solvers codes:
* Add unit tests for SweepIterator
* Fix issues with SweepIterator, including the case of 1 sweep (see #262) and fixing `length` for SweepIterator
* Support for `outputlevel` added back to `eigsolve` and `applyexp`
* The `nsites` keyword is no longer required to construct a SweepIterator (was accidentally required by not providing a default in `region_plans/euler_plans.jl`.

This PR supersedes #262 by including all changes from it.


mtfishman:
I'll close this since the most pressing issue (the `outputlevel` issue) was addressed in #268, and we can discuss and work on the other parts separately.

PR #317, opened 2026-04-14
Upgrade ITensorNetworks to use DataGraphs v0.4.0 and NamedGraphs v0.11.0.

jack-dunham:
This package now used the proper `AbstractDataGraph` interface from DataGraphs, and the `similar_graph` interface from `NamedGraphs`.


mtfishman:
@jack-dunham can you look into the version, format, and docs failures?

jack-dunham:
Yes, I am just waiting for the tests to pass locally.

mtfishman:
(The docs failures are a bit odd, they may be related to https://github.com/ITensor/ITensors.jl/pull/1729 so I can look into that in parallel as well, but please do take a look.)

src/treetensornetworks/treetensornetwork.jl
```diff
@@ -76,16 +77,20 @@ end
...
-    ITensorNetwork(tn::TreeTensorNetwork) -> ITensorNetwork
+    itensornetwork(tn::TreeTensorNetwork) -> ITensorNetwork
 
 Convert a `TreeTensorNetwork` to a plain `ITensorNetwork`, discarding orthogonality
 metadata. The returned network shares the same underlying tensor data.
 
 See also: [`TreeTensorNetwork`](@ref), [`ttn`](@ref).
 """
-ITensorNetwork(tn::TTN) = getfield(tn, :tensornetwork)
+itensornetwork(tn::TTN) = getfield(tn, :tensornetwork)
```
mtfishman: Is this the source of the docs issue? It seems like we should keep `ITensorNetwork(tn::TTN)` around, that's a reasonable conversion to do. Honestly I think `itensornetwork(tn::TTN)` is a bit confused for this, maybe we could define `ITensorNetwork(tn::TTN)` as a copy constructor (i.e. `copy(tn.tensornetwork)`) and then just use field access `tn.tensornetwork` where you are now using `itensornetwork(tn)`.

mtfishman: But also we should add a direct test for it so it doesn't get caught just in the docs.

jack-dunham: Yes this is it. The reason I changed this is that it seemed strange to me that one calls a constructor to access a field (I was interpreting that bit of code less like constructing an `ITensorNetwork` and more like returning some data).

mtfishman: I agree, it should have been copying, not doing direct field access. Really we should have a more generic constructor like `ITensorNetwork(tn::AbstractITensorNetwork) = ITensorNetwork(map(v -> v => tn[v], vertices(tn)))` and then no specialized `ITensorNetwork(tn::TTN)`. We could do that here but maybe that is better left for future PRs where we simplify/update the ITensorNetwork constructors more broadly, up to you.

mtfishman: The issue with `ITensorNetwork(map(v -> v => tn[v], vertices(tn)))` right now is that it in general will be pretty slow since we don't have the cached index to vertex looked that we discussed, so we kind of need a specialized `ITensorNetwork(tn::TTN) = copy(tn.tensornetwork)` implementation right now anyway.

jack-dunham: Let's think about this in a future PR. I'll fix the copy method and add a test.

src/treetensornetworks/abstracttreetensornetwork.jl
```diff
@@ -18,7 +18,7 @@ end
 # Field access
 #
 
-ITensorNetwork(tn::AbstractTTN) = error("Not implemented")
+itensornetwork(tn::AbstractTTN) = error("Not implemented")
```
mtfishman: Should this be reverted?

src/indsnetwork.jl
```diff
@@ -309,5 +312,5 @@ end
 # based on `ITensorVisualizationCore`.).
 using ITensors.ITensorVisualizationCore: ITensorVisualizationCore, visualize
 function ITensorVisualizationCore.visualize(is::IndsNetwork, args...; kwargs...)
-    return visualize(ITensorNetwork(is), args...; kwargs...)
+    return visualize(tensornetwork(is), args...; kwargs...)
```
mtfishman: Same thing here, this looks like a mistake.

mtfishman:
Thanks Jack, glad to see this getting in!

PR #346, opened 2026-04-27
Document interface of library

emstoudenmire:
This PR adds documentation of which methods we may consider "interface", and thus should keep and add docstrings for, versus deprecated methods we may want to delete. Methods not in the two files `interface_methods.md` and `deprecated_methods.md` are considered internal, developer-facing methods.


docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+* From a vector of `ITensor`s, with vertex labels auto-assigned to `eachindex(ts)`.
+  Edges are inferred from shared indices (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(ts::AbstractVector{ITensor})
+  ```
+
+* From a single `ITensor`. Wraps the tensor in a single-vertex network (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(t::ITensor)
+  ```
```
mtfishman: This should be moved to deprecated, this was a bad constructor in retrospect.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+
+* From `IndsNetwork`. Initializes ITensors with `undef` storage on each vertex
+  of the `IndsNetwork` with the corresponding indices (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(eltype::Type, undef::UndefInitializer, is::IndsNetwork; kwargs...)
+  ITensorNetwork(eltype::Type, is::IndsNetwork; kwargs...)
+  ITensorNetwork(undef::UndefInitializer, is::IndsNetwork; kwargs...)
+  ITensorNetwork(is::IndsNetwork; kwargs...)
+  ```
```
mtfishman: IndsNetwork may be in "purgatory" right now. With the new plans for the constructors, it becomes less important, however there may be some use cases for it beyond constructors where it is useful. Basically it has been used as a stand-in for the tensors that stores the information about the space of the tensors but without the tensors themselves, but it may be more explicit to pass around NamedGraphs and dictionaries of indices (which is basically all it is, just wrapped up in a single type). Maybe we should have three files: depracted_methods.md, interface_methods.md, and purgatory.md (or a less cute name if you can think of one).

docs/src/deprecated_methods.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Local Operations on ITensorNetworks
+
+* Combine (fuse) every link index of a tensor network, or a chosen set of edges, into
+  a single index per edge using `combiner` tensors. (`abstractitensornetwork.jl`):
+  ```julia
+  linkinds_combiners(tn::AbstractITensorNetwork; edges = edges(tn))
+  combine_linkinds(tn::AbstractITensorNetwork, combiners)
+  combine_linkinds(tn::AbstractITensorNetwork; edges = edges(tn))
+  ```
```
mtfishman, line 70: I think this kind of functionality is useful but maybe this isn't the right interface for it, so I'm not completely convinced this should be deprecated but maybe rethought. For example, it may be better to fold this into a more general interface `gauge_transform(f, tn::AbstractITensorNetwork)` that applies a gauge transformation `X_e X_e^{-1}` on each edge `e`. So maybe make a comment about that and move it to "purgatory.md"?

docs/src/deprecated_methods.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Global Operations on ITensorNetworks
+
+* Scale tensors at chosen vertices by per-vertex weights, either out-of-place or in-place (`abstractitensornetwork.jl`):
+  ```julia
+  scale(tn::AbstractITensorNetwork, vertices_weights::Dictionary; kwargs...)
+  scale(weight_function::Function, tn; kwargs...)
+  scale!(tn::AbstractITensorNetwork, vertices_weights::Dictionary)
+  scale!(weight_function::Function, tn::AbstractITensorNetwork; kwargs...)
+  ```
```
mtfishman: Not sure if this should be depracated, but maybe the details of the interface need to be rethought (or it should be folded into a more general function `map_tensors(f, tn::AbstractITensorNetwork)`). So again, maybe a candidate for "purgatory.md" with some comments with ideas for improvements.

docs/src/deprecated_methods.md
```diff
@@ -0,0 +1,88 @@
...
+  ```julia
+  mpo(os::OpSum, external_inds::Vector; kws...)
+  mpo(os::OpSum, s::IndsNetwork; kws...)
+  ttn(o::Op, s::IndsNetwork; kws...)
+  ttn(o::Scaled{C, Op}, s::IndsNetwork; kws...)
+  ttn(o::Sum{Op}, s::IndsNetwork; kws...)
+  ttn(o::Prod{Op}, s::IndsNetwork; kws...)
+  ttn(o::Scaled{C, Prod{Op}}, s::IndsNetwork; kws...)
+  ttn(o::Sum{Scaled{C, Op}}, s::IndsNetwork; kws...)
+  ```
```
mtfishman, line 131: I guess we'll have to think about the alternatives to these if we remove them. Maybe it would be something like `ttn(opsum::OpSum,  graph::NamedGraph, siteinds:Dict{V, <: Index})`, i.e. you specify the graph and site indices? Though I guess we could make it more general, i.e. `ITensorNetwork(opsum::OpSum,  graph::NamedGraph, siteinds:Dict{V, <: Index})`, and error if the graph isn't a tree.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+
+Recommended methods for building applications on top of ITensorNetworks.
+
+## ITensorNetwork Constructors
+
+These ITensorNetwork constructor interfaces are foundational to other constructors:
+
+* Default constructor and copy constructor (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork{V}()
```
mtfishman: Probably this is a strange constructor to keep around, and users can do `ITensorNetwork(Dict{V}())` if they specifically want an empty tensor network.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+  edge_sequence(::Algorithm"forest_cover", g; root_vertex = GraphsExtensions.default_root_vertex)
+  ```
+
+* Parallel sequence: each edge (and its reverse) as its own one-edge group, suitable
+  for parallel BP updates (`edge_sequences.jl`):
+  ```julia
+  edge_sequence(::Algorithm"parallel", g)
+  ```
+
+#### AbstractBeliefPropagationCache
```
mtfishman: I would put this in a new "developer.md" file, this isn't something users should generally be using directly.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+  ```
+
+* Return a copy of the cache with messages on `pes` rescaled to unit norm and
+  symmetrized so the bond region scalar is 1 (`caches/beliefpropagationcache.jl`):
+  ```julia
+  rescale_messages(bp_cache::BeliefPropagationCache, pes)
+  ```
+
+
+## ProjTTN System
```
mtfishman: I would put this in "purgatory.md", the plan is to replace it with some combination of a quadratic form and BP cache.

mtfishman:
@emstoudenmire can you bump the version, fix the docs build, and format just so we can get tests passing?

Also, I noticed you are just including the function signatures in the docs. For functions that have docstrings, can you insert the docstring instead of just putting the function signature?

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+* Indices common to the ITensors on the vertices connected by the edge (`abstractitensornetwork.jl`):
+  ```julia
+  commoninds(tn::AbstractITensorNetwork, edge)
+  linkinds(tn::AbstractITensorNetwork, edge)
+  ```
+
+* Collection of tensors neighboring the given vertex (`abstractitensornetwork.jl`):
+  ```julia
+  neighbor_tensors(tn::AbstractITensorNetwork, vertex)
+  ```
```
mtfishman: Probably this doesn't need to be its own function, and can just be done by the user with a combination of `neighbors` and `tn[v]`. So, deprecated.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+
+* Indices on the source tensor of `edge` that are not shared with the destination tensor.
+  For `edge = (v1 => v2)`, returns every index of `tn[v1]` except the bond(s) to `tn[v2]`, that is,
+  `v1`'s site index plus its link indices to all of `v1`'s other neighbors. Useful for QR/SVD
+  across that bond: these are the indices you keep on the `v1` side (`abstractitensornetwork.jl`):
+  ```julia
+  uniqueinds(tn::AbstractITensorNetwork, edge::AbstractEdge)
+  uniqueinds(tn::AbstractITensorNetwork, edge::Pair)
+  # Alias for `uniqueinds`.
+  siteinds(tn::AbstractITensorNetwork, edge)
```
mtfishman: That doesn't sound right... `uniqueinds(n::AbstractITensorNetwork, edge)` would include the non-shared linkinds as well, so `siteinds` seems like the wrong name for that. Either the description here is wrong or the code definition is bad.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+
+* Tags on the link index (or indices) associated with `edge` (`abstractitensornetwork.jl`):
+  ```julia
+  tags(tn::AbstractITensorNetwork, edge)
+  ```
+
+* Iterate over the tensors at the given vertices, default all vertices (`abstractitensornetwork.jl`):
+  ```julia
+  eachtensor(tn::AbstractITensorNetwork, vertices = vertices(tn))
+  ```
```
mtfishman: This seems like an unnecessary function, deprecate.

docs/src/deprecated_methods.md
mtfishman: Ah, I remember why we don't just use `map_tensors`: for scaling we know the graph is preserved. That won't be a concern when we add a reverse index map, but scaling tensors may be a standard enough operation that it deserves its own function. However, it should probably be renamed to `scale_tensors`.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+  addtags(tn::AbstractITensorNetwork, args...; kwargs...)
+  removetags(tn::AbstractITensorNetwork, args...; kwargs...)
+  replacetags(tn::AbstractITensorNetwork, args...; kwargs...)
+  settags(tn::AbstractITensorNetwork, args...; kwargs...)
+  swaptags(tn::AbstractITensorNetwork, args...; kwargs...)
+  sim(tn::AbstractITensorNetwork, args...; kwargs...)
+  ```
+
+
+## Form Networks
```
mtfishman: This should be moved to developer.md.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+  ```
+
+* Apply the corresponding vertex map to a single vertex `v` (`formnetworks/abstractformnetwork.jl`):
+  ```julia
+  operator_vertex(f::AbstractFormNetwork, v)
+  bra_vertex(f::AbstractFormNetwork, v)
+  ket_vertex(f::AbstractFormNetwork, v)
+  ```
+
+#### LinearFormNetwork
```
mtfishman: developer.md.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+  copy(lf::LinearFormNetwork)
+  ```
+
+* Replace the ket-side tensor at the original vertex `original_ket_state_vertex` with
+  `ket_state` (graph-preserving update) (`formnetworks/linearformnetwork.jl`):
+  ```julia
+  update(lf::LinearFormNetwork, original_ket_state_vertex, ket_state::ITensor)
+  ```
+
+#### BilinearFormNetwork
```
mtfishman: developer.md.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,1328 @@
...
+  ttn(os::OpSum,sites::IndsNetwork; kws...)
+  ```
+
+* From `OpSum`, assuming path graph (`opsum_to_ttn.jl`):
+  ```julia
+  mpo(os::OpSum, external_inds::Vector; kws...)
+  mpo(os::OpSum, s::IndsNetwork; kws...)
+  ```
+
+#### AbstractTreeTensorNetwork Type
```
mtfishman: purgatory.md.

docs/src/interface_methods.md
emstoudenmire: I'm making another file called "Experimental Methods" which can be the "purgatory" file. I'll move IndsNetwork and other experimental designs there. (By experimental I mean sort of like "past experiments which may or may not have panned out".)

docs/src/deprecated_methods.md
emstoudenmire: I added the core of this comment to the blurb around these methods, now in the Experimental Methods file.

docs/src/deprecated_methods.md
emstoudenmire: Ok great, moving this one to interface with a comment.

docs/src/deprecated_methods.md
emstoudenmire: It was my bad to include the `mpo` ones here. I had put them in "interface" actually, just accidentally duplicated them here. 

Were your comments above about a more general constructor that would take the place of the `mpo` ones?

The ones really intended to be here in Deprecated are the ones which take other Op-adjacent data like `Scaled{C,Op}`, `Prod{Op}`, etc. My guess was we might want those but later on through Lukas's system so maybe better not to have them here? They are little one-liners though so not causing any big issues. What do you think of these?

mtfishman: I see, I thought you were including all `mpo` and `ttn` constructors. That sounds reasonable to get rid of `Scaled{C,Op}`, `Prod{Op}`, etc. for now (non-OpSum ones).

I was indeed suggesting getting rid of `ttn` and `mpo` entirely and just focusing on a more general `ITensorNetwork` constructor that just restricts to trees with a runtime check.

mtfishman: Of course, that would be a longer term change which assumes we are getting rid of the TTN type, and since the current DMRG/TDVP code still expects a TTN then we could replace `mpo` and `ttn` with using the TTN type directly, i.e. remove `mpo` and `ttn` functions in favor of just making a TTN constructor that accepts an OpSum.

docs/src/interface_methods.md
emstoudenmire: Yes there are two of them which are literally one-line aliases for `uniqueinds`. So I'll move them (with a note about where they are) to Deprecated.

docs/src/deprecated_methods.md
emstoudenmire: I think the above would be a good plan. One way we could go would be to start replacing tree tensor network methods with `ITensorNetwork` ones (just restricted to trees with a runtime check) and then after some rounds of this see how much of a "footprint" is left by the actual `TreeTensorNetwork` type. It might not be that much after a handful of key changes. 

My guess is probably the largest footprint is through parts of the code, like solvers, which rely on the gauging capabilities of tree tensor networks but also of course that's something that could be switched over to ITensorNetworks with a runtime tree check. Also I remember you had some thoughts about gauging and caches that sounded more deep.. we should discuss more.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,380 @@
...
+
+* Indices on the source tensor of `edge` that are not shared with the destination tensor.
+  For `edge = (v1 => v2)`, returns every index of `tn[v1]` except the bond(s) to `tn[v2]`, that is,
+  `v1`'s site index plus its link indices to all of `v1`'s other neighbors. Useful for QR/SVD
+  across that bond: these are the indices you keep on the `v1` side (`abstractitensornetwork.jl`):
+  ```julia
+  uniqueinds(tn::AbstractITensorNetwork, edge::AbstractEdge)
+  uniqueinds(tn::AbstractITensorNetwork, edge::Pair)
+  # Alias for `uniqueinds`.
+  siteinds(tn::AbstractITensorNetwork, edge)
```
mtfishman: Still needs moving to deprecated.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,380 @@
...
+
+* Collect all site indices (per-vertex) of a network as a flat vector (`abstractitensornetwork.jl`):
+  ```julia
+  flatten_siteinds(tn::AbstractITensorNetwork)
+  ```
+
+* Collect all link indices (per-edge) of a network as a flat vector (`abstractitensornetwork.jl`):
+  ```julia
+  flatten_linkinds(tn::AbstractITensorNetwork)
+  ```
```
mtfishman: Maybe these should be moved to experimental, they are pretty specialized.

docs/src/interface_methods.md
```diff
@@ -0,0 +1,380 @@
...
+  contract(tn::AbstractITensorNetwork, edge::Pair; kws...)
+  ```
+
+* "Split" an edge index by applying a map to each copy of it on the adjacent ITensors.
+  By default the `dst(edge)` copy is primed and the `src(edge)` copy is unchanged (`abstractitensornetwork.jl`):
+  ```julia
+  split_index(tn::AbstractITensorNetwork, edges_to_split; 
+              src_ind_map::Function = identity,
+              dst_ind_map::Function = prime)
+  ```
```
mtfishman: Probably should be moved to experimental.

docs/src/interface_methods.md
emstoudenmire: Just the `siteinds` alias, correct? I.e. keep `uniqueinds` in interface?

mtfishman: Yeah.

emstoudenmire: I could also see getting rid of the `uniqueinds` interface that takes an `edge` and just keeping the version of `uniqueinds` that takes a `vertex`.

mtfishman: I think the motivation for it is that it is convenient for edge factorizations, i.e. `factorize(tn, e) = (domain = uniqueinds(tn, e); factorize(tn[src(e)] * tn[dst(e)], domain); ...)`. But also I've been leaning towards having fewer helper functions that can be written in a few lines anyway, I think in ITensorMPS (and now in ITensorNetworks) we just have too many functions. In fact in NamedDimsArrays and ITensorBase I think I got rid of uniqueinds, commoninds, etc. entirely and instead you just call `setdiff(inds(a1), inds(a2))` which is nearly as short, pretty clear, and doesn't require us supporting those functions... So I think that was a long-winded way of saying I agree with you and probably it is better to remove as much as we can and see what we want to bring back.

emstoudenmire: Ok I was inspired by that point about set functions and just moved all of the uniqueinds and commondinds methods to Deprecated. If you want to do something more conservative happen to move some back, but could be interesting to do everything through set functions.

docs/src/experimental_methods.md
```diff
@@ -0,0 +1,575 @@
...
+  ```
+
+* Construct an `ITensorNetwork` from an `IndsNetwork`. Initializes ITensors with `undef` storage on each vertex
+  of the `IndsNetwork` with the corresponding indices (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(eltype::Type, undef::UndefInitializer, is::IndsNetwork; kwargs...)
+  ITensorNetwork(eltype::Type, is::IndsNetwork; kwargs...)
+  ITensorNetwork(undef::UndefInitializer, is::IndsNetwork; kwargs...)
+  ITensorNetwork(is::IndsNetwork; kwargs...)
+  ```
```
mtfishman: I think these should be moved to deprecated.

docs/src/interface_methods.md
mtfishman: Sounds good to me, that's what we are trying out in ITensorNetworksNext and I think it is better that way.

docs/src/experimental_methods.md
emstoudenmire: Done

docs/src/interface_methods.md
```diff
@@ -0,0 +1,338 @@
...
+* From a collection of ITensorNetworks. Merges (Kronecker or tensor product) of input networks (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(itns::Vector{ITensorNetwork})
+  ```
+
+* From a vector of `ITensor`s, with vertex labels auto-assigned to `eachindex(ts)`.
+  Edges are inferred from shared indices (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(ts::AbstractVector{ITensor})
+  ```
```
mtfishman: I think we can compress this down:
```julia
# `keys(tensors)` are vertices, `values(tensors)` are tensors on those vertices
ITensorNetwork(tensors)
ITensorNetwork{V}(tensors)

# Collection of vertices, collection of tensors
ITensorNetwork(vertices, tensors)
ITensorNetwork{V}(vertices, tensors)
```
That actually encompasses all of them (potentially even `ITensorNetwork{V}(tn::ITensorNetwork)`, since `keys(tn)` _should_ be the vertices and `values(tn)` _should_ be the tensors, though we'd have to double check if that would work in practice).

Also, I don't particularly like the `ITensorNetwork(itns::Vector{ITensorNetwork})` since it doesn't fit into that generic framework very well, I think we should move that to "deprecated".

Note that this doesn't necessarily fully reflect the current state of the code: the constructors are not that generic, some of the ones I wrote above may not exist yet, etc. However, I think for these docs we can do some mixture of summarizing the current code and resummarizing in a way we _want_ the code to be (though please write a comment when these docs diverge from the current code, indicating that they reflect plans for the code).

emstoudenmire: Interesting point. Just made this change and I like the idea of the docs being slightly "aspirational" so we can keep working the code toward the docs.

About the last constructor above, though, the `ITensorNetwork(ts::AbstractVector{ITensor})` one. That doesn't seem to quite fit either description. I guess keep it as a separate item?

emstoudenmire:
By the way, the interface page has gotten quite small... It's nice to see. Once we can build the main features back on top (assuming they aren't already) , such as BP, constructor codes (e.g. fermionic states) based on BP, and solvers, that will be a good situation.

docs/src/interface_methods.md
mtfishman: That's encompassed by `ITensorNetwork(tensors)`, `keys(::AbstractVector)` and `values(::AbstractVector)` are defined as we would want them to be for this design:
```julia
julia> keys(["x", "y", "z"])
3-element LinearIndices{1, Tuple{Base.OneTo{Int64}}}:
 1
 2
 3

julia> values(["x", "y", "z"])
3-element Vector{String}:
 "x"
 "y"
 "z"

```

emstoudenmire:
Ready to merge this? Or let me know if other changes seem needed.

docs/src/deprecated_methods.md
```diff
@@ -11,6 +11,11 @@ Suggestions of methods which could be deleted.
   ITensorNetwork{V}()
   ```
 
+* From a collection of ITensorNetworks. Merges (Kronecker or tensor product) of input networks (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(itns::Vector{ITensorNetwork})
+  ```
```
mtfishman, line 24: I left a comment explaining that this can also be folded into the generic `ITensorNetwork(tensors)` constructor, did you see that?

mtfishman: I.e. https://github.com/ITensor/ITensorNetworks.jl/pull/346#discussion_r3179033421.

mtfishman: Oh sorry, I didn't see that this was in "deprecated".

docs/src/interface_methods.md
```diff
@@ -0,0 +1,328 @@
...
+  ```julia
+  ITensorNetwork(vertices, tensors)
+  ITensorNetwork{V}(vertices, tensors)
+  ```
+
+* From a vector of `ITensor`s, with vertex labels auto-assigned to `eachindex(ts)`.
+  Edges are inferred from shared indices (`itensornetwork.jl`):
+  ```julia
+  ITensorNetwork(ts::AbstractVector{ITensor})
+  ```
```
mtfishman: I think we don't need this as a separate one since it is encompassed by `ITensorNetwork(tensors)`. We could comment above `ITensorNetwork(tensors)` that it works for any collection, so includes `AbstractVector{ITensor}`.

emstoudenmire: Moved to deprecated with:
- a note about whether we should actually remove it or just streamline documentation
- a note about the generic constructor that it can (or should) handle this case and behavior of how vertex names are assigned

